PR #49, opened 2015-04-03
Imports Boston, Bournemouth, Runnymede, Wokingham Polling Stations

davidmiller:


polling_stations/apps/data_collection/management/commands/import_wokingham.py
```diff
@@ -0,0 +1,36 @@
...
+        #     'internal_council_id': record[-1],
+        #     'name': record[0],
+        # }
+        return
+
+    def station_record_to_dict(self, record):
+        return {
+            'council'            : self.council,
+            'internal_council_id': " ".join(record),
+            'postcode'           : '(No postcode supplied)',
```
symroe: Worth geocoding at point of import?  Google's reverse geocode is good enough most of the time.

symroe: Oh, looks like we have the point, just not the postcode?

davidmiller: Exactly - it has the point data, which is kind-of more important?

I wasn't quite sure what to do with this though.
We can ask for them to supply the postcode in the shapefile maybe? 
At the moment I'm concentrating on getting them to supply us with the districts in the FOIA chain...

symroe: Yeah, postcode isn't so important – let's ignore it for now.  I'll merge.

PR #57, opened 2015-04-04
Improves visual display & organization

davidmiller:
* Adds touch and favicons
* Clean up the display of results pages.
* Always put addresses in a right hand column.
* Set flags about the status of the data for this postcode in the view because David is not clever enough to hold all the state required to read the old template logic.
* Search councils by name in admin
* Update league table to show which councils we've loaded data for.


symroe:
> Always put addresses in a right hand column.

Sorry I've not checked this out myself, but have you tested this on various mobile sizes?  What does "right hand column" mean in this situation?

davidmiller:
It's bootstrap 3 responsive columns - Tablet landscape displays addresses to the right of the map (Council, polling station when we know).

screen shot 2015-04-04 at 15 24 04

Tablet portrait & thinner, it drops below the map.

screen shot 2015-04-04 at 15 21 52

Could probably use some tweaks to spacing / padding for mobile devices

symroe:
Ace!

PR #58, opened 2015-04-06
Importers for a number of councils.

davidmiller:
Imports data for Brent, Tyneside, Lichfield, East Riding, Spelthorne, Gravesham.

Adds a KML base importer.

Adds a new status page at /coverage that gives us some headline coverage metrics.

Refactors /api into it's own module because having it in urls.py was probably not sensible.


polling_stations/apps/data_finder/views.py
```diff
@@ -67,3 +67,28 @@ def get_context_data(self, **context):
...
+        for council in Council.objects.all():
+            if council.pollingstation_set.count() > 1:
+                stations += 1
+                covered.append(council)
+            if council.pollingdistrict_set.count() > 1:
+                districts += 1
+
+        context['num_councils']          = num_councils
+        context['num_district_councils'] = districts
+        print districts
```
symroe: Can you just remove these?

davidmiller: Presuming these == the two print statements?

Yes !

PR #108, opened 2015-04-27
Design changes to postcode page

symroe:
New design changes after @PhilRumens emailed some suggestions.

screen shot 2015-04-27 at 16 30 10


PhilRumens:
Looks really good. For the embedded version it might be good to remove "We think we know where you live", "Council Contact info" and the Disclaimer.

symroe:
@PhilRumens Can you explain why?  Is it because the data is better quality for you, so it might not be needed?  I'd still like to include the council phone number, as there may be other reasons to contact the council.

PhilRumens:
Some organisations (including councils) have a policy of not publishing phone numbers online. There's a whole debate about digital exclusion here that I'm not going to go into, but even those that don't do this generally promote online channels of communication over more traditional media, online.

So yes, perhaps it's best to include the phone number, but if it's provided to you, add social media and website details too, above the phone number.

What do you think?

Issue #60, opened 2015-04-07
Import E07000108-Dover


davidmiller:

chris48s:
I started writing an importer for this, but I noticed some of the postcodes (those with a 4 character postcode district) in the GIS layer provided are truncated e.g:
CT15 7B
CT17 9H
CT14 7B
Not sure how you want to deal with this?
- Import it anyway for now (the rest of the data looks good)
- Request corrected data
- Correct/Fill in the gaps by geocoding the addresses first

It looks like there will be a number of data files where missing/invalid postcode or point data is a problem, so perhaps it is worth writing a little wrapper class for the [google maps geocoding api](https://developers.google.com/maps/documentation/geocoding/intro) that can be used in importer classes. It is not always perfect, but it will probably be better than missing/inaccurate data. If you would like to see a different approach, let me know.

Can you assign this one to me so others don't duplicate, please?

Issue #22, opened 2015-04-02
Import E07000135-Oadby


davidmiller:

chris48s:
I started writing an importer for this, but there are some missing postcodes. See issue #60 for notes as it has similar problems.

Can you assign this one to me so others don't duplicate, please?

PR #136, opened 2016-02-03
Import scripts for Dover and Oadby & Wigston

chris48s:
A few notes on this:

1. I've excluded the proposed polling stations for Oadby & Wigston as none of them have grid references and based on the ids/codes it looks like some of them may be intended to replace existing polling stations. I've modified `import_polling_station_shapefiles()` so it is possible to conditionally ignore records by returning `None` in `station_record_to_dict()`.

2. I've attempted to fill in the missing/invalid postcodes by geocoding, which has helped but there are still some missing postcodes for these councils.

refs #22, #60

### DIFF
--- a/polling_stations/apps/data_collection/google_geocoding_api_wrapper.py
+++ b/polling_stations/apps/data_collection/google_geocoding_api_wrapper.py
@@ -1,0 +1,32 @@
+"""
+Wrapper class for Google Maps Geocoding API
+"""
+import urllib.parse, urllib.request
+import json
+
+
+class GoogleGeocodingApiWrapper:
+
+    def __init__(self, address):
+        self.address = address
+
+    def geocode(self):
+        address = urllib.parse.quote(self.address)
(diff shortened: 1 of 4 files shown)


polling_stations/apps/data_collection/google_geocoding_api_wrapper.py
```diff
@@ -0,0 +1,32 @@
...
+        address = urllib.parse.quote(self.address)
+        url = 'http://maps.googleapis.com/maps/api/geocode/json?sensor=false&address=%s' % address
+        req = urllib.request.urlopen(url)
+        str_response = req.read().decode('utf-8')
+        json_response = json.loads(str_response)
+        return json_response
+
+    """
+    Returns postcode if we have a decent match
+    Otherwise returns empty string
```
symroe: Might it be a little more pythonic to raise a (custom) exception here?

chris48s:
Agreed - updated as requested.
I've also added a sleep to `geocode()` to ensure we stay under the API's usage limit.

Issue #138, opened 2016-02-05
Serve compressed static assets from a CDN


symroe:
From @chris48s:
> At the moment, you're serving up some reasonably large js files (especially leaflet.js) from /static - but jquery, bootstrap and leaflet could all be pulled in from content delivery networks to reduce bandwidth requirements and reduce server load a bit. This probably isn't the most urgent improvement but it is very easy to do, so it seems like an easy win. I'll knock up a PR for this at the weekend, including a fallback to /static to avoid introducing an additional point of failure.

:+1: This, plus I expect we can optimize the assets we're using anyway.  I'm planning on swapping to Foundation (to keep us in line with the other DC projects), and we might find that using the SASS version of that will reduce the overall file size.

chris48s:
Finished/closed in new theme. :thumbsup:

PR #142, opened 2016-02-06
More efficient approach to loading javascript libraries

chris48s:
As discussed in #138.

A few notes:
- I've used [fallback.js](https://github.com/dolox/fallback) to provide resilience - it is a very lightweight library. Fallback can also do the same kind of thing for css, but I haven't done anything about that for the moment as you've said you plan to make significant changes to the css in the near future. Worth bearing in mind as there is further scope for optimisation if relying on generic css.
- I've put the same bit of code in base_full and base_embed (although base_embed doesn't actually seem to get used?) and some of the other templates also don't seem to be referenced anywhere - I guess some of it is WIP? If the intention is that both base templates will be used, I can refactor so this is more DRY as there is duplication here. Can you provide a bit of guidance on what's going on with the templates?
- While I was doing this, I spotted a few bits of invalid markup so I tidied those as well.


symroe:
On the templates:

The idea is that we can have a white label version of the site, currently at `/embed/` only, but in future this could be at `/foo/` and use a custom template (is one council wanted a specific design, for example).

To do this, I've hacked about with the URL configs in interesting ways.  You're best off looking at the `whitelabel` app for more on this, mainly `urls.py`, `context_processors.py` and `middleware.py`.

You'll see that `context_processors.py` is setting the `base_template` var to one of a list of values, with the last being `base_embed.html`.

The other fun thing is that the `whitelabel` app mounts the core URLs at the embed prefix – this is a bit of a hack, but it seems to work alright!

symroe:
Thanks for this, merged for now, and I'll get the common foundation stuff set up soon (it'll be a python package with the base templates and sass etc in)

PR #145, opened 2016-02-14
Refactor base templates to reduce repitition

chris48s:
Following on from Pull Request #142, I've factored some repeated content out of the base templates. Hopefully this will make them a bit easier to maintain moving forwards.
refs #138

### DIFF
--- a/polling_stations/templates/template.html
+++ b/polling_stations/templates/template.html
@@ -1,11 +1,10 @@
 {% load staticfiles %}<!DOCTYPE html>
 <html lang="en" ng-app>
   <head>
+    <title>{% block title %}Polling Stations (Alpha){% endblock title %}</title>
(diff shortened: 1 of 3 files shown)


polling_stations/templates/template.html
```diff
@@ -0,0 +1,67 @@
+{% load staticfiles %}<!DOCTYPE html>
+<html lang="en" ng-app>
+  <head>
+    {% block meta %}{% endblock meta %}
+
+    {% block icons %}{% endblock icons %}
+
+    <title>{% block title %}Polling Stations (Alpha){% endblock title %}</title>
```
symroe: Something tells me the title should be before the meta and icons blocks – I think older IE version have a problem when they don't find the title in the first n bytes of the document…

chris48s:
good spot - cheers

Issue #140, opened 2016-02-05
Write tests and set up CI


symroe:
Shamefully we don't have any tests for this project – this is really because it was a quite proof of concept that is turning in to a live service.

We should address this techincal debt!

symroe:
CI implemented in 8f808495b1d31cad6f6e12db4f69e98f94d09ae1

chris48s:
I've set up a wiki and added some manual testing resources

* Postcodes which take you on particular paths through the system or trigger various unusual conditions.
* A list of SQL statements I have found very useful for checking imported data.

In the short term, we should add more useful postcodes / testing resources to the wiki as we find them and make use of these when we change stuff.

Stretch goals should be:

* Use this as a basis to create testing fixtures which replicate these conditions and write automated tests to ensure website behaviour is correct and future changes don't break it.
* Use these SQL statements to add a task/report which can be run after importing new data files to flag up possible error conditions to investigate.

chris48s:
Closing this as we do now have some tests and Travis is set up. Specific issues relating to testing can now be raised as distinct issues (e.g: #290)

PR #149, opened 2016-02-19
test for json importer

chris48s:
Thinking a little more about issue #140, the base importers are quite difficult to write granular unit tests for in their current form. Given this, I think a good place to start is to write some higher level functional/integration type tests like this one. This isn't going to be terribly useful for identifying edge cases or telling you exactly where the error is if it doesn't pass. What this will do is provide some indication of "if this fails, something important is broken".

Does this seem like a reasonable approach?

refs #140


symroe:
This is an excellent approach – even if it's just a smoke test, it's far better than nothing.

Thanks!

PR #155, opened 2016-03-03
NUW Wales Theme & Cardiff importer

symroe:


chris48s:
Just one small suggestion.. apart from that, looks sensible to me

symroe:
Should have said – this is by no means the final design or anything!  I just threw it all together as an example of it working, and so there is something on the domain.

Given that I'll merge, and fix the alt attributes when we're working with a proper design.

PR #177, opened 2016-03-14
New theme

symroe:
This PR moves us to using the new `dc_base_theme` package (install with pip via requirements.txt) and sets up i18n.

Waiting on a postcode to test the `address_select_view`


symroe:
Reckon this is good to go…

Issue #189, opened 2016-03-19
Improve UX for address_select view on Android (possibly other mobile devices?)


chris48s:
If I put in a postcode like CF15 7RP on any desktop browser, the `address_select` view makes it very clear what the user is supposed to do next:

<img src="https://cloud.githubusercontent.com/assets/6025893/13899125/985aa33e-eddd-11e5-9fcd-2821b9b1a6b3.png" width="455" height="256"/>

<img src="https://cloud.githubusercontent.com/assets/6025893/13899126/9ab4de1a-eddd-11e5-9781-5891863d8b02.png" width="455" height="256"/>

When I tested the site on Chrome in Android, the way it presents the `<select>` element is far less clear (the `<option>`s don't appear till you touch the box). It works, but the first time I saw this I thought it had dropped me back to the home form because they look so similar:

<img src="https://cloud.githubusercontent.com/assets/6025893/13899137/b0726a60-eddd-11e5-851e-558fcf8304ef.png" width="240" height="430"/>

<img src="https://cloud.githubusercontent.com/assets/6025893/13899139/b56c1a52-eddd-11e5-9de1-bfd4c7d334cb.png" width="240" height="430"/>

1. What can we do to differentiate these 2 forms or make it a bit clearer what to do next for Android users?
2. Does this behave the same way in Safari on iOS? Does anyone have an iThingy they can test this with? I suspect the behaviour will also be similar on Windows Mobile.

This is a bit of an edge case because it only affects:
* users in some local authorities
* whose postcode is split between polling stations
* on some platforms

so it probably isn't top priority.

Issue #188, opened 2016-03-19
Update API page to use new theme


chris48s:
http://pollingstations.democracyclub.org.uk/api/beta/ still uses DRF bootstrap-based theme.
Updating is probably not high priority.

chris48s:
actually, lets get https://github.com/DemocracyClub/UK-Polling-Stations/issues/1040 applied to main site and api docs before requesting help with this

Issue #187, opened 2016-03-19
Issues with DC Theme


chris48s:
While I was doing some testing on the website, I found a few small niggles with the way that the website and the DC theme interact with each other. I haven't attempted to do anything about these because they aren't major problems and they potentially impact on other sites that use the theme (not just Polling Stations). I've raised an issue in case you think any of them are serious enough to warrant addressing.

* NUS Wales Themed version of the website throws a fatal error `Start tag “body” seen but an element of the same type was already open` in HTML validators because it defines 2 body tags - 1 in the DC base theme and then again in the Polling Stations NUS Wales theme:
```
<body >    
<body class="embed">
```
See https://validator.w3.org/nu/?doc=http%3A%2F%2Fwheredoivote.wales%2F
I've not seen any browsers which actually fail to render the page due to this, but valid markup would be ideal.

* Pages have empty title tags: `<title></title>`. This isn't terribly important for UX, but may impact on SEO a bit? Can this be moved into the site template without breaking other sites that use the base template?

* Header image feels like it should be a link, but it isn't - I keep clicking on it trying to get back to the home form. Because the header is defined in the base, it doesn't know what it should be a link to yet, so I understand why it isn't a link. This isn't a big problem - I think it is only annoying for me because I'm using the site in an unusual way - the typical user journey only involves searching for one postcode - not 50! ;) Maybe there should be a link back to home somewhere?

symroe:
Fixed in b2e7c4c752b230c5f7671432edd23f894b8fb19c

Issue #103, opened 2015-04-15
KeyError for invalid postcode


PhilRumens:
I get a KeyError when searching for http://pollingstations.democracyclub.org.uk/postcode/RG141AB/

RG141AB isn't a valid postcode, but it would be nice to get friendlier error.

davidmiller:
+10 !

tfgg:
Fixed in #195

PR #190, opened 2016-03-19
Website QA

chris48s:
OK, so I've spent this morning really tearing apart the front-end markup and javascript with a HTML validator and JSHint. I've also run through the core user journeys:

* home -> postcode view
* home -> we don't know where you should vote
* home -> address select view -> address view

in the following browsers:

* Firefox
* Chrome (Windows)
* Chrome (Android phone)
* Microsoft Edge (Windows 10)
* IE 11 (Windows 10)
* IE 10 (Windows 7)
* IE 9 (Windows 7)
* IE 8 (Windows 7)

and just for a laugh..
* IE 6 (Windows XP)

I've not tried it in Safari on OSX or iOS though. Can someone else give that a bash?

I was expecting to suggest that we add something like
```
<!--[if lt IE 10]>
    <p class="browserupgrade">You are using an <strong>outdated</strong> browser.<br />Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
<![endif]-->
```
but I was actually really surprised at how well the site works even in really old versions of IE (although I wasn't looking for small niggly issues in those browsers - just "does it roughly work").

Back to IE 9 everything seems to work (and look) totally fine for those core paths through the system. In IE 8 the styling starts to look a bit ropey, but it downgrades relatively gracefully and the map still works. In IE 6 (is anyone really still using this?), the map doesn't work and the site looks stylistically terrible, but everything is roughly in the right places and you can put a postcode in and get a polling station address or council details out of it.

I've proposed some edits to the markup and javascript here, but it is all very minor stuff. Please don't feel that I am [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality) over silly things with this PR. I think that the fact that I've tried really hard to find problems and all the changes I've proposed are essentially trivial is evidence of what a good job you've done on producing a theme that looks great while still being really compatible/accessible across such a range of platforms! Great work on that :thumbsup:

I did find a few other things which were less trivial or clear-cut, so I've raised issues for them: #187, #188, #189

### DIFF
--- a/polling_stations/apps/nus_wales/templates/nus_wales/base.html
+++ b/polling_stations/apps/nus_wales/templates/nus_wales/base.html
@@ -20,12 +20,12 @@
     <footer class="footer">
       <div class="row">
         <div class="columns large-12">
-          <img src="{% static 'images/nus_logo.jpg' %}" width=200>
+          <img src="{% static 'images/nus_logo.jpg' %}" width=200 alt="NUS logo">
           <a href="https://democracyclub.org.uk">
           {% if LANGUAGE_CODE == 'cy-gb' %}
-          <img src="{% static 'images/logo-cy.png' %}" width=200>
+          <img src="{% static 'images/logo-cy.png' %}" width=200 alt="Clwb Democratiaeth">
           {% else %}
-          <img src="{% static 'images/logo-with-text.png' %}" width=200>
+          <img src="{% static 'images/logo-with-text.png' %}" width=200 alt="Democracy Club">
           {% endif %}</a>
         </div>
       </div>
--- a/polling_stations/templates/base.html
+++ b/polling_stations/templates/base.html
@@ -9,7 +9,7 @@
 
 {% block ga_tracking_code %}
 {% if not debug %}
-<script>
+<script type="text/javascript">
   (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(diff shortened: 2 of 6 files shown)


JoeMitchell:
Nice. I've played with Safari and all seems good on those same user
journeys.

Did make me realise we probably need an About, Privacy, Terms at some
point. (I'm not sure what the legal requirements are). Probably a cookies
thing (i wouldn't bother, but since WGovt are sponsoring). Should I make
these issues and write some copy?

And if you put a postcode that's not recognised you get a Whoops! message,
which should probably be "We are sorry, we did not recognise this postcode.
Please try again / Please test your postcode via Postcode Finder
<http://www.royalmail.com/find-a-postcode>."

JoeMitchell:
Oh one small thing. On Chrome on Android, I get this on the address
selection... (It's fine if I tap the box, but it's not obvious that there's
anything there)

chris48s:
Cheers for having a look, Joe.

An about/FAQ/privacy policy/terms and conditions page is probably a good idea. IMO a sentence or 2 about cookies on there is probably sufficient. There is already an about page at http://pollingstations.democracyclub.org.uk/about which you could use as a base to start from - it just isn't linked anywhere (obvious).

Whoops for invalid postcode is already logged as issue #103
UX improvement on address select for android users is logged as #189

JoeMitchell:
Thanks Chris, you're all over it!

symroe:
Thanks for the thorough testing, Chris!

PR #191, opened 2016-03-19
More fixes to get_polling_station() logic

chris48s:
Polling Station IDs and Polling District IDs are only unique within councils - they're not unique on a national basis so I've constrained the lookups by id within the relevant local authority boundary.
This does not actually cause a bug right now, but it will probably save someone a headache in future when we inevitably encounter a collision.

### DIFF
--- a/polling_stations/apps/data_finder/views.py
+++ b/polling_stations/apps/data_finder/views.py
@@ -21,12 +21,12 @@ from whitelabel.views import WhiteLabelTemplateOverrideMixin
 from .forms import PostcodeLookupForm, AddressSelectForm
 from .helpers import (
     geocode,
+    PostcodeError,
     get_ors_route,
     get_google_route,
     OrsDirectionsApiError,
     GoogleDirectionsApiError,
 )
-
 
 # sort a list of tuples by key in natural/human order
 def natural_sort(l, key):
@@ -139,7 +139,12 @@ class PostcodeView(BasePollingStationView):
         return directions 
 
     def get_context_data(self, **context):
-        l = geocode(self.kwargs['postcode'])
+        try:
+            l = geocode(self.kwargs['postcode'])
+        except PostcodeError as e:
+            context['error'] = e
+            return context
+
         context['location'] = Point(l['wgs84_lon'], l['wgs84_lat'])
 
         context['council'] = Council.objects.get(
--- a/polling_stations/apps/data_finder/views.py
+++ b/polling_stations/apps/data_finder/views.py
@@ -13,7 +13,6 @@ from django.utils.translation import ugettext as _
 from councils.models import Council
 from data_finder.models import LoggedPostcode
 from pollingstations.models import (
-    PollingDistrict,
     PollingStation,
     ResidentialAddress
 )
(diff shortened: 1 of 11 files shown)


chris48s:
Tracked down some more strange edge cases in the view logic. Hopefully the comments I've added explain the problems I'm trying to prevent with these changes and make the flow a bit clearer. I've done a fair bit of testing on this, but this one could benefit from thorough review before merging. Bounce it back to me if you find any problems.

symroe:
We want to add `[unique_together](https://docs.djangoproject.com/en/1.9/ref/models/options/#unique-together)` on `council,internal_council_id` as another way to catch this class of problem in the importers.

polling_stations/apps/data_finder/views.py
```diff
@@ -103,29 +103,49 @@ class BasePollingStationView(TemplateView, LogLookUpMixin):
...
+    def get_polling_station(self, location, council_id):
         try:
             polling_district = PollingDistrict.objects.get(
                 area__covers=location)
         except PollingDistrict.DoesNotExist:
             return None
 
+        if not polling_district:
+            # is this different to PollingDistrict.DoesNotExist exception?
+            return None
```
symroe: I think this will never be hit, as per the comment.  `.get` returns either the object, or raises a `DoesNotExist` exception.

symroe:
Nice logic clean up.

Thinking about it, we should really move this logic to the model's manager, or a helper class.  I'm not very keen on having a load of logic in the view layer (I know I put it there in the first place!).  This would also allow for simpler API code, too.

I'll make a different issue for that though, rather than block this PR.

symroe:
See #202 for the issue about the above comment.

chris48s:
Right.. easy stuff first: I've removed the redundant check, so hopefully this one is good to go now - pending refactoring it into a ModelManager?


Just to clarify, the **main** purpose of this clean up was to get away from looking up `polling_station_id=''` on every request and relying on it returning more than one result and move to only trying that lookup when polling_station_id is non-empty. I hadn't actually spotted any failures due to dupes on `(internal_council_id, council_id)`, but now you mention it...


> We want to add unique_together on council,internal_council_id as another way to catch this class of problem in the importers

Good call. I think when I add this, I want to start from an empty DB and run all of the import scripts to see if any now fail due to this constraint (and if they do try to fix them!). I think this is a larger undertaking than I will have time for tonight, so I'm going to defer this one till the weekend - then I'll add the commit(s) for that to PR #193. Does that sound OK?

I've just done a quick
```sql
SELECT
DISTINCT(internal_council_id, council_id), COUNT(*)
FROM pollingstations_pollingstation
GROUP BY internal_council_id, council_id
HAVING COUNT(*)>1
ORDER BY COUNT(*);
```

and

```sql
SELECT DISTINCT(internal_council_id, council_id), COUNT(*)
FROM pollingstations_pollingdistrict
GROUP BY internal_council_id, council_id
HAVING COUNT(*)>1
ORDER BY COUNT(*);
```

on my dev DB and those returned some records for Bristol, Brent and Canterbury, so it looks like adding that check should flag some problems for those areas. My data for those areas comes from the `initial_data.json` in the repository - not running the import scripts. The last time I rebuilt my dev DB, I imported `initial_data.json` and then ran the import scripts that I wrote on top of that. However, I found that there were some differences in the older data between my dev copy and the live site. Then I re-ran some of the older import scripts and the differences went away, so some of those could be false positives now?

This is actually one of the things on my list of stuff that I didn't have time to deal with at the weekend that I meant to follow up with you:
- When you last rebuilt the DB on the live site, did you use the init fixture that is in the repo, or did you just run all the import scripts from scratch?
- At some point, when we've got these data quality issues nailed down, it is probably worth updating the init fixture in the repo so that it is easier to build a dev DB consistent with the live site for testing.

PR #193, opened 2016-03-20
Change ALL THE THINGS!

chris48s:
Over the last week or so, we have:
* Fixed a lot of strange behaviour in the `get_polling_station()` logic and template code
* Implemented station_id -> district_id lookup
* Talked a lot about data quality

As a result of all this, I've really understood a lot more about how the finding process works and what conditions in the data shouldn't happen or might cause unintended behaviour to occur. Also, squashing some of the bugs in `get_polling_station()` has helped to provide a front end which makes it much more obvious when strange conditions exist in the data, making problems much easier to find.

Off the back of that, I've done an absolute tonne of testing on the data I've imported since I started working on this and noticed a **lot** of bugs that needed fixing or improvements that could be made. In some cases, I have made changes which just improve performance. For example, in some scripts I've just added an ID so that the district/station lookup can be done more efficiently (it will be much faster to do a query like `...WHERE internal_council_id='AA'` than a spatial query like `...WHERE ST_Contains(polygon, point)`). In others, I've spotted and fixed bugs that either weren't obvious because of bugs in the view logic, or which were there just as a result of me not understanding the project/data well enough when I wrote those import scripts :(

Notes:

1. Before re-running these import scripts, the fix in PR #191 must be merged because now that there are a lot more councils with content in the `polling_district_id` and `polling_station_id` fields, there are a lot of collisions. To give one example, the polling district code 'GA' is used independently in Plymouth, Watford, Mid Sussex, Sheffield, Kingston and Cardiff to identify different areas.

2. Having been through this process, I suspect that a lot of the older import scripts that I used as templates for these may have similar problems, but to start with I have concentrated on the import scripts that I wrote myself. There is definitely some additional testing that needs to be done here.

### DIFF
--- a/polling_stations/apps/data_collection/management/commands/import_dover.py
+++ b/polling_stations/apps/data_collection/management/commands/import_dover.py
@@ -1,12 +1,13 @@
 """
 Import Dover
 """
+import shapefile
+from django.contrib.gis.geos import Point
 from data_collection.management.commands import BaseShpShpImporter
 from data_collection.google_geocoding_api_wrapper import (
     GoogleGeocodingApiWrapper,
     PostcodeNotFoundException
 )
-
 
 class Command(BaseShpShpImporter):
     """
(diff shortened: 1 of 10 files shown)


symroe:
These all look great, although I'd be much happier once `unique_together` was added, as per #202.

I guess this PR is blocked awaiting #202 being merged?

PR #195, opened 2016-03-21
Add postcode-not-found response, issue #103

tfgg:
This is a fairly hacky fix for #103. Currently you get an error if you put in something that looks like a postcode but actually isn't. This adds an exception and a branch in the template.

It should probably do the checking in the form validation code, but currently that'd invoke two postcode lookups, requiring caching. This'll do for a fairly rare error.

image

### DIFF
--- a/polling_stations/templates/home.html
+++ b/polling_stations/templates/home.html
@@ -17,7 +17,9 @@
       <label class="card_header">{% trans "Enter Your Postcode" %}
         <input id="id_postcode" name="postcode" type="text" aria-describedby="postcode_help_text" autofocus >
       </label>
-      <p class="help-text" id="postcode_help_text">{% trans 'e.g. <a href="/postcode/CF105AJ/">CF10 5AJ</a>' %}</p>
+      <p class="help-text" id="postcode_help_text">
+        {% trans 'e.g.' %} <a href="{% url 'postcode_view' postcode="CF105AJ" %}">CF10 5AJ</a>
+      </p>
       <button type="submit" class="button" id="submit-postcode">{% trans "Find your Polling Station" %}</button>
     </form>
   </div>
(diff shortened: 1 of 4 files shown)


polling_stations/templates/home.html
```diff
@@ -17,7 +17,7 @@
       <label class="card_header">{% trans "Enter Your Postcode" %}
         <input id="id_postcode" name="postcode" type="text" aria-describedby="postcode_help_text" autofocus >
       </label>
-      <p class="help-text" id="postcode_help_text">{% trans 'e.g. <a href="postcode/CF105AJ/">CF10 5AJ</a>' %}</p>
+      <p class="help-text" id="postcode_help_text">{% trans 'e.g. <a href="/postcode/CF105AJ/">CF10 5AJ</a>' %}</p>
```
symroe: This will break when in embed mode (see http://pollingstations.democracyclub.org.uk/embed/), so might need to be replaced by a `{% url %}` tag, and to be tested at that URL prefix.

polling_stations/templates/postcode_view.html
```diff
@@ -9,12 +9,25 @@
...
 {% block content %}
-
 <div class="row">
   <div class="columns large-8">
     <div class="card">
+    {% if error %}
+
+      <h2>Sorry, we can't find {{ postcode }}</h2>
+      <p>This doesn't appear to be a valid postcode.</p>
+      <form method="post" action="/" class="form form-inline">
```
symroe: As above, the action here should be a `{% url %}`

polling_stations/templates/postcode_view.html
```diff
@@ -9,12 +9,25 @@
...
+          {% csrf_token %}
+          <label class="card_header">{% trans "Enter Your Postcode" %}
+            <input id="id_postcode" name="postcode" type="text" aria-describedby="postcode_help_text" autofocus >
+          </label>
+          <p class="help-text" id="postcode_help_text">{% trans 'e.g. <a href="/postcode/CF105AJ/">CF10 5AJ</a>' %}</p>
```
symroe: Dito

tfgg:
Seems to work in the /embed/ workflow.

PR #196, opened 2016-03-21
Change to MapQuest tiles

tfgg:
I think the MapQuest style tiles are less confusing, but this could be controversial. User testing required.

Before:

image

After:

image

You could also worry about their usage/SLA: http://wiki.openstreetmap.org/wiki/MapQuest#MapQuest-hosted_map_tiles

### DIFF
--- a/polling_stations/templates/postcode_view.html
+++ b/polling_stations/templates/postcode_view.html
@@ -121,9 +121,12 @@
       // map.doubleClickZoom.disable();
       map.scrollWheelZoom.disable();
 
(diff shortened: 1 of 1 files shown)


chris48s:
Definitely agree that the less 'busy' tile layer looks better. Not sure if the usage limit is a problem or not. Anyone closer to the project got any idea what sort of level of usage you're expecting?

symroe:
I think we might go over that limit, but only over a day or two – on a yearly average we'll be under the threshold ;)

I don't think they can block us either, as all the requests will come from client computers.  We don't want to break their trust, but I think it's fine for a lumpy (in terms of usage) app like this.

Given that, I think it's ok to go in.

PR #212, opened 2016-03-26
Fix issues #192 and #211 + some refactoring

### LINKED ISSUE #192: HomeView and PostcodeView should route consistently
At the moment, if you type a postcode like CH7 4BR into the postcode search, it does not give the same behaviour as going directly to the URL `/postcode/CH74BR`. `HomeView` and `PostcodeView` should implement the same routing.

### LINKED ISSUE #211: "name 'build_directions_url' is not defined" on /address endpoint
5abe603 removes the function `build_directions_url()` and moves the logic into the `PostcodeView` class, but the `AddressView` class is still trying to call `build_directions_url()`. This function should be moved out to somewhere `PostcodeView` and `AddressView` can both reference it and the `/address` endpoint fixed.

chris48s:
Sorry - this is another PR that changes lots of things in one pull request, but once I started factoring stuff out of the views that didn't need to be there, everything just started coming together so I ran with it..

My original intention was to fix #211 so that I could test #192 properly, fix that, batch those 2 into one pull request and be done with it. Once I did that, I found that I was left with `PostcodeView.get_context_data()` and `AddressView.get_context_data()` doing roughly the same things in different ways (and some more inconsistencies that we hadn't previously spotted) so ended up doing more refactoring.

The net result of all this is:
* It should be easier to make changes to the polling station view without `AddressView` and `PostcodeView` diverging because as much of the code as possible now lives in `BasePollingStationView`. However, it is still important to test changes against both of the endpoints representing a sucessful user journey.
* It should be a lot clearer where `PostcodeView` and `AddressView` are the same and where they are different.
* It is easier to write tests because a lot of the logic we want to test is now in view helpers or model managers and (mostly) decoupled from code that is calling external APIs or dealing with HTTP responses... to which end, I've added some tests :)
* There is less repetition.
* The views are now a lot 'thinner', which is good.

Obviously this makes a lot of changes to important things. I've checked it against all of the postcodes in the [wiki](https://github.com/DemocracyClub/UK-Polling-Stations/wiki/Postcodes-for-Testing) to cover as many different logical paths through the system as possible, but a good review/some additional testing would be useful! So much for freezing `views.py`, eh! ;)

Finally, apologies if the unified diffs for 9eb1ab2 and 24675a9 are a bit hard to read (less has changed than it looks like - there's just a lot of moving stuff about). It might be clearer to look at a split diff for those commits.

Closes #192
Closes #211
Refs #140

### DIFF
--- a/polling_stations/apps/data_finder/views.py
+++ b/polling_stations/apps/data_finder/views.py
@@ -1,6 +1,7 @@
 import re
 import requests
 
+from abc import ABCMeta, abstractmethod
 from operator import itemgetter
 
 from django.contrib.gis.geos import Point
@@ -59,7 +60,59 @@ class HomeView(WhiteLabelTemplateOverrideMixin, FormView):
 
 
 class BasePollingStationView(TemplateView, LogLookUpMixin):
+    __metaclass__ = ABCMeta
+
     template_name = "postcode_view.html"
+
+    @abstractmethod
+    def get_location(self):
+        pass
+
+    @abstractmethod
+    def get_council(self):
+        pass
+
+    @abstractmethod
+    def get_station(self):
+        pass
+
+    def get_directions(self):
+        if self.postcode and self.station and self.station.location:
+            dh = DirectionsHelper()
+            return dh.get_directions(
+                start_postcode=self.postcode,
+                start_location=self.location,
+                end_location=self.station.location,
+            )
+        else:
+            return None
+
+    def get_context_data(self, **context):
+        try:
+            l = self.get_location()
+        except PostcodeError as e:
+            context['error'] = e
+            return context
+
+        if l is None:
+            # AddressView.get_location() may legitimately return None
+            self.location = None
+        else:
+            self.location = Point(l['wgs84_lon'], l['wgs84_lat'])
+
+        self.council = self.get_council()
+        self.station = self.get_station()
+        self.directions = self.get_directions()
+
+        context['postcode'] = self.postcode
+        context['location'] = self.location
+        context['council'] = self.council
+        context['station'] = self.station
+        context['directions'] = self.directions
+        context['we_know_where_you_should_vote'] = self.station
+
+        self.log_postcode(self.postcode, context)
+        return context
 
 
 class PostcodeView(BasePollingStationView):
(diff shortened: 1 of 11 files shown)


polling_stations/apps/data_finder/views.py
```diff
@@ -59,103 +60,114 @@ def form_valid(self, form):
...
     template_name = "postcode_view.html"
 
+    @abstractmethod
+    def get_location(self):
+        pass
 
-class PostcodeView(BasePollingStationView):
+    @abstractmethod
+    def get_council(self):
+        pass
```
symroe: Should these raise a [`NotImplementedError`](https://docs.python.org/2/library/exceptions.html#exceptions.NotImplementedError) rather then passing?

symroe:
This is excellent!  Yeah, hard to read the diff, but everything is so much cleaner now.

The only thing I'd like to change (but this wont block the merge), is abstracting further the logic in `RoutingHelper`.  I'd like a method that returns the actual queries directly, and then `get_endpoint` can call that method.

The reason for this is that I'm looking in to the API a little bit this morning, and found myself copying the logic from HomeView (now in `RoutingHelper`) in order to return the API response.  This abstraction doesn't help me with this, as it's only giving me a routing endpoint, not the querysets I need to display in the APIView.

I'll do a PR for the API stuff now (it's rough, so still WIP), so you can see what I'm trying to do.  Happy to merge this (after the comment about `NotImplemented`) and then I can deal with abstracting `RoutingHelper`.

symroe:
See #214 for my work on the API – I've just rebased onto this branch and made the changes that clean things up a bit in RoutingHelper.

chris48s:
> Should these raise a NotImplementedError rather then passing?

The `@abstractmethod` decorator is sort of a shorthand for this, although it raises `Can't instantiate abstract class [classname] with abstract methods [method name]` when you try to instantiate the class if any of the abstract methods aren't implemented rather than instantiating the class, potentially running some of the methods and then throwing an exception when it encounters a not implemented method (which could leave you in an inconsistent state). It isn't necessary to manually raise another exception here.

However, it is really useful you've flagged this though because in checking this I have realised that the way you're supposed to use the `ABCMeta` class is different in Python 3 ( http://stackoverflow.com/a/7196503 ) and I've done it in the old Python 2 way (which is maybe why an exception was not thrown if you commented out a method in `AddressView` or `PostcodeView`?). As such, I've added another commit which updates to the Python 3 way of using `ABCMeta`. If you comment out a method in one of the concrete classes, you should now find an exception like `Can't instantiate abstract class PostcodeView with abstract methods get_location` is thrown - as intended.

Changes in 4032d23 look sensible - it is probably better that you did that as you had a clearer idea how you wanted to use those properties in the API code :)

Cheers

symroe:
Really interesting, I had no idea about `@abstractmethod`!

PR #213, opened 2016-03-26
Importer for Merthyr Tydfil

### LINKED ISSUE #198: Import W06000024-Merthyr-Tydfil

symroe:
This closes #198


chris48s:
Ran the import script (Christ that takes a long time!). It ran for about 3 hours then died with:
```
  File "..../polling_stations/apps/data_finder/helpers.py", line 26, in geocode
    raise PostcodeError("Mapit error {}: {}".format(res_json['code'], res_json['error']))
data_finder.helpers.PostcodeError: Mapit error 400: Postcode 'CF481OT' is not valid.
```
Did you get the same thing, or did it finish for you?

Anyways, if you look at the csv in a text editor, there's a whole load of lines on the end of the file that are just `,,,,,` (in fact this is the majority of the file!), but you're still processing them all - this is inserting a blank address with a polling station id of "-" in the `ResidentialAddress` table. It is also making the script take way longer than necessary to run. The easiest way to deal with this is probably just to delete them from the csv.

I'm confused about `known_stations` - you never seem to put anything in it.. am I missing mods to the base class here?

There's a bunch of addresses which say stuff like "Other Electors, Other Electors Address", "OTHER ELECTORS, PANTYSCALLOG SOUTH, MERTHYR TYDFIL" - we're never going to match any of this data based on a user's postcode search, so there's probably no point putting them in the table, although I don't think they necessarily cause a problem - I wasn't able to test properly because the import never finished.

One of the polling stations in the file is "to be confirmed". We should remove this so the user goes to the "we don't know where you should vote - call your council" page, rather than "we know where you should vote - your polling station address is: to be confirmed".

Some of this is commentary on the data, rather than the import script.

In general we need to make the `BaseAddressCsvImporter` more efficient - it must be driving you nuts trying to write/debug this, but a lot of it will be due to all the redundant lines of `,,,,,`. To make it less painful, you could just take a copy of the file, open it in a spreadsheet editor, delete the `Property` and `UPRN` columns, de-dupe the file and save off a separate stations csv.

chris48s:
> data_finder.helpers.PostcodeError: Mapit error 400: Postcode 'CF481OT' is not valid.

Just realised while checking other import scripts this is because as of c68c043, `geocode` now throws `PostcodeError` instead of `KeyError`.

symroe:
Yeah, sorry, I'd forgotten that I got that postcode error.  Good spot on the "TBC" station – we've asked the council about both issues.

chris48s:
This is still handling the wrong exception - it needs to catch `PostcodeError` like in 00f3d52

I made that change and ran it, but now as of 35ba2f6 the values in `ResidentialAddress.polling_station_id` don't match up with the values in `PollingStation.internal_council_id`, so searching for any postcode says "we don't know where you should vote - contact your council".

symroe:
Updated the data again, and fixed the ID issue – this seems to be working now?

PR #222, opened 2016-04-01
Use slugs instead of ids for /address endpoint

### LINKED ISSUE #194: /address endpoint should use predictable URLs
At the moment, the address endpoint uses the primary key, so we generate URLs like `/address/123456`. This works, but if we need to re-run the import script (e.g: because we've fixed a bug in it, or because a council provides updated data), all the IDs will change. If a user is relying on the URL because they've bookmarked it for later use or shared it (e.g: with a family member or housemate), the URL will now throw a 404.
`ResidentialAddressImporter` should be modified to generate a repeatable unique ID so the user sees something more like `/address/1-foo-street-bar-town-aa11aa` to minimise disruption to the public URL schema.

chris48s:
Because slug is `UNIQUE` and `NOT NULL`, you won't be able to apply the migration with data already in the DB, so you'll need to clear the ResidentialAddress table, apply the migration and re-run `import_denbighshire` and `import_rct`.

Closes #194

### DIFF
--- a/polling_stations/apps/data_collection/management/commands/__init__.py
+++ b/polling_stations/apps/data_collection/management/commands/__init__.py
@@ -358,28 +358,37 @@ class BaseAddressCsvImporter(BaseImporter):
         value = re.sub('[^\w\s-]', '-', value).strip().lower()
         return mark_safe(re.sub('[-\s]+', '-', value))
 
+    def get_slug(self, address_info):
+        # if we have a uprn, use that as the slug
+        if 'uprn' in address_info:
+            if address_info['uprn']:
+                return address_info['uprn']
+
+        # otherwise build a slug from the other data we have
+        return self.slugify(
+            "%s-%s-%s-%s" % (
+                self.council.pk,
+                address_info['polling_station_id'],
+                address_info['address'],
+                address_info['postcode']
+            )
+        )
+
     def add_residential_address(self, address_info):
 
         """
         strip all whitespace from postcode and convert to uppercase
         this will make it easier to query this based on user-supplied postcode
         """
-        postcode = re.sub('[^A-Z0-9]', '', address_info['postcode'].upper())
+        address_info['postcode'] = re.sub('[^A-Z0-9]', '', address_info['postcode'].upper())
 
         # generate a unique slug so we can provide a consistent url
-        slug = self.slugify(
-            "%s-%s-%s-%s" % (
-                self.council.pk,
-                address_info['polling_station_id'],
-                address_info['address'],
-                postcode
-            )
-        )
+        slug = self.get_slug(address_info)
 
         ResidentialAddress.objects.update_or_create(
             council=self.council,
             address=address_info['address'],
-            postcode=postcode,
+            postcode=address_info['postcode'],
             polling_station_id=address_info['polling_station_id'],
             slug=slug
         )
--- a/polling_stations/apps/data_collection/management/commands/__init__.py
+++ b/polling_stations/apps/data_collection/management/commands/__init__.py
@@ -9,6 +9,7 @@ import re
 import shapefile
 import sys
 import tempfile
+import unicodedata
 import urllib.request
 import zipfile
 
@@ -18,6 +19,8 @@ from django.core.management.base import BaseCommand
 from django.contrib.gis import geos
 from django.contrib.gis.gdal import DataSource
 from django.contrib.gis.geos import Point, GEOSGeometry
+from django.utils.encoding import force_text
+from django.utils.safestring import mark_safe
 
 from councils.models import Council
 from pollingstations.models import PollingStation, PollingDistrict, ResidentialAddress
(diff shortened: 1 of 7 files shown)


polling_stations/apps/data_collection/management/commands/__init__.py
```diff
@@ -331,6 +334,30 @@ def import_polling_stations(self):
 
 class BaseAddressCsvImporter(BaseImporter):
 
+    def slugify(self, value):
```
symroe: Can we use django's slugify here?  Will they produce different results?

polling_stations/apps/data_collection/management/commands/__init__.py
```diff
@@ -339,11 +366,22 @@ def add_residential_address(self, address_info):
         """
         postcode = re.sub('[^A-Z0-9]', '', address_info['postcode'].upper())
 
+        # generate a unique slug so we can provide a consistent url
+        slug = self.slugify(
```
symroe: Can this be a `get_slug` method that importers can overwrite?  My thinking here is that we should use UPRNs as slugs wherever possible – they are better unique IDs than this (but we don't always have them)

chris48s:
OK, so first off I've moved slug generation out to a function `get_slug()`. This can be overriden if necessary, but I've also set it up so that if the dict returned by `address_record_to_dict()` contains a key `'uprn'`, that can be used for the slug.

I haven't actually done this in either `import_rct` or `import_denbigshire` for the following reasons:

Denbighshire
------------
Denbighshire haven't provided full addresses - where all the addresses in a single street or village map to the same polling station, they've just given us the street name or village name and then duplicated that for each uprn so using the uprn as the slug means we insert duplicates, producing confusing output in the `AddressFormView` (I'm relying on `insert_or_update()` de-duping for us).

RCT
---
There is at least one duplicate UPRN in the RCT data. For example, the addresses 

> 128 YNYSHIR ROAD, YNYSHIR, PORTH, CF39 0EW

> 128A YNYSHIR ROAD, YNYSHIR, PORTH, CF39 0EW

both have the UPRN 200001485430. Obviously **in principle** UPRN should be a unique id for every address, but I **suspect** that for the purposes of the electoral roll if you tell them your address is something which hasn't been assigned a UPRN (e.g: because you live in a property which has been unofficially sub-divided, or you permenantly reside in an institution, or whatever) they will still let you register with whatever address you provide.. or possibly it is just a mistake? Either way, that's a fun edge case.

Also I've not done this on the Merthyr script because the same condition also exists in the Merthyr data. For example the following 2 addresses:

> WARDS 3 TO 4 AND 9 AND 19, NORTH GLAMORGAN NHS TRUST, PRINCE CHARLES HOSPITAL, GURNOS, MERTHYR TYDFIL, CF47 9DT

> FLAT 1, NORTH GLAMORGAN NHS TRUST, PRINCE CHARLES HOSPITAL, GURNOS, MERTHYR TYDFIL, CF47 9DQ

are both listed with the UPRN 10034652635

Having seen this, although it would be nice to be able to always use UPRN as slug, I think the reality is that we are likely to find this in the data provided by most councils (except Ceredigion, presumably). Why can't stuff just be simple goddammit!!?

Custom slugify() function
-------------------------
I realise that defining a custom `slugify()` is a bit of a strange thing to do, I've given an example of why in the comments on lines 339-354. To give a real-world example from the data: in RCT, the following 2 different addresses:

> 3/4 HOPKIN STREET, TREHERBERT, TREORCHY, CF42 5HL

> 34 HOPKIN STREET, TREHERBERT, TREORCHY, CF42 5HL

generate the same slug using django's built-in `slugify()` because the `/` character is simply discarded whereas the custom `slugify()` I've defined will replace it with a -, generating 2 different slugs for these (different) addresses.

chris48s:
In working with the AddressBase data on Ceredigion (and trying to work out why the number of addresses coming out is larger than the number of UPRNs going in), I have actually found out the answer to this. See the question "How will multiple occupancy properties be shown in AddressBase?" at https://www.ordnancesurvey.co.uk/business-and-government/help-and-support/products/addressbase-technical-faqs.html (sorry I can't lik directly to that question).

It seems that internally subdivided properties share a UPRN, so they're not actually unique to an **address** - just a **property** (which are subtly diferent). In one sense, this may not be a big problem because I think it would be fairly mental for a council to say "flat 1 in this building goes to polling station A and flat 2 goes to polling station B" (famous last words!!). On the other hand, if we want to slug it by UPRN, we would probably need a way to collapse "Flat 1, 1 Foo Street, Bar Town" and "Flat 2, 1 Foo Street, Bar Town" down to "Flats 1 and 2, 1 Foo Street, Bar Town" or "1 Foo Street, Bar Town" or something in the import scripts. I think that is just making more work for ourselves than necessary though, given that the formatting of addresses can be a bit haphazard.

symroe:
I should have known better than to assume we could rely on a 3rd party ID system for anything!  :D

Thanks for the reasoning on slugs – I understand now :)

Issue #175, opened 2016-03-13
Obtain data directly from the web in preference to files (where possible)


chris48s:
Thinking ahead a bit, as I've noted in issues #8, #93, #167, #170, #171, #172, #173 there are a number of councils where we can obtain data from a (predictable) URL. In these situations, the import scripts should extract the data directly from their website/API instead of from a file. This will make it easier to update the data if we become aware of updated information or for future elections.

As a side note, we can search google for things like `allinurl: gov uk getcapabilities` or `allinurl: gov uk wms` to discover more WMS API description endpoints we might be able to harvest further station/boundary data from in future (I think in some cases, this will include some councils we've already written import scripts for), but I've been thinking about Wales and London for now.

chris48s:
Closed in favour of #356

PR #225, opened 2016-04-03
Import script for Camden

### LINKED ISSUE #171: Import E09000007-Camden
Camden council publish high quality data for their polling stations and districts in a variety of useful formats:
https://data.gov.uk/dataset/camden-polling-districts
https://data.gov.uk/dataset/camden-polling-stations

chris48s:
Given the number of formats Camden publish, I thought this would be easy, but I couldn't get their KMZ output to parse in GEOS. The KML would parse, but the metadata was stored in `ExtendedData` tags with no schema definition in the head, so GEOS wouldn't parse anything other than the boundaries.. In the end, I went for their 'JSON with WKT strings embedded in it' (if only some kind of specialised JSON-based format for geographic data existed!) and XML outputs. Maybe it would have been quicker to just download the SHP and CSV files, but hopefully the `BaseGenericApiImporter` class will be useful in the future and this will be easier to update when they publish new data.

The meta-data says this is the polling data for "GLA Mayoral and Assembly Elections - Thursday 5 May 2016" so it is probably worth checking back on this before the EU Referendum.

We're missing a station for district ED - I checked the outputs in other formats, but it is missing from all of them :(

Closes #171
Refs #175

### DIFF
--- a/polling_stations/apps/data_collection/management/commands/__init__.py
+++ b/polling_stations/apps/data_collection/management/commands/__init__.py
@@ -283,7 +283,7 @@ class BaseKamlImporter(BaseImporter):
             tmp.close()
 
 
-class BaseApiKmlKmlImporter(BaseKamlImporter):
+class BaseGenericApiImporter:
     srid             = 4326
     districts_srid   = 4326
     districts_url    = None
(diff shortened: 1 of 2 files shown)


Josh--M:
Hi

1. The station for district ED was missed in error. The [dataset](https://opendata.camden.gov.uk/Your-Council/Camden-Polling-Stations/5rhh-fxna) has now been updated to include the station, so please re-run the import when you get the chance.

2. At present this version of the dataset only applies to the May 5th elections. To be honest, it's 98% likely to remain the same for the EU referendum, but we will re-up the dataset with the election name 'EU referendum June 23rd' to be explicit. If there are any changes I'll be doubly sure to give y'all a nudge.

Josh
(I work at Camden - have been liaising with Sym on making this data available)

chris48s:
Just re-ran the importer locally and it is grabbing all the stations now - cheers @Josh--M :)

PR #232, opened 2016-04-08
Data Quality Report

chris48s:
I decided to formalise a bunch of the checks I've been doing on imported data into a report we can run to flag up possible problems with data or import scripts and make it easier to debug import scripts when writing them. Using this code, we can run something like:
`python manage.py data_quality_report S12000036`
to spit out a data quality report for Edinburgh.

Note that the warnings aren't necessarily errors - in some cases, they just flag limitations of the data. For example, if we run
`python manage.py data_quality_report E09000009`
it will tell us there are 5 invalid district references. This isn't a problem we can fix - as I've noted in PR #221 this is just what the data says, but it is useful to be able to find this stuff out with one command rather than lots of tedious manual checks - as I've been doing.

For the moment I've just provided this as a standalone management command, but I could also modify the base classes so that at the end of `handle()`, we call
```
report = DataQualityReport(self.council.pk)
report.output_report()
```
and output this as the final step of each import script, or we could put some of this info into `CoverageView` (although most of this is probably too detailed for that).

### DIFF
--- a/polling_stations/apps/data_collection/data_quality_report.py
+++ b/polling_stations/apps/data_collection/data_quality_report.py
@@ -1,0 +1,374 @@
+from django.db import connection
+from django.db.models import Q
+from pollingstations.models import PollingStation, PollingDistrict, ResidentialAddress
+
+
+# define some methods we can use to print coloured console output
+class OutputFormatter:
+    OKGREEN = '\033[92m'
+    WARNING = '\033[93m'
+    ENDC = '\033[0m'
+    BOLD = '\033[1m'
+
+    @staticmethod
+    def print_ok(text):
+        print(OutputFormatter.OKGREEN + text + OutputFormatter.ENDC)
+
+    @staticmethod
+    def print_warning(text):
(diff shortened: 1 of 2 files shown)


polling_stations/apps/data_collection/data_quality_report.py
```diff
@@ -0,0 +1,374 @@
+from django.db import connection
+from django.db.models import Q
+from pollingstations.models import PollingStation, PollingDistrict, ResidentialAddress
+
+
+# define some methods we can use to print coloured console output
+class OutputFormatter:
+    OKGREEN = '\033[92m'
```
symroe: It might be better to use https://docs.python.org/3/library/curses.html for colour stuff – I think it can get quite complex…

symroe:
This is amazing!  Thanks!  The comment I left can be ignored for now, and just turned in to an issue, as it's only us who will run it for the time being (i.e., there is little chance of our shells not supporting the implementation of colours).

PR #239, opened 2016-04-15
Publish data quality reports on website

### LINKED ISSUE #237: Produce web page version of data quality report
Run data quality report at the end of each import script and populate a static/cached database for performance reasons. Link it from the API page/coverage page/other 'nerd zone' area of the site.

chris48s:
Following on from our conversation in the standup on Tuesday, I've added the data quality report output to the website. At the moment, it is just chucking the same output we send to the console into a text field and rendering it in a `<pre>` tag, but it does the job - we can always tidy up later. Thinking it through, the league table is actually the sensible place for this to be linked so I've made some improvements to that as well.

I've marked this WIP because we probably don't want to merge this (or at least shouldn't deploy it to live) until we tear down and re-run the import scripts because the import scripts will populate the `DataQuality` model with suitable static/cached values.

Closes #237

### DIFF
--- a/polling_stations/apps/data_collection/data_quality_report.py
+++ b/polling_stations/apps/data_collection/data_quality_report.py
@@ -307,6 +307,7 @@ class DataQualityReportBuilder():
 
     def __init__(self, council_id):
         self.council_id = council_id
+        self.report = []
 
     def build_header(self):
         self.report.append({ 'style': None,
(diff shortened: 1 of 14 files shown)


chris48s:
Removed WIP tag as PR #241 addresses the issue of populating `DataQuality` model

Issue #231, opened 2016-04-07
Import E06000023-Bristol (new data available)


chris48s:
The data we have for Bristol is out of date. We should delete the stale data and import their new data. We should be able to harvest it from their arcgis server. The metadata for the relevant datasets is at:
https://maps.bristol.gov.uk/arcgis/rest/services/ext/localinfo/MapServer/15
https://maps.bristol.gov.uk/arcgis/rest/services/ext/localinfo/MapServer/133
but I haven't figured out how to construct a [query](https://maps.bristol.gov.uk/arcgis/rest/services/ext/localinfo/MapServer/15/query) that says 'gimmeh all teh datas!' yet.

chris48s:
We can get the data out:

stations
----
https://maps.bristol.gov.uk/arcgis/rest/services/ext/localinfo/MapServer/15/query?where=OBJECTID+LIKE+%27%25%27&text=&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&relationParam=&outFields=&returnGeometry=true&maxAllowableOffset=&geometryPrecision=&outSR=&returnIdsOnly=false&returnCountOnly=false&orderByFields=&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&gdbVersion=&returnDistinctValues=false&f=pjson

https://maps.bristol.gov.uk/arcgis/rest/services/ext/localinfo/MapServer/15/query?where=OBJECTID+LIKE+%27%25%27&text=&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&relationParam=&outFields=&returnGeometry=true&maxAllowableOffset=&geometryPrecision=&outSR=&returnIdsOnly=false&returnCountOnly=false&orderByFields=&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&gdbVersion=&returnDistinctValues=false&f=kmz

districts
----
[...]

PR #240, opened 2016-04-16
Add ability to serve custom content for specific areas

### LINKED ISSUE #174: Councils with their own polling stations finders
Some councils have their own polling station search. For example, Hammersmith & Fulham have a polling station search on their website: https://www.lbhf.gov.uk/polling-stations-search (it isn't great because it doesn't seem to know which district you are in - it just lists stations near you).
I don't think we can extract the data from it conveniently (I think we would have to write a script to feed in postcodes one at a time and scrape the output), but if we don't get any data from Hammersmith & Fulham, should we link to their finder from the "we don't know where you should vote" page rather than just telling users to phone them?

As a related issue, if we don't manage to get as far as importing data for Carmarthenshire and Hackney (as noted by @davidmiller in issue #167 we should prioritise councils where we have complete data), they also have polling station maps on their websites that we could link to:
http://ilocal.carmarthenshire.gov.uk/my-nearest/
http://www.map.hackney.gov.uk/lbhackneymap/

### LINKED ISSUE #179: Point NI postcodes to official site
For the time being, http://www.eoni.org.uk/Offices/Postcode-Search-Results?postcode=BT5+7TQ is going to be far better than us.  We should detect if a postcode is in NI and offer a link to that site.

### LINKED ISSUE #199: 'Coming soon' text for postcodes in councils that have provided data?
Ceredigion Council were a bit sad that they provided data but we haven't been able to use it (due to AddressBase lack thereof). 

They wondered if we could put a holding message for their pages. e.g. "This council has supplied the data. We're working on building maps with it." Or similar...

Not a high priority given lack of traffic I think.

chris48s:
I've generalised this feature a bit so that as well as covering custom polling station finders (Issues #179, #174, GLA tool), it can also be used to show a custom message (e.g: "coming soon") as requested in #199.

Notes on this:

Seed data
----
I haven't provided any seed data for this - just some rough examples of how you might populate the DB in this comment: https://github.com/chris48s/UK-Polling-Stations/blob/152f0ca8e1cac16cfd15986b51bbd883ab4713b7/polling_stations/apps/pollingstations/models.py#L135
I'll leave the finer points of which areas you want to include and how you want to word the copy to you. When you decide what you're putting in there, could you let me know/commit a migration/send me a SQL dump or something so I can mirror the same content in my dev db.

Translations
----
This commit adds text which will need to br translated. Also you'll need to get translations for any text which goes into the `CustomFinder.message` field. I've marked it for translation using `ugettext`.

Order of precedence
----
I didn't get any feedback on my request for comment in #179, so I have made an assumption and set this up so that if a custom finder/message exists for an area where we hold data, it will prefer our data. This means if you add a record for the GLA's finder with the code `E15000007` and a user puts in a postcode in a London local authority where we have recent data (e.g: Camden, City of London, Ealing), etc we will serve up the 'we know where you should vote' output and only direct users to the GLA's finder in areas where we don't hold data. If you want it to work the other way round and always prefer the GLA site in this situation, let me know and I'll modify the logic accordingly.

API
----
Hopefully the way I've done the API output is sensible in terms of how you want to use it?

Other custom finders
----
In terms of organisations that have their own polling station finders (other than Northern Ireland and probably GLA), the ones I'm aware of and think are worth linking to are Stroud and Bristol:
Stroud's is at
https://stroud.maps.arcgis.com/apps/webappviewer/index.html?id=ea6bf4b3655542c1a05c8d7e87d32bb1
but doesn't accept postcodes as a URL param.
Bristol's is at
https://www.bristol.gov.uk/voting-elections/polling-station-finder
and you can feed a postcode into it like this:
https://www.bristol.gov.uk/voting-elections/polling-station-finder?execution=e1s7&p_p_id=bccpollingstationfinder_WAR_bccpollingstationfinderportlet&_eventId_lookupAddress=&postCode=AA1+1AA
(will accept postcode with a space, but doesn't require it like NI's does). That would also mean we don't have to deal with the data I managed to find, which I think will be problematic - see issue #231.
I think all of these:
https://www.lbhf.gov.uk/polling-stations-search
http://ilocal.carmarthenshire.gov.uk/my-nearest/
http://www.map.hackney.gov.uk/lbhackneymap/
are not worth linking to. I haven't really come across any others, but there are probably more out there. Is it worth asking in Slack/on twitter/on your mailing list/somewhere if anyone knows if their local council has one that is any good now that we can actually link to them?

Closes #174
Closes #179
Closes #199
Refs #231

### DIFF
--- a/polling_stations/api.py
+++ b/polling_stations/api.py
@@ -6,7 +6,7 @@ from django.utils.encoding import smart_str
 from rest_framework import routers, serializers, viewsets, views
 from councils.models import Council
 from pollingstations.models import (PollingStation, PollingDistrict,
-    ResidentialAddress)
+    ResidentialAddress, CustomFinder)
 
 # Fields define serialization of complex field types (GEO)
 class PointField(serializers.Field):
@@ -148,6 +148,14 @@ class PostcodeViewSet(viewsets.ViewSet):
             ret['polling_station_known'] = True
             ret['polling_station'] = PollingStationSerializer(
                 polling_station, context={'request': self.request}).data
+
+        if not ret['polling_station_known']:
+            finder = CustomFinder.objects.get_custom_finder(l['gss_codes'], postcode)
+            if finder.base_url:
+                ret['custom_finder'] = {}
+                ret['custom_finder']['base_url'] = finder.base_url
+                ret['custom_finder']['can_pass_postcode'] = finder.can_pass_postcode
+                ret['custom_finder']['encoded_postcode'] = finder.encoded_postcode
 
         return Response(ret)
 
--- a/polling_stations/apps/data_finder/helpers.py
+++ b/polling_stations/apps/data_finder/helpers.py
@@ -25,9 +25,14 @@ def geocode(postcode):
     if 'error' in res_json:
         raise PostcodeError("Mapit error {}: {}".format(res_json['code'], res_json['error']))
     else:
+        gss_codes = []
+        for area in res_json['areas']:
+            if 'gss' in res_json['areas'][area]['codes']:
+                gss_codes.append(res_json['areas'][area]['codes']['gss'])
         return {
             'wgs84_lon': res_json['wgs84_lon'],
             'wgs84_lat': res_json['wgs84_lat'],
+            'gss_codes': gss_codes,
         }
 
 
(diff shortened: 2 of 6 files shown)


JoeMitchell:
Cheers Chris - I like the review of others' finders!

Good question on GLA ... There's doesn't exist yet, right? So I think your
way makes sense...

JoeMitchell:
There's also this from Trafford, still working out if useful: http://www.trafford.gov.uk/about-your-council/elections/polling-places.aspx

PR #253, opened 2016-04-26
Use nice icons for house and polling station

### LINKED ISSUE #203: Do we need a house icon for the start location?
Esko suggested that we add a house location, and @JoeMitchell point out that the start looks a little empty now we have route planning on the map.

We should consider this, esp when useability testing.

chris48s:
Yay - pretty pictures!
Closes #203


chris48s:
I've rebased this onto master to resolve the conflict

PR #258, opened 2016-05-01
Use custom mapit install with env vars, handle hitting rate limit

chris48s:
* We can now set environment vars `MAPIT_URL` and `MAPIT_UA` to specify a custom url and user agent for calls to mapit. If not set, the default is http://mapit.mysociety.org/ and blank user agent.

* If using MySociety's mapit install and we hit their rate limit, show a slightly more user-friendly error message.

### DIFF
--- a/polling_stations/api.py
+++ b/polling_stations/api.py
@@ -91,7 +91,12 @@ class PollingDistrictViewSet(viewsets.ModelViewSet):
     serializer_class = PollingDistrictSerializer
 
 from rest_framework.response import Response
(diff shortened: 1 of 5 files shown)


chris48s:
Sorry - forgot to tag some text for translation (and spotted some more that also wasn't tagged yet). These errors get shown on all branded versions so should ideally be translated if possible.

symroe:
Can we send an email to `ADMINS` when we raise the rate limit exception?

chris48s:
Updated as requested

Issue #267, opened 2016-05-11
Crowdsource Polling Stations


chris48s:
Overview
-----
* Latest OS Boundary line contains polling district boundaries for all 325 councils in England - that's 31,825 boundaries.
* Councils publish PDFs with their polling station addresses and district codes. e.g:

http://www.adur-worthing.gov.uk/media/media,100300,en.pdf

http://www.stalbans.gov.uk/Images/ldrsPollingStations_tcm15-40023.pdf

http://www.derby.gov.uk/media/Derby_City_Council_Polling_Station_List_2016.pdf

http://www.richmondshire.gov.uk/council-democracy/elections/divisions-wards-polling-district-and-polling-stations

* We could use this to try and crowdsource full (or full-ish) coverage of England.

Absolute bare minimum viable product
-----

1. Write an import script for English Polling Districts file in latest Boundary Line. The polling stations shape file contains 31,825 polling districts - so that's the number of data points we would be looking to source. Non-trivial as shape file does not have GSS codes for councils - only names, so requires some faffing about! :(

2. Devise a mechanism for people to provide data. Absolute bare minimum implementation would be to set up a google spreadsheet for people to enter data into. It should hold:
council name (pre-populated by us)
council ons/gss code (pre-populated by us)
polling district id (pre-populated by us)
address (user fills in)
postcode (user fills in)
source (user fills in/we provide if possible - we need this to ensure some basic level of data quality)

3. Generate the seed data (fairly easy once 1 is done)

4. Write an import script to take CSVs exported from google docs and bash them into the PollingStations model (+geocode if postcode present so we can do maps where possible)

5. Work out how to extract new/latest data as its entered and import it.. perhaps the simplest way to manage this would be to set up ~325 seperate spreadsheets and then import each one as it is filled up rather than try to manage one huge 31,825 line spreadsheet - can we automate bulk-generating google spreadsheets from CSV templates? That way we could assign people a spreadsheet per council with a link to PDFs?

6. Source as many links to council polling station PDFs as possible.

7. Recruit army of data wombles.

8. ???

9. Profit

JoeMitchell:
"a spreadsheet per council with a link to PDFs" +!

andylolz:
It sounds like there are two workstreams here:

 * polling district shapefile stuff
 * polling station locations (inc polling district reference)

It sounds like these can be completed independently.

It’s also worth noting that official EU referendum polling station location documents will be released 2 June 2016 at the latest.

[...]

Issue #172, opened 2016-03-13
Import E09000012-Hackney


chris48s:
We can obtain polling station data for Hackney from their website:
http://www.map.hackney.gov.uk/geoserver/wms/kml?mode=download&layers=hackneymap:Polling%20Stations

chris48s:
Consolidated into #356

Issue #170, opened 2016-03-13
Import E09000003-Barnet


chris48s:
Barnet council publish a csv of their polling stations:
https://data.gov.uk/dataset/polling-stations-in-barnet
https://open.barnet.gov.uk/dataset/polling-stations-in-barnet

chris48s:
Consolidated into #356

PR #268, opened 2016-05-15
[WIP] Crowdsourcing starter kit

chris48s:
This addresses steps 1 to 3 under the technical side discussed [here](https://github.com/DemocracyClub/UK-Polling-Stations/issues/267#issuecomment-218831211). Here's the lowdown:
## Importing Polling Districts
1. Whereas we usually store data in the repository boundaries for England massively exceed GitHub's 100Mb file size limit. To import the districts:
   - Grab the latest OS Boundary Line from https://www.ordnancesurvey.co.uk/opendatadownload/products.html
   - Run `python manage.py import_england_districts '/path/to/download/bdline_gb-2016-05/POLLING DISTRICTS_(ENG)/Shape'`
2. There are some problems with the data (different boundaries with the same code, multiple slightly different versions of boundaries, duplicates). I've contacted Ordnance Survey Open Data Helpdesk about it and @tfgg is going to follow it up with a contact at OS who may be able to help. You can catch up with the discussion in Slack. For the moment, I've just excluded any councils with one or more problematic boundaries. It gives us more than enough data to test/pilot with. Depending on whether we get an updated/corrected layer from OS, we can review this at a later date.
3. Data quality report is intentionally not produced/written to DB for boundaries only - we populate it for each local auth when we import the stations.
## Generating CSVs
1. We can generate CSV templates for collecting crowdsourced data using the script provided, but I've pre-generated what we need for testing by running `python manage.py generate_csvs /path/to/my/checkout/data/crowdsourcing/pilot`.
2. I think a reasonable way to manage this would be for those who review/approve the google docs to export a CSV and then submit the reviewed CSV (or batches of them) to the repo as PRs. Happy to discuss that.. The data quality report will also be a useful tool for checking quality of crowdsourced data, speaking of which..
## Importing data
1. Once we've pulled data into the repo, we can run a command like `python manage.py import_crowdsourced_csv X01000001 /path/to/your/checkout/data/crowdsourcing/pilot/X01000001.csv` to import the data and produce a quality report.
2. One issue I've anticipated is character encoding, so there is an optional `-c/--character-encoding` param, but there will probably be others I've not considered.
3. I've assumed that we're not going to have grid refs and we'll have to geocode.
## ...and finally

Having a dataset of polling districts also means that there are some areas which publish polling stations in a more useful format than PDF, or with grid refs (e.g: [Hackney](https://github.com/DemocracyClub/UK-Polling-Stations/issues/172), [Barnet](https://github.com/DemocracyClub/UK-Polling-Stations/issues/170)). We've been ignoring these, but we can do something with them now. It is probably worth looking out for these where possible as it may provide better quality data or reduce the need for data entry.

Refs #267


chris48s:
Based on discussions on Slack with @andylolz, I've made a couple of further modifications to this:

1. In some cases, we may have good data for most polling districts in a local auth but one or 2 records that don't match up with the codes we have or are in some way ambiguous. It would be useful to be able to leave these blank and just serve up the "we don't know - call your council" page for those districts. Added some code to discard records with blank/address and postcode (as opposed to inserting empty string x2).

2. There are some councils who publish their polling stations in a non-PDF format with district codes, but we had been ignoring them because they didn't publish district boundaries to go with them.. this brings them back into play. For example:
https://data.gov.uk/dataset/polling-stations9
https://data.gov.uk/dataset/polling-stations-trafford
https://data.gov.uk/dataset/bristol-polling-stations
https://data.gov.uk/dataset/salford-polling-stations
https://data.gov.uk/dataset/lichfield-district-council-polling-stations
https://data.gov.uk/dataset/northumberland-polling-district-breakdown-including-voter-totals
(but we need to check when they were updated). In some cases, these may also have grid refs, so I've factored some stuff out into a base class so that we can write other variations on the Stations Only importer more easily.

chris48s:
As requested, CSVs can now be fetched direct from google docs. Syntax for `import_crowdsourced_csv` changes to now require either `-u` or `-f`, so you can run one of:
`python manage.py import_crowdsourced_csv X01000001 -f data/crowdsourcing/ref.2016-06-23/X01000001.csv`
or
`python manage.py import_crowdsourced_csv X01000001 -u 'https://docs.google.com/spreadsheet/ccc?key=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx&output=csv'`


`ba57109` is particularly important in this situation IMO as it effectively automates a consitency check that previously relied on manually inspecting the data quality report. My initial thought was to replace all the `update_or_create()` calls in `__init__.py` and target a PR at master, but then I did that, did a teardown, tested against all the existing import scripts and realised there are existing import scripts which rely on the `update_or_create()` behaviour to deal with dupes in council-supplied input files. This probably is something to address in future, but I think this is a reasonable approach at this point in time.


Tagged this as WIP because it probably still is.

chris48s:
Abandoned :(

chris48s:
We're getting the band back together!


Having reviewed this, here's a quick rundown of how to repurpose this for a electoral roll-based crowdsourcing effort:

1. Get rid of `import_england_districts.py` (this imports the OS BoundaryLine data) and replace it with a new import script for the electoral roll data e.g: `import_electoral_roll.py`.

2. You'll need to keep `EXCEPTIONS`. When you import Electoral Roll, you don't want to import that electoral roll data for any areas where we already hold `ResidentialAddress` or `PollingDistrict` data and we won't want to generate CSVs for them either. We can lose `BAD_DATA`. `BAD_DATA` is only relevant to the OS BoundaryLine data.

3. `EXCEPTIONS` will need to be updated to include ONS codes for the latest import scripts. I haven't bothered to update it to include any areas where I've added import scripts since closing this PR.

4. Once Electoral Roll is imported, you'll want to rewrite `generate_csvs.py` to use Electoral Roll/`ResidentialAddress` model as a base data source for this (as opposed to OS BoundaryLine districts/`PollingDistricts` model). Don't rely on the CSVs generated from OS data (side issue: might be interesting to compare them at a later date). Codes may or may not be the same. You'll also want to reference `EXCEPTIONS` here so we don't generate CSVs for areas where we already hold council-supplied data.

5. Any new scripts (e.g: `import_electoral_roll.py`) should be added to exclude[] in `import.py`

6. `names_codes.json` can be deleted. That was only necessary for the OS BoundaryLine data.

7. `import_crowdsourced_csv.py` should pretty much stand as it is, assuming we're crowdsourcing from PDF.

chris48s:
Now working on this in #427

PR #270, opened 2016-05-15
Replace register to vote text (w/7th June date)

chris48s:
Put the 'You need to be registered in order to vote' text back in with registration deadline for EU Referendum.

Note: because I've changed the date, the English text for this `blocktrans` has changed so we'll need to update the translation accordingly for the Welsh version

### DIFF
--- a/locale/cy_GB/LC_MESSAGES/django.po
+++ b/locale/cy_GB/LC_MESSAGES/django.po
@@ -108,10 +108,15 @@ msgstr "\n<h3>Nid yw cerdyn pleidleisio yn angenrheidiol er mwyn pleidleisio</h3
 msgid ""
 "\n"
 "        <h3>You need to be registered in order to vote</h3>\n"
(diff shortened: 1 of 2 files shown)


andylolz:
Looks good to me. Is there a way to do this such that it doesn’t break the Welsh language site?

chris48s:
Yes you're right. I've fiddled about with the trans blocks and po file so we only need a translation for "before the 7th of June" now, assuming that's a sensible place to split it (apparently phrases that involve numbers are prone to mutation changes).

andylolz:
I _think_ house style for date format is of the form “7 June” rather than “the 7th of June”.

JoeMitchell:
Yes Andy! Style guide:
https://docs.google.com/document/d/1jEPCXk8AMzX9R10GI0KmOvW_RRNc_coCQFq-oCdVgDc/edit

chris48s:
missed the boat on this one! :anchor:

PR #277, opened 2016-05-20
Do not log page when no-log param is present

andylolz:


chris48s:
This will cause all requests to the `/address` endpoint to fail with `'AddressView' object has no attribute 'do_not_log'`. You also need to set `self.do_not_log` in `AddressView.get()`. Probably in order to achieve the intended behaviour that will mean you need to pass into `RoutingHelper`'s constructor and return it in `get_endpoint()`.

polling_stations/apps/data_finder/views.py
```diff
@@ -135,6 +136,7 @@ class PostcodeView(BasePollingStationView):
 
     def get(self, request, *args, **kwargs):
         rh = RoutingHelper(self.kwargs['postcode'])
+        self.do_not_log = request.GET.get('no-log') is not None
```
symroe: I think this can either remove the `is not None` as if `self.do_not_log` is None it will still evaluate to `False`, or it could be `request.GET.get('no-log', False)`.

chris48s:
Also remember to test in the case where the user has been though the `AddressFormView` between `/postcode` and `/address`

andylolz:
Sorry, this is WIP. Shouldn’t have sent PR (I haven’t even tried running it yet).

symroe:
:D the Slack integration is cool (because I wanted to jump in and look as soon as `#bots` told me about it!)

As an aside, maybe we should look at why the tests passed on Travis. I'm happy to write some more tests to catch this sort of thing if needed.

andylolz:
> maybe we should look at why the tests passed on Travis

+1 I was surprised to see that notification

chris48s:
Also +1

The reason no tests failed is because the test suite only covers `RoutingHelper`, `PollingStationManager` and the import classes and this didn't break any of them, but tests for the views would definitely be great.

This is an off-the-cuff response, but I think the barrier to properly testing these is that at the moment, calls to external APIs (mapit, directions finders) aren't injected so probably the first step to being able to properly write tests for these is to refactor so that the response from mapit can be mocked. The directions finder apis are probably more marginal.

PR #304, opened 2016-06-11
Integration tests

symroe:
Super basic, but it should be easy enough to write more. The main point is to check both 'happy paths' in a smoke test sort of way.


polling_stations/apps/data_finder/features/index.feature
```diff
@@ -1,14 +1,24 @@
 Feature: Check Postcodes
-    Scenario: Check Islington
-    Given I put in the postcode "N19SN"
```
davidmiller: Didn't realise that you were explicitly checking that pollingstations will work for me personally on every test run!

symroe: All part of your Personalised Democracy Experience.

chris48s:
Pulled this down to have a quick peek, but I'm getting `FileNotFoundError: [Errno 2] No such file or directory: 'phantomjs'` when I try to run the tests. Is PhantomJS a package that also needs to be added to the testing requirements file, or is it a system-level dependency that should be added to the install instructions in the readme?

symroe:
Ah sorry, I missed this as I obviously had `phantomjs` installed already, and Travis comes with it pre-installed too. 

On OS X I can just do `brew install phantomjs` – it looks like there isn't a Ubuntu package for it, so it might not be so simple. Are you able to have a go and document if this works? http://phantomjs.org/build.html

symroe:
Or `npm -g install phantomjs` might work?

chris48s:
Cool. Added a commit with that.
Pulled the branch down, tried breaking a few obvious things and ensured the tests fail so I reckon this is probably good. As you say, more cases can be added to check for more subtle errors at a later date.

symroe:
KK, thanks, I'll merge as it seems it's not actually making things worse :)

Issue #295, opened 2016-06-06
Import E09000031-Waltham-Forest


JoeMitchell:

chris48s:
Will probably have similar issue to Coventry - we're missing the names for named houses/flat numbers/etc. Haven't checked how many addresses might be affected.

chris48s:
Closing this to make priorities clearer. We've got an import script that provides nearly-complete coverage - can't see any of us tweaking this in the next week.

PR #305, opened 2016-06-11
Import script for Waltham Forest

chris48s:
Refs #295

Decided to be nice and tackle some of the data that is a royal PITA and leave you with some of the easier ones to deal with.
See commit log for notes on what I've done with the notes/grid refs. I've taken the same approach as we did with Coventry and just removed cases that cause ambiguity. Not marking this for closure as there is maybe potential to improve by using AddressBase to derive full addresses rather than just removing the ambiguous rows.


symroe:
Thanks - this looks great, but what a pain! We need to find a way to deal with these cases without having to edit the data ourselves!

Issue #265, opened 2016-05-10
Wrong polling station returned


PhilRumens:
We've noticed that the wrong polling station is returned in a few cases. Two example postcodes are

RG184NX
RG189PT

You can see the correct stations by putting the postcode in the search on http://www.westberks.gov.uk, selecting an address, and expanding "Your Councillors".

We've received complaints about this from the public, so we'll have to take your search of our site if this isn't resolved. :(

chris48s:
Hi Phil,

First off, I'm really sorry to hear your users are having problems and that you have had complaints about this.

If we look at RG18 4NX:
http://ww2.westberks.gov.uk/InternetMapping/MyProperty.aspx?uprn=200004738234 and 
http://ww2.westberks.gov.uk/InternetMapping/MyProperty.aspx?uprn=100081027372 (for example)
you'll note that these 2 properties in the same postcode have different polling stations because the postcode crosses a polling district boundary.

With the data West Berkshire have provided (i.e: shapes representing polling districts), we aren't able to split postcodes which cross a boundary - a postcode maps to one polling district. This means in situations like this where a postcode crosses a boundary, there is necessarily a small but non-zero error rate: some of the properties in that postcode will map to the wrong polling station.

We can solve this with the right data though - some councils have provided explicit mappings of addresses/UPRNS to polling stations, which allows us to split out the addresses in postcodes which have this property. For example, if you search for CF15 7RP on http://pollingstations.democracyclub.org.uk/ - you'll get a list of addresses. 1-4 Rhymney Cottages map to one polling station and 'The Old Forge' maps to a different one.

You can read more about this in the technical specs for providing pololing station data at https://democracyclub.org.uk/projects/polling-stations/technical/ - this issue is why we recommend Address to Station lookups as the gold standard for supplying data as it ensures accuracy.

Are you able to supply us with the data in that format? If so, we can do the same for West Berkshire and solve this problem for you.

I'm a bit more confused about RG18 9PT - I've been through every address listed at http://ww2.westberks.gov.uk/InternetMapping/AddressFinder.aspx?src=snop&snop=RG189PT and they all map to "Riverside Community Centre Rosemoor Gardens Newbury RG14 2FG" - and that's where our site sends people as well: http://pollingstations.democracyclub.org.uk/postcode/RG189PT/ . Can you give any more detail on this one?

Cheers,
Chris

PhilRumens:
[...]

PR #308, opened 2016-06-12
AddressBase postcode fixes

symroe:
This work will help in cases where we only have shape data (Silver quality data).

It imports AddressBase and creates ResidentialAddress models in the case where properties within a postcode span a polling district.

Refs #265

TODO

* [x] Deal with case where the polling district doesn't know the polling station ID in `make_addresses_for_postcode` (do we have a generic abstraction of this logic)
* [x] Run after importing new districts, in the base importer


chris48s:
> do we have a generic abstraction of this logic

It is all tied up in `get_polling_station()` : https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/pollingstations/models.py#L37

You would probably need to split that function up into a number of smaller functions in order to extract the bits you need:
* The case where stations store a district id and
* The case where we do a polygon lookup (currently only used for City of London I think). As we've done with `get_polling_station_by_id()` Fortunately that function is well covered by the tests so it should be reasonably easy to refactor and be confident nothing has broken.

symroe:
Ok, this is in a fairly good shape I think.

On the data quality report:

I'm a little confused as to how I should make this – @chris48s can we have a chat about the best way to add the following to the report:

1. `postcodes_contained_by_district`
* `postcodes_with_addresses_generated`

I can add them to the model directly, but I'm not sure how best to add them to the actual report

symroe:
The other thing to note about this is that it slows down the importers quite a lot. I think this is reasonable, but it would be nice to see if there's any speed optimisations to be made.

chris48s:
> On the data quality report:
>
> I'm a little confused as to how I should make this – @chris48s can we have a chat about the best way to add the following to the report:
>
>    postcodes_contained_by_district
>    postcodes_with_addresses_generated
>
> I can add them to the model directly, but I'm not sure how best to add them to the actual report

OK. Off the top of my head, the `build_report()` method will summarise anything in Stations, Districts or Addresses:
https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_collection/data_quality_report.py#L470
If the target local auth for the report has one or more record in any of those tables, you'll get some metrics so in principle you should already be seeing 3 sections in the data quality report.

I'd probably add an appropriate `get_thing()` method to `DistrictReport` (for the report about districts) and an appropriate `get_thing()` method to `ResidentialAddressReport` (for addresses). Call your model methods in here and then add calls to them in `DataQualityReportBuilder.build_district_report()` and `DataQualityReportBuilder.build_residential_address_report()` in sensible places with appropriate formatting.
See https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_collection/data_quality_report.py#L476 for style values. The existing code will handle outputting that or saving to database.

Assuming this is going to get run after every districts importer, there shouldn't be anything to break particularly there, but its worth ensuring that whatever gets added does something sensible for a straight up `ResidentialAddress` dataset.

If you add any additional object properties to `..ReportBuilder` classes, remember to re-initialise them in the constructor as we sometimes generate instances of this in a loop (e.g: unit tests or the command that runs all the importers for a given election).

That should be it...probably.

polling_stations/apps/addressbase/management/commands/import_cleaned_addresses.py
```diff
@@ -0,0 +1,28 @@
...
+
+from django.db import connection
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'cleaned_ab_path',
+            help='The path to the cleaned AddressBase file'
```
chris48s: Minor point, but can this be clarified to "The path to the folder containing the cleaned AddressBase CSVs" or similar

polling_stations/apps/addressbase/management/commands/clean_addressbase.py
```diff
@@ -0,0 +1,100 @@
...
+            line['DEPARTMENT_NAME'],
+            line['PO_BOX_NUMBER'],
+            line['SUB_BUILDING_NAME'],
+            line['BUILDING_NAME'],
+            line['BUILDING_NUMBER'],
+            line['DEPENDENT_THOROUGHFARE'],
+            line['THOROUGHFARE'],
+            line['POST_TOWN'],
+            line['DOUBLE_DEPENDENT_LOCALITY'],
+            line['DEPENDENT_LOCALITY'],
```
chris48s: I realise this is the order they appear in the source CSVs, but the addresses read strangely in this order (look at some addresses you know). The order should be:

```
line['DOUBLE_DEPENDENT_LOCALITY'],
line['DEPENDENT_LOCALITY'],
line['POST_TOWN'],
```

symroe:
Due to time constraints I'm going to ignore the report for the time being – I'll make an issue about it

chris48s:
I ran a couple of `ResidentialAddress` importers and A/B tested the data - its exactly the same, which is good.
I've tried a few district-based import scripts but all of the ones I've tried fail with

```
AttributeError: 'NoneType' object has no attribute 'internal_council_id'
helpers.py line 45, in make_addresses_for_postcode
```

The ones I tried running are West Berkshire, Camden and City of London.

symroe:
@chris48s I've added the thing you suggested in slack: passing `council_id` to `make_addresses_for_postcode`. Does that fix the "`NoneType`" error you posted above? (I still can't re-produce)

chris48s:
Aside from the issue regarding postcodes with only one address which we talked about on Slack, I've tried running all the existing districts import scripts and hit some more errors.

I've got
`import_conwy` and
`import_south_cambridge`
throwing
`pollingstations.models.MultipleObjectsReturned: get() returned more than one PollingDistrict -- it returned 2!`

and

`import_calderdale`
throwing
`django.contrib.gis.geos.error.GEOSException: Error encountered on GEOS C predicate function "GEOSContains_r".`

Can you reproduce those? Not sure what the issues are there. Maybe some special cases we need to handle (first one might just be adding a `try/except` block). Alternatively, if time is short and we know there's problems there but not how to solve them, we could exclude data from those areas as a quick fix.

chris48s:
I've done a quick picture check on a few local auths. What I've done is grabbed all the UPRNs from the ResidentialAddress table, attached grid refs and plotted them on top of the district boundaries. This won't flag up specific issues, or UPRNs we're not including but should be. However, if what we're doing is roughly sane then what we should see is clusters of points around the district boundaries with no/very few points in the centre of districts..

Here's Camden:
camden

Here's West Berks:
westberks

...and I found this little extract from Vale of Glamorgan particularly pleasing! :D (is this just me?)
vale


Vale of Glamorgan actually has quite a lot of these problems, so it would be really helpful to get this one updated!


So yeah - that looks pretty sensible. Probably doing more good than harm, excepting the outstanding issues raised above.


Moved on to start looking testing via the web front end, but I am getting `ImportError: No module named 'dc_theme.context_processors'` on every page now. I think that's related to PR #309 rather than this though. I'd guess that since PR #309 was merged there is some kind of additional compilation step or dependency I am now missing? Can you advise?

Cheers

symroe:
> pollingstations.models.MultipleObjectsReturned: get() returned more than one PollingDistrict -- it returned 2!

[...]

> Can you reproduce those? Not sure what the issues are there. Maybe some special cases we need to handle (first one might just be adding a try/except block). Alternatively, if time is short and we know there's problems there but not how to solve them, we could exclude data from those areas as a quick fix.

On this, I've looked at South Cams and it looks like it's coming from bad shapes:

screenshot 2016-06-19 11 16 59

You can see the top and bottom districts overlap, and there is an address point that throws that exception right in the middle of the two.

Seeing as we have no way of knowing what the correct polling station is here, I think it's best to ignore these addresses.

chris48s:
> Seeing as we have no way of knowing what the correct polling station is here, I think it's best to ignore these addresses.

Agreed - in a situation where the user's property is covered by >1 district, the correct behaviour should be to show "we don't know - call your council"

symroe:
> import_calderdale
> throwing
> django.contrib.gis.geos.error.GEOSException: Error encountered on GEOS C predicate function "GEOSContains_r".

Looks like this is an existing problem with the Calderdale data that was exposed by looking up more points in the area than we had before. Fixed by adding the same shape fix from the Cardiff data.

symroe:
> I am getting ImportError: No module named 'dc_theme.context_processors' on every page now. I think that's related to PR #309 rather than this though. I'd guess that since PR #309 was merged there is some kind of additional compilation step or dependency I am now missing? Can you advise?

Have you done a `pip install -U -r requirements/local.txt`?

polling_stations/apps/addressbase/helpers.py
```diff
@@ -0,0 +1,90 @@
...
+            district = PollingDistrict.objects.get(
+                area__covers=address.location, council_id=council_id)
+        except PollingDistrict.DoesNotExist:
+            # Chances are this is on the edge of the council area, and
+            # we don't have data for the are the property is in
+            continue
+        except PollingDistrict.MultipleObjectsReturned:
+            # This is normally causes by districts the overlap
+            # Because we have no way of knowing what the correct district is,
+            # we need to ignore this address
```
chris48s: OK, so what this is going to do is just exclude the one problematic address (i.e: if there are n addresses covered by this postcode, and one sits on an overlap, we insert n-1 addresses). This isn't **wrong** as such, but if you are the person who lives at that address, its a bit confusing as there is no clear "what do I do next" action for them at this point. One really hacky "fix" would be to insert that address with an intentionally invalid station id reference (which I think should send them to "we don't know - call your council"). Alternatively it could be dealt with in a more generic fashion with the "my address is not in the list" button proposed in https://github.com/DemocracyClub/UK-Polling-Stations/pull/326#issuecomment-226967544 (which also solves issues like "my address is newbuild and was registered with Royal Mail/Highways Agency since the last AddressBase was published, etc)
Probably not a blocker for this PR though.

chris48s:
> Looks like this is an existing problem with the Calderdale data that was exposed by looking up more points in the area than we had before. Fixed by adding the same shape fix from the Cardiff data.

Sorted! :)

chris48s:
> Have you done a `pip install -U -r requirements/local.txt`?

Hmm.. maybe I have only updated my base requirements. I'll give it a go when I'm in front of the relevant PC. Thanks

polling_stations/apps/addressbase/helpers.py
symroe: I prefer the 'address not listed' option – I'll make an issue for it.

chris48s:
OK then.. done some more tests. Calderdale, South Cambs and Conwy all complete now, so that's good.

I ran Cardiff and that is throwing `'NoneType' object has no attribute 'internal_council_id'`. I guess I missed that one the other day - oops! If you can reproduce, my **guess** would be that given the polygons were dodgy there is a gap between 2 polygons and we've got a point that falls in the gap (i.e: the inverse of the problem with South Cambs observed above). That's just conjecture though, so needs checking.

I had another look at the postcodes with only one address, and nothing seems to have changed there. All the import scripts are still inserting the same number of `ResidentialAddress` records. For example, E1 6HB, SE1 1NL, SE1 9BG and EC1A 4JQ are all exhibiting the same issue I noted the other day. Looking into these a little deeper, they are all postcodes which cross over the edge of the neighbouring local authority boundary. This probably still needs dealing with somehow, ideally.

Everything else I've checked seems fine, so apart from those 2 issues above, I think I've stopped finding things wrong with it at this point.

chris48s:
Thinking about it, I'd be interested to see what happens with SE1 1NL and SE1 9BG in your database because you'll have Southwark imported as well: these 2 postcodes have most postcodes in Southwark and one in City of London (and hence the centroid is in Southwark). In my database I haven't imported Southwark, so both of those postcodes just automatically send me to a polling station in City of London. Obviously that's not that useful because we've moved from an output that is right more often than it is wrong (based on centroid) to an output which is wrong more often than it is right (based on one address being in the `ResidentialAddress` table) :(

I have no idea if the behaviour you will observe will be:
1. You get shown a list containing all the addresses in Southwark, but not the one in City of London
2. You get sent straight to a polling station in Southwark
3. You get a list of all the containing all the addresses (regardless of whether they are in Southwark or City of London) and get sent to the appropriate polling station for each.
Obviously the third would be the desirable behaviour - interested to see which it is though.

Perhaps its more useful to consider an example where we don't have the neighbouring local auth imported (South Cambs or Calderdale or somewhere like that) because the ideal behaviour in this situation would be that for a postcode like this we get a list of all addresses with that postcode and the ones that aren't inside a local auth we have coverage for send us to the "we don't know - call you council" page for the neighbouring council we don't hold data for. Although predictably, that's the most difficult to achieve.

symroe:
Hmm, both postcodes give City Of London polling stations :/ I'll have to look in to this properly

symroe:
Ok, actually, that was because I'd not re-run the City of London importer since the addressbase stuff.

Now I get the following:

## SE1 9BG

Learning Centre
Atrizan Street Library and Community Centre
1 Artizan Street
E1 7AE

City of London Corporation council.

This is the wrong council, but looking at it, 'SE1 9BG' does't appear in the data from Southwark. Looking at AddressBase, each address is a business.  I'm still not sure why we're getting this false positive, but in this case it doesn't actually matter.

## SE1 1NL

Address picker

screenshot 2016-06-19 21 11 34

The top two addresses go to the same polling stations, and the last one is a business address.


Considering these, two, I should have filtered out non-residential addresses when i imported AdderessBase, but I think that can be left for future, as the worst case is just some odd addresses being suggested.

symroe:
> 'NoneType' object has no attribute 'internal_council_id'

I reproduced it, and fixed it in 877ee2a

chris48s:
OK cool, so in the case where a postcode is split across 2 local authorities and we do have the neighbouring authority imported it is fine (we get addresses in the list from both auths - the desired behaviour), but we have a bug where we don't have the neighbouring authority, or the scope of the data is different (e.g: we've got business addresses in one or not the other). Your latest response is helpful in clarifying this.

Having thought this through properly, this is actually no worse than the situation where one local authority has provided a ResidentialAddress dataset which we've imported and we have no data for the next-door authority - it creates the same problem.

So, for example:
* A postcode exists XX1 1XX which contains 51 addresses: 1 is in council A and 50 are in council B.
* With no data imported for councils A or B, a search for XX1 1XX says "we don't know - call council B"
* If we import data for council A and we have no data for council B, searches for that postcode now say "go to polling station X in council A" instead of "we don't know - call council B" on the strength of one address.
* This is the case whether we have imported an address dataset for council A or (now) a shapes dataset - it has just never occurred to me until I started looking for issues in this.
* What probably **should** happen is the user gets a list of 51 addresses, 50 of which map to "we don't know - call council B" and the other one maps to "go to polling station X in council A"

Does that make sense/sound reasonable?

I think fixing this is an AddressBase-related improvement that can/should be made, but given its not an issue inherent to what has been done here (it also applies everywhere we have imported a ResidentialAddress dataset), the best solution is to say it is out-of-scope for this PR, merge this (everything else is now in order as far as I can see :+1: ) and raise a new issue for it. Once we've got an AddressBase app in the project, the groundwork is there to make it easier to sort out.

Does that sound reasonable?

side note: fair play - you are smashing the London import scripts at the moment. Think you've submitted 2 new PRs in the time it took me to write that!! :fire:

symroe:
Ok, reckon I'll merge and leave the other bits for another day…

PR #321, opened 2016-06-16
Performance improvements

symroe:
Various commits that reduce the number/code of DB queries.

Each commit should explain itself…


polling_stations/apps/pollingstations/models.py
```diff
@@ -83,14 +83,15 @@ def get_polling_station_by_id(self, internal_council_id, council_id):
...
 
 class PollingStation(models.Model):
-    council             = models.ForeignKey(Council, null=True)
-    internal_council_id = models.CharField(blank=True, max_length=100)
-    postcode            = models.CharField(blank=True, null=True, max_length=100)
-    address             = models.TextField(blank=True, null=True)
-    location            = models.PointField(null=True, blank=True)
+    council = models.ForeignKey(Council, null=True, db_index=True)
+    internal_council_id = models.CharField(
+        blank=True, max_length=100, db_index=True)
```
chris48s: I originally added single-column indexes like this when I was looking at performance improvements in PR #260 but I removed them because I decided they didn't do anything (except slow down the imports slightly by forcing an index rebuild after each insert). Any sensible query we are running should be of the form `...WHERE council_id='X01000000' AND internal_council_id='AA'` or whatever, which should be hits on the composite indexes added in f1cc609. If we run a query of the form `...WHERE internal_council_id='AA'`, that may be a bug. Does this appear to be speeding any queries up?

polling_stations/settings/base.py
```diff
@@ -185,6 +185,8 @@
 
 INSTALLED_APPS += PROJECT_APPS
 
+SESSION_ENGINE = "django.contrib.sessions.backends.signed_cookies"
+
```
chris48s: `base.py` already defines a session engine on line 109. If you're changing the session engine, remove that declaration for clarity

symroe: Thanks, caught!

polling_stations/apps/pollingstations/models.py
symroe: I did speed up queries on the polling station page, yes, but only by a little bit. You're right that it will never be a huge query, but I think for the time being the advantage of speeding up reads in anyway at all overweighs the disadvantage of slower imports. Can we keep it in for now and re-consider later?

symroe: Maybe we need to generic way to drop indexes for importers

symroe:
Reckon this is good enough – The index isn't making anything worse, so let's keep it for now

PR #330, opened 2016-06-17
Enable ManifestStaticFileStorage for fingerprinted assets for better cacheability

ashb:
This means we end up with static urls like 'css/styles.55e7cbb9ba48.css', and when the content changes the URL changes. This lets us cache all static assets for years without risk of serving an old version to a client.

The only restirction is that you have to use the `{% static "path" %}` function which we were doing almost everywhere anyway.

This will need DemocracyClub/dc_base_theme#10 to work fully


ashb:
I might have broken editing/auto rebuilding of the base theme when installed via `pip install -e` mode. I'll check

ashb:
I'm unable to see any difference in behaviour between libsass and scss compilers (which makes some sense as libsass subclasses the scss compiler) but I'm not sure exactly what behaviour I'm looking for.

symroe:
This looks good to me, and we can always figure out how to fix any local re-compiling issues later.

PR #339, opened 2016-06-20
Multidb

symroe:


symroe:
For deployment, the main thing here is to add the `logger` DB and the `DATABASE_ROUTERS` list to the settings (using `local.py`).

PR #340, opened 2016-06-20
Remove link to WhoCanIVoteFor in sidebar

andylolz:
WCIVF isn’t relevant for the EU ref, and shouldn’t be mentioned.


symroe:
Ace, thank you!

PR #347, opened 2016-07-16
Post-EU Ref text edits

chris48s:


chris48s:
I've added a little [checklist](https://github.com/DemocracyClub/UK-Polling-Stations/wiki/Pre-election-tasks) to the wiki for next time..

PR #348, opened 2016-07-16
Add 'my address is not in the list' button to address select

### LINKED ISSUE #332: Add "my address is not in the list" to address picker
As per this comment https://github.com/DemocracyClub/UK-Polling-Stations/pull/308#discussion_r67615450

chris48s:
Note: this has some text which will need translating for the NUS Wales site
Closes #332

### DIFF
--- a/polling_stations/api.py
+++ b/polling_stations/api.py
@@ -172,7 +172,7 @@ class PostcodeViewSet(viewsets.ViewSet, LogLookUpMixin):
         log_data['council'] = council
         log_data['brand'] = 'api'
         log_data['language'] = ''
-        self.log_postcode(postcode, log_data)
+        self.log_postcode(postcode, log_data, 'api')
 
         return Response(ret)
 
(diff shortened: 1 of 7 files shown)


symroe:
This is a good UI change, but I think we should log something when this button is pressed. I expect it will be a good indicator of problems in data provided by councils, or out of date AddressBase data from us.

chris48s:
Agreed. I've set it up so it will now log the view that was used to render the respose, so we will log one of `PostcodeView`, `WeDontKnowView`, `AddressView` or `api` against each log record.

One further point on this:
When I was testing this I found a couple of times I selected an address then clicked 'my address is not in the list'. I'm not sure if this is because the buttons aren't laid out in the most sensible way, or just because I was running through a bunch of different test cases quickly.

If you get a chance to do any UX testing at any point, it might be worth reviewing the layout of those buttons:
- Do users find this confusing?
- Are the buttons the right way round?
- Should one be left-aligned and one right-aligned?
- Should the buttons be different sizes or colours?

I think I've got too close to it to make a judgement.

PR #358, opened 2016-08-17
Change default MAPIT_URL to http://mapit.democracyclub.org.uk/

chris48s:
Anonymous usage of Mysociety's Mapit is now rate-limited to 10 calls per minute: http://mapit.mysociety.org/account/signup/

I've proposed switching the default to DC's mapit as this is what I'm using in dev at the moment. Possibly we want to add the ability to specify a mapit key? Can discuss when we catch up..

Just to give you a quick update on the bigger picture, I know I've been quiet for a bit, but I've got some major changes to the base import classes in progress - just not ready for a PR yet. I am on the case though :)

### DIFF
--- a/polling_stations/apps/data_collection/constants.py
+++ b/polling_stations/apps/data_collection/constants.py
@@ -2,7 +2,7 @@ import os
 
 from django.conf import settings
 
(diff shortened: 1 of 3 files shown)


symroe:
I think they've whitelisted requests from us, but this is a reasonable change anyway – we don't want to hammer their servers on election days if we don't need to.

PR #359, opened 2016-08-21
Restructure base import classes

chris48s:
Right, first off apologies: This PR does far more than it is reasonable to do in one pull request, the diff is going to be tough to read (as far as `/polling_stations/apps/data_collection/management/commands/__init__.py` goes, it is probably easier to just look at https://github.com/chris48s/UK-Polling-Stations/blob/dcf9f5daac691886729bb2c8595264f24f709216/polling_stations/apps/data_collection/management/commands/__init__.py rather than try to make sense of the diff) and the description of what I've changed is a small novel. Unfortunately in order to make a lot of the improvements that needed doing, I had to do a lot of refactoring and then the 2 processes became completely intertwined - sorry.

On the other hand, I've made quite a lot of improvements :)

## Improvements to process of writing import scripts:
* ResidentialAddress import scripts now run an order of magnitude faster - we are now bulk inserting 3,000 records at once instead of running insert statements one at a time. This makes a huge difference.
* `--verbosity`/`-v` parameter allows us to log some additional debug info which will hopefully be helpful when writing import scripts. Default is `-v 1` Running with `-v 2` or `-v 3` outputs verbose or super-verbose debug info. Running with `-v 0` suppresses the data quality report and I've used this to not output the data quality report when running unit tests to clean up the travis builds a bit.
* `--noclean`/`-n` parameter disables the process that uses AddressBase to identify postcodes overlapping district boundaries. This is the slowest part of running a districts+stations import script and is not necessary if we're doing something like tweaking address formatting. Using this param shortens the feedback loop between making changes and seeing results. Obviously this is off by default.

## Improvements to code structure:
The old base classes had poor separation of concerns. File IO code, database IO code and core application logic relating to processing the polling station/district/address data was very tightly coupled. I've restructured to decouple these concerns. This does several useful things:
* Less duplication: Previously code for each file format defined its' own methods for processing the station/district/address data - now all the importers share a single `import_polling_stations()` / `import_polling_districts()` / `import_residential_addresses()` method, which should be easier to maintain.
* Code is more consistent: Because each format-specific importer was defining its own method for processing the station/district/address data, there were a number of cases where they worked a bit differently for no real reason. Maybe we had added code for dealing with some edge case to one importer but not another because we happened to have encountered it when processing a KML file, but not a SHP file (for example). Now because all the format-specific importers share the same code for dealing with station/district/address data, they all behave the same way and account for the same special cases.
* It is now much easier to define base import classes if we encounter new file formats or new combinations of file formats. For example, if we needed to define an importer to deal with Stations in SHP format and Districts in GeoJSON format, we can write:

  ```
  class BaseShpStationsJsonDistrictsImporter(BaseStationsDistrictsImporter):

      stations_filetype = 'shp'
      districts_filetype = 'json'
  ```

  ...and we're done (ignore the example given in the commit message for da7a65e - I simplified it some more later). To add a completely new file format we need to also define a new file helper in `/polling_stations/apps/data_collection/filehelpers.py` and add a case for it to `FileHelperFactory.create()`. Either way, it is a lot simpler.

## Handling edge cases:
* All `_to_dict()` methods can now optionally return `None` if we need to exclude a station/district/address (this was one of those things that was previously implemented in some base importers but not others).
* `station_record_to_dict()` may optionally return a list of dicts. This is helpful if we encounter a polling station record with a delimited list of polling districts served by this polling station: it allows us to add the same station address/point many times with different district ids.

## Improvements to handling of ambiguous addresses:
* Instead of putting known bad data into the database and then deleting it again, we now identify and remove the ambiguous records before inserting them - this seems like a more sensible way round to do it. It might also cut down on database IO a bit in some cases, but any improvement is fairly marginal.
* This process is now tolerant of formatting differences. The previous process was not tolerant of formatting differences and would have considered the following to be 3 different addresses:
  * 5-6 Mickleton Dr, Southport, PR82QX
  * 5/6, Mickleton Dr.  Southport, PR82QX
  * 5-6 mickleton dr southport, PR82QX

  Now these will be considered identical when identifying/excluding ambiguous address data.

## Data integrity:
Attempting to insert a duplicate on `(council_id, internal_council_id)` in the `PollingStation` or `PollingDistrict` model now raises an `IntegrityError`. Previously we were relying on `update_or_create()` to implicitly de-dupe and just overwrite with whatever the last record encountered was. In the case where there are duplicates on `(council_id, internal_council_id)` we now need to explicitly handle this in an appropriate manner in our import scripts. In some cases (where we've got exact dupes), we will need to essentially replicate this behaviour, but making this change highlights some situations where use of `update_or_create()` introduced bugs/inaccuracies. There are various examples of how to handle this in f00b2de and 081fc44.

## Other changes:
* Added a bunch of new unit tests covering new stuff I've added and things we just didn't previously have tests for.
* Used a more consistent naming convention for base classes
* Switched to DC mapit by default (see PR #358). I've got the same 2 commits on here because I needed to do that in order to test stuff properly, but I can delete those if you want to treat that as a separate issue.
* I've updated all the old import scripts. This was mostly about being able to test changes to the base classes with concrete implementations and real data rather than because I think we will need to import the old data again, but they do also serve as useful examples of how to write import scripts/deal with various conditions in the context of this new structure.

One final thing to note is that this totally breaks the crowdsourcing stuff I started (PR #268). I'm going to leave the PR open for the moment just to remind me it is there. I'll update it if crowdsourcing becomes an option we want to pursue again in future.

Again, sorry there's so much going on in one PR! :(


polling_stations/apps/data_collection/management/commands/__init__.py
```diff
@@ -40,159 +38,209 @@
 from addressbase.helpers import create_address_records_for_council
 
 
-class CsvHelper:
+class Slugger:
```
symroe: Can this have some tests? Just to ensure it's dealing with the "1/2" vs "12" issue properly, and future ones we might come across. Seems like a high chance of hard to initially spot regression in future versions otherwise.

symroe:
Thanks for this! Comments are going to be as hard as reading, so I think I'll just add them here. It's a shame GitHub doesn't support threaded comments…

symroe:
`StationList`/`DistrictList`/`AddressList`

1. I think these should have a generic base class as they're more or less the same at their core. 
2. Why are you defining the `_raw` and `_db` lists on the class as well as in `__init__`?
3. Should the `_raw` and `_db` lists be sets? I guess we'd need to make sure the `__eq__` method actually worked properly on them though.
4. Should we return something from `save()`? Like the number of inserted rows? I can't think of a reason for this at the moment, so just an open question.
5. Should we allow passing the `batch_size` param through to `bulk_create` in `save`?
6. I _think_ it's better to use the newer style `class Foo(object):` syntax for magic reasons I don't really understand.

chris48s:
Cheers for having a look :) Comments all seem reasonable. It has been a couple of weeks since I did anything on this now, so I'll get back into it over the weekend and update/respond as applicable. Ta

symroe:
File structure:

I've never really liked putting a load of logic in an `__init__`, and now it's got a lot of other helper classes in, I think we should move it somewhere else. The `data_collection` app isn't very well named, so maybe we could rename it to `data_importers` or something, and then have some well named files in the root dir of that app that the importers can use.

This is especially the case for classes like `Slugger`, that don't really need to be in the same file as the `BaseImporter ` classes. Maybe I'm just being too picky? This is absolutely not something that should block this PR, and it might be easier to wait until it's merged to make this change…

chris48s:
> Can this have some tests...

I didn't add a specific test for `Slugger` because I was being lazy and some of the integration tests contain tests cases which implicitly cover it: For example, if you break `slugify`, `test_address_import` should fail. However, it is useful to make the expected behaviour of this totally explicit so I have added tests for it in 9e2b43a.



> I think these should have a generic base class as they're more or less the same at their core.

Good plan - done in 475d348.



> Why are you defining the _raw and _db lists on the class as well as in `__init__`?

There is no sensible reason for this - they should only be defined in `__init__`. I've fixed that in 7038fee. There's also some other places in the codebase where I've (incorrectly) done the same thing e.g: https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_collection/data_quality_report.py#L31
I'll have a look at these separately to avoid pulling more unrelated changes into a PR which is already a bit cumbersome.



> Should we return something from `save()`? Like the number of inserted rows? I can't think of a reason for this at the moment, so just an open question.

The data quality report already reports this (and a whole bunch of other stuff) so I'm not sure it adds any value.



> Should we allow passing the `batch_size` param through to `bulk_create` in `save`

Don't see why not. I've added a command line param for it in b0031d3.



> I *think* it's better to use the newer style `class Foo(object):` syntax for magic reasons I don't really understand.

I hadn't really come across this before so I did a bit of reading. In python 3, `class Foo(object):` and `class Foo:` are exactly the same - they're only different in python 2. In a project which only supports python 3 (like this one), it doesn't make any difference which syntax is used, but it is an important distinction if you're writing python 2 code or a library which supports both python 2 and 3.
If you want to pick one syntax or another to use as a stylistic convention for consistency purposes, that would probably be sensible.

In the existing codebase there is already inconsistent use of syntax, for example:
* `class Foo:` https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_collection/google_geocoding_api_wrapper.py#L17
* `class Foo():` https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_finder/helpers.py#L102
* `class Foo(object):` https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_finder/views.py#L38

I propose we pick a notation and then make all of that consistent as a separate task to avoid changing more unrelated content in this PR.



> I've never really liked putting a load of logic in an `__init__`...

Yep - on board with this. I had already done a bit of this by moving some stuff out into `filehelpers.py` and `loghelper.py`, but I've finished the job in ef3ff1c, e097e9a and 0249246.



> The `data_collection` app isn't very well named, so maybe we could rename it to `data_importers` or something...

I don't disagree with this in principle, but renaming a django app is a bit of a pain. This [stackoverflow post](http://stackoverflow.com/questions/8408046/how-to-change-the-name-of-a-django-app/8408131#8408131) claims to give all the steps, but it is missing at least one thing which is that you have to manually edit all your migration files because those reference your app name too. Personally I'd tend to avoid this unless we wanted to create another app which was more deserving of the name `data_collection` or split this up into several smaller apps or something that *requires* a rename. Either way, lets treat it as a separate issue.



> Should the `_raw` and `_db` lists be sets? I guess we'd need to make sure the `__eq__` method actually worked properly on them though.

Hmm.. you've got me thinking now..
The simple answer is I didn't do it that way because you can only put hashable types in a set and a dict is not hashable: Attempting to add a dict to a set throws `TypeError: unhashable type: 'dict'`.
An alternative approach would appear be to store a set of `PollingStation`/`PollingDistrict`/`ResidentialAddress` objects and define the necessary `__eq__()`/`__hash__()` methods on those classes: I think this is what you're proposing.
This also doesn't work. Django model objects are natively hashable if they've already been persisted to the database: 2 model objects are considered equal if they are of the same type and have the same primary key. If they haven't been persisted yet, they are unhashable. Attempting to add one will throw `TypeError: Model instances without primary key value are unhashable`.
The reason for this is that hashable types are supposed to be immutable - it shouldn't be possible to put an object in a set and then change that object (by reference), because then you might change it so it doesn't belong in the set. Dicts and django model objects (that haven't been persisted yet) are both mutable.
I don't think would be a sensible idea to try and define a custom `__hash__()` method in our model classes for this reason. It may not even work - I've not tried it.
I think to make this approach work, we need to declare an intermediate immutable hashable datatype for each 'entity' (station/district/address), build them from dicts, store them in a set and then convert from the intermediate format to model objects in `save()`. Although that does add some complexity in one sense, I *think* this would actually solve some problems elsewhere. For example, it would probably make some of the things I've done in 081fc44 unnecessary - although I've not actually fully thought through every single example.
I've not had a go a this yet - its taken me a while to think it through this far, but it does seem like a good idea so leave the PR open for now and leave it with me..

chris48s:
Data structures for stations/districts/addresses are now implemented as sets of named tuples rather than lists of dicts.

I think that probably covers everything now, unless you have any further feedback or disagree with anything in the post above.

Cheers

PR #369, opened 2016-10-06
Use logger DB for data_collection app

### LINKED ISSUE #357: League table doesn't work on live site
https://wheredoivote.co.uk/league_table/

This is another one I can't reproduce locally

symroe:
Fixes #357

### DIFF
--- a/polling_stations/db_routers.py
+++ b/polling_stations/db_routers.py
@@ -1,12 +1,17 @@
+APPS_THAT_USE_LOGGER = [
+    'data_finder',
+    'data_collection',
+]
+
 class LoggerRouter(object):
 
     def db_for_read(self, model, **hints):
-        if model._meta.app_label == 'data_finder':
+        if model._meta.app_label in APPS_THAT_USE_LOGGER :
             return 'logger'
         return None
 
     def db_for_write(self, model, **hints):
-        if model._meta.app_label == 'data_finder':
+        if model._meta.app_label in APPS_THAT_USE_LOGGER:
             return 'logger'
         return None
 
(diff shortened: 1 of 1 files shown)


symroe:
Actually, although the change in this PR is the right solution, it won't actually produce good results at the moment, as the logger DB isn't updated when we import new data.

This is a different issue though, so I think this PR stands.

chris48s:
Sorry - I'm still a bit confused about how this problem has occurred and the proposed solution. Is the reason this is happening in prod because there are multiple back-ends and the writes are only getting applied to one of the DB servers?

chris48s:
Just to add a bit more info based on comment in issue #357

> How are the DataQuality objects created? Is it ad import time?

The actual records are created using this post migrate hook:
https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_collection/app.py

The `teardown` command and import scripts just clear/update fields in records which should already exist in the table.

symroe:
Sorry I wasn't clear:

"Prod" is one of a `n` servers at any given time. So that we scale well, each server takes a 'golden' copy of the DB with all the data in it, and then uses that local copy to power itself.

This works because the data is all read only, apart from the logging functions. Hence the DB router for a single shared DB that each instance accesses.

I expect the data quality repot got mixed up somehow on #EURef day when we did some custom importing on each instance that was running.

Because the importers don't run on the instances, the table will never get re-written to the DB. The 'logger' DB is a good place for this data, as we can update it once when a new golden image is made.

Thinking about this more though: I expect the instances will just remove the logger DB version if an ad-hoc import is done on an instance. I might need to think about this a bit more TBH!

chris48s:

Issue #356, opened 2016-07-24
Write more import scripts for APIs/open data


chris48s:
Developing issue #175 a bit further, we've now got a few import scripts which obtain data directly from APIs. This is useful for several reasons:
- It is easy to update the data if it changes - we just re-run the script. We aren't relying on councils to notify us of changes.
- We can front-load the work: import scripts can be written now and run later, rather than having to be written in the frantic few weeks leading up to an election.
- If a council publishes their data via an API, we can just harvest their data. We don't need to find a co-operative contact within that council to help us, which can sometimes be difficult.
- We know the data will be in a consistent format.
- Now that we have the AddressBase fixes applied in PR #308, we don't have to worry about the accuracy issues associated with using district shapes.
- It rewards/promotes councils who are actively publishing open data about polling districts/stations - presumably they are doing this because they want people to use it.

We should be doing more of this.

There are a variety of councils who are actively publishing open data via APIs/open data portals. For example:
- http://datagm.org.uk/dataset/salford-polling-stations
- http://dataworks.calderdale.gov.uk/dataset/polling-stations
- https://data.yorkopendata.org/dataset/polling-stations

Additionally it is possible to discover council WFS/ArcGIS APIs serving up this information by either searching for terms like `allinurl: gov uk getcapabilities`, or looking at polling station finders on council websites to see where they pull their data in from. In some cases this data is explicitly open. In some it may be better described as not-closed data, but it is nothing that can't be found using a web browser, a search engine and a copy of firebug :)

I've collated as many links as I can find in this google doc:
https://docs.google.com/spreadsheets/d/1RDmyxF4hQS0PcSahHCARBXfqLA1v-PAnTWE6cYnRfW8
- Quite a few of these serve up district and station data, which will allow us to increase our base level of coverage.
- Some of them only provide station data only - this could be useful if we can get hold of address to district mappings from the electoral roll or district shapes from the ordnance survey.
- Some of them only provide district shapes - this could be useful if combined with PDFs of polling station addresses.

The one issue with this is that in some cases (particularly where I've found an API that isn't really publicised), it is diffuclt to know if the data is actively maintained/updated. I have an idea for this though: If we set a scraper on [morph.io](http://morph.io/) to run on a schedule for each one and then pull the data from there. That would tell us when the data changes so we can work out if any are stale and retire them. I'll have an experement with this..

chris48s:
[...]

PR #370, opened 2016-10-08
Add BaseMorphApiImporter class

chris48s:
# What is this for?

As discussed in issue #356, I've found a bunch of APIs/open data portals serving up data on polling stations and districts. To try and find out how up-to-date/stale this data is, I've written a bunch of [scrapers](https://morph.io/wdiv-scrapers) that run every day, poll the apis to detect changes and also pull the data into morph because:
* It allows us to do a bit of pre-processing/getting the data into a more consistent format
* Yay - open data for everyone!

This PR adds an base class for importing this scraped data from morph.

# API Key

In order to download data from morph, you need an API key. If you sign in to morph with GitHub it will give you a key for your github account and any orgs your are a part of. You can use your own, or the DC org's key, but I don't think it really makes any difference.

In order to run the import scripts can either set an env var `MORPH_API_KEY` or add `MORPH_API_KEY = 'foobarbaz'` to `local.py`

# Import scripts

I've started adding a few import scripts mainly just to test/provide examples. A few notes:
* I realise some of the import scripts I've done are for areas that aren't having elections in 2017, but doing import scripts for City of London and Ealing allowed me to A/B test the result of the new morph-based importers against the result of running the importers that fetch data directly from the API. Guildford and Salford are relevant for May 2017
* I currently have no idea how up-to-date this data is/will be. Some of these scripts may be importing stale data, but part of the purpose of this is to answer that question and allow us to retire importers which are fetching data from stale APIs/prod the relevant councils
* I'm aware that the next plan is to rip all this out and move it to https://github.com/DemocracyClub/UK-Polling-Station-Data-Processing but this gives us a base to get @hishivshah working on something fairly accessible as a contributor in the meantime
* For now I have tagged the scripts with whatever election/s they are next relevant to or were last relevant to. e.g: `['local.surrey.2017-05-04']`, ['gla.c.2016-05-05', 'gla.a.2016-05-05', 'mayor.london.2016-05-05', 'ref.2016-06-23'] but obviously the aim is that these scripts should apply to any election that is relevant to this area (assuming we find that the APIs are maintained). It would be nice if we didn't have to re-tag them every election, but lets deal with that as another issue.

Refs issue #356

### DIFF
--- a/polling_stations/apps/data_collection/base_importers.py
+++ b/polling_stations/apps/data_collection/base_importers.py
@@ -14,9 +14,11 @@ from argparse import ArgumentTypeError
 from django.core.management.base import BaseCommand
 from django.conf import settings
 from django.contrib.gis import geos
+from django.contrib.gis.gdal import DataSource
 from django.contrib.gis.geos import Point, GEOSGeometry
 from django.db import connection
 from django.db import transaction
+from django.utils.encoding import force_bytes
 
 from councils.models import Council
 from data_collection.data_types import (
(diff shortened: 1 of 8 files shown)


symroe:
Really nice!

Not to block this getting merged, but it might be nice to consider using the ['checks' system](https://docs.djangoproject.com/en/1.10/topics/checks/) for warning about missing keys like `MORPH_API_KEY`. We're getting a few of them, and missing keys might not get picked up in tests or prod builds.

Issue #303, opened 2016-06-09
Move data to it's own repo


symroe:
There is quite a lot of data in this repo now, making it slower to checkout the code itself. Also we might want to make a private repo with some of the data (I'm not sure about this, but it would be nice to have the option).

It's possible git isn't even the correct storage medium for this data anyway! Maybe we should look at some other sort of hosting.

JoeMitchell:
Good points!

chris48s:
I have ideas about this - lets discuss next time we do a hangout

PR #373, opened 2016-10-15
Add ability to import data from amazon S3

chris48s:
Putting this here so you can see where I've got to with it but not ready to merge yet.

For the moment, I've added the ability to grab data from S3 alongside the existing options for importing from data in the repo or a local private dir without 'disturbing' that logic too much. However this is still WIP - there's a few things I'd like to dicuss before finalising/tidying up so not ready to merge yet.

Rough topics are:
- Should we take this opportunity to re-organise the structure of the data, or just replicate the existing structure?
- How do we enable people other than use (i.e: who don't have an account on the DC S3 bucket) to write import scripts?
- Do we want to retain the ability to import data held locally?
- How much do we need to retain backwards compatibility for old import scripts?
- Also having started splitting stuff out into website/importers I've hit a couple of things I hadn't anticipated - hence targeting this at the main website repo for now.

..since we've got a hangout scheduled next week, lets review and chuck some ideas about and then I'll be able to finish it off..

Refs #303

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,6 @@ polling_stations/settings/local.py
 
 # editor cruft
 *~
+
+# cached local files from amazon s3
+s3cache/*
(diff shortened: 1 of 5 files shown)


chris48s:
WIP tag removed - this is ready for review now.

chris48s:
OK. I've modified that so it expects the keys from a `.boto` file or `boto.cfg` or whatever as that seems to be the convention for where to define this. I've also rebased onto master to sort out the merge conflict on the settings file.

symroe:
Nice! 👍

Issue #367, opened 2016-10-06
List of things we think need UX testing


symroe:
Not comprehensive, but a starting point for anyone who wants to test specific elements of the UX:
## 'my address is not in the list' button (address picker page)

@chris48s said: 

> When I was testing this I found a couple of times I selected an address then clicked 'my address is not in the list'. I'm not sure if this is because the buttons aren't laid out in the most sensible way, or just because I was running through a bunch of different test cases quickly.
> 
> If you get a chance to do any UX testing at any point, it might be worth reviewing the layout of those buttons:
> - Do users find this confusing?
> - Are the buttons the right way round?
> - Should one be left-aligned and one right-aligned?
> - Should the buttons be different sizes or colours?

symroe:
Lots about addresses here https://designpatterns.hackpad.com/Addresses-CgrMSGRAhRc#:h=Examples-on-GOV.UK

mtsk3000:
hi guys (my first response!)

Done a bit of UX review and think there are a couple of quick things we can do to tighten it up a bit

![image](https://cloud.githubusercontent.com/assets/20727101/25562566/f714d5fa-2d80-11e7-89a3-ff72b19ad8e6.png)

The key thing is the lack of information hierarchy. We got 2 big main call-to-actions here, while only one is relevant at a time (ie if I have not select my address, the "i have selected..." btn should be disabled/hidden). The current flow is that both btn are clickable, and if you click "find my polling station" without selecting an address, the site then handle the flow error by showing showing an error msg.

A potentially better solution could be this:
![image](https://cloud.githubusercontent.com/assets/20727101/25562579/abfe1a94-2d81-11e7-989d-2538974602a1.png)
Nothing selected

![image](https://cloud.githubusercontent.com/assets/20727101/25562582/d0e66cbc-2d81-11e7-885c-21444fa50b29.png)
Address selected - I am not entirely sure if the "X Remove selection" bit is a bit over-designing, but I think we should at least allow the user to click the selected option to unselect, that allows the user to access the "My address is not in the list" button again (not typical behaviour I know, but why now 😛 ). The tick icon just adds a bit more confident and positivity to the "find my polling station" button to make decision a bit easier. every little helps.

Hope that helps!

chris48s:
@mtsk3000 - Cheers for having a look at this for us. I've submitted PR #750 for review based on your ideas.

JoeMitchell:
[...]

PR #374, opened 2016-10-15
Make address_select view clearer on Android

chris48s:
This fixes issue #189 for Chrome on Android.

I used http://www.browsershots.net/ to generate some screenshots on various versions of Safari on iOS and it appears that this isn't an issue for iPhones because Safari on iOS displays this element in the same way as a desktop browser. It would be useful if someone with an actual iJobbie of some description could confirm that.

The reason I've added the css conditionally using javascript like this is you don't want to set `'-webkit-appearance:menulist;'` because this makes it look odd when webkit-based desktop browsers display the full-height version.

### DIFF
--- a/polling_stations/templates/address_select.html
+++ b/polling_stations/templates/address_select.html
@@ -29,5 +29,17 @@
   </div>
 </div>
 
-
 {% endblock content %}
+
+{% block in_page_javascript %}
+<script type="text/javascript">
+fallback.ready(['jQuery'], function() {
+    $(document).ready(function() {
+        var isAndroid = /(android)/i.test(navigator.userAgent);
+        if (isAndroid) {
+            $('.select_multirow').css('-webkit-appearance', 'menulist');
+        }
+    });
+});
+</script>
+{% endblock in_page_javascript %}


symroe:
I wonder if this might be better addressed by using `widget=forms.RadioSelect`?

There is some good research that select boxes generally aren't great (although to be fair, this is mostly at single select boxes, not multiple selects).

A `RadioSelect` widget has the advantage that we can control the HTML and style the output with a lot more control than relying on the browser/OS implementation of the widget.

You can see a nice example of this in the "Stacked radio buttons" section here: http://govuk-elements.herokuapp.com/form-elements/#form-radio-buttons

symroe:
Actually, that might be an answer to the 'my address is not in the list' point in #367: the 'or' section in the stacked example above might work really nicely.

chris48s:
This is similar to how chrome on android implements drop-downs as it is. Just to add some visual help to this, the issue before was that the drop down wasn't differentiated from the text box on the previous page (see screenshots in issue #189 for that).

So now there is a small arrow to help clarify:



but when you actually drop it down, chrome represents it as a radio list:




This does bring up an interesting question though.. this is a screenshot from browsershots of what that page allegedly looks like in Safari on an iPhone 6:

3a2951daa28ed0b8cbd6f68eaf626287

I wonder if that multi-row select is cumbersome to use on a touch device? I suspect it may be, although I'm not sure what happens next if you touch in the `<select>` element.


I used a very similar approach to the one you've linked to (radio buttons display horizontally if there is room but then stack vertically if at narrow resolutions) on an online survey system I worked on recently.

My gut instinct is that this approach works really well for short lists where each label is fairly short e.g:

* Yes
* No

or

* Northern Ireland
* Isle of Man or the Channel Islands
* I am a British citizen living abroad

(as in the examples shown)

but it may be more marginal for long lists where each label is long (e.g: the list of every address described by a postcode, which can be a lot longer than in the example I've shown). I quite like the fact that the button (i.e: the next call to action) is usually visible on first visual impression.

Shall we add this to the list in issue #367 with a particular focus on how best to present this on mobile devices as it is probably something that needs consideration.

symroe:
Ok, merging this pending some testing to see if there is a better interface. This is better than what we have at the moment, for sure :)

PR #400, opened 2016-11-20
Re-organise settings + use Django System Checks to check for missing/bad settings

### LINKED ISSUE #376: Use Django System check framework to warn about missing settings
See notes at https://github.com/DemocracyClub/UK-Polling-Stations/pull/370#issuecomment-255406091

PR #373 introduces some new settings, so lets deal with this once that is merged.

Also, we've got various app-specific constants/settings in the main settings file and a bunch more in `/polling_stations/apps/data_collection/constants.py` - some of which aren't specific to the data_collection app. We should also re-organise these into a more consistent layout.

chris48s:
This one requires review as it has the potential to break important stuff.

Changes in this PR:
* Re-organise settings:
  * Move all app constants into settings dir
  * Separate app constants from django settings
  * Organise app constants into logical groups
* Make use of Django System Checks Framework:
  * Ensure `MORPH_API_KEY` is set with system check framework instead of throwing `NotImplementedError()`
  * Add checks for other settings-related issues
  * Enforce them in sensible places

Intended behaviour should be:
* If running an import script which extends `BaseMorphApiImporter`, a halting `Error` should be raised if `MORPH_API_KEY` not set
* If running another import script (e.g: shp, kml, etc) or an unrelated task (e.g: `python manage.py runserver`) this check should be ignored
* If `TILE_LAYER == 'MapQuestSDK'` and `MQ_KEY is None` a halting `Error` should be raised when running `python manage.py runserver`, `python manage.py check` (this will break the maps), etc but management commands should ignore it
* If `MAPIT_UA is None` a non-halting `Info` should be raised when running `python manage.py runserver`, `python manage.py check`, etc (usage limits will apply, but this is fine for local dev) but management commands should ignore it
* If `GOOGLE_API_KEY == ''` a non-halting `Info` should be raised when running `python manage.py runserver`, `python manage.py check`, etc (usage limits will apply, but this is fine for local dev) but management commands should ignore it

If we want to define any global system checks which should be respected by all commands, we can define them in the `pollingstations` app, but I haven't defined any.

Closes #376


symroe:
This looks really nice! I'll have a proper read through, but initial thought: can we have `manage.py check` in travis please?

chris48s:
Yep - no probs.

Issue #412, opened 2016-11-26
Latest dc_base_theme breaks stuff


chris48s:
There are various places where we are relying on fallback.js as a loader.

We have been relying on the fact that this is pulled in by the base theme, but the latest version breaks this because [this commit](https://github.com/DemocracyClub/dc_base_theme/commit/2d0f601778c440c045ca9d070089d97e0296c02f#diff-6a7d249a60db1a8de8890953efd8d4a8L13) removes it.

Should we:
* Put fallback.js back in the base theme?
* Load fallback.js in the wheredoivote template instead?

This may also affect other sites (e.d: whocanivotefor) if we've used the same approach there.

symroe:
Right, sorry about that. First off, we should version the DC base theme
in requirements.

The reason for removing fallback.js is that I switched to
django-pipeline, and that concats the JS files in to a single request,
as well as minifying them.

Pipeline doesn't expose the filenames to be comperssed, meaning we can't
write the filename out in the fallback.js block in the base theme – only
the whole tag(s).

In order to keep this and fallback.js, I had to remove the fallback code
from dc base theme and add this ugly hack to the project's base theme:



```html
{% block extra_javascript %}
{% if debug %}
    <script>
    // shim fallback in debug mode
    if (typeof fallback == 'undefined') {
        var fallback = {
            ready: function(l, f) {
                if (typeof l == "function") {
                    l()
                } else {
                    f();
                }
            }
        }
    }
    </script>
    {% javascript 'scripts' %}
{% else %}
    <script src="{% static "dc_theme/js/fallback.min.js" %}"></script>
    <script type="text/javascript">
        fallback.load({
            scripts: [
                '{% static "js/scripts.js" %}'
            ],
        });
    </script>
{% endif %}
{% endblock extra_javascript %}
```

The reason for shimming fallback in debug mode is because in-page
scripts need to be wrapped inside a fallback.ready block, so they work
when debug is off.

I expect more of this should be moved to the base theme, but it will
always require an update to project's base.html template, I expect.

PR #416, opened 2016-12-09
Improve acceptance tests

chris48s:
making sure build fails on travis.. tbc


chris48s:
OK, so based on the comments in issue #412, what I've done is:

* Added a check to the acceptance tests which will cause the build to fail if javascript errors are logged to the console
* Fixed it by pinning dc_base_theme to last good commit (`04efca5047b31d705e63d516b3b5dc5d0f0694a7`) pending a solution

This will help us pick up any errors caused by upgrading the theme in future and also flag other possible problems

chris48s:
Also while looking at the acceptance tests, I've added an extra one for the "my address is not in the list" case

Issue #342, opened 2016-06-20
Add option to open external links in a new window on embedded version


symroe:

chris48s:
Fixed in PR  #417

PR #417, opened 2016-12-11
open external links in new tab

chris48s:
Open external links in new tab

I've done this in response to issue #342 where it is particularly important because it is really annoying if external links open inside an iframe.
That said, I think the correct behaviour should be that regardless of embedded/non-embedded, links to external sites (google maps, electoral commission, facebook, etc) should open in a new tab anyway (as we're already doing with the "Read more about voting in the UK" link). However, this is not a completely clear-cut issue - not everyone has the same preference.

Reckons?

### DIFF
--- a/polling_stations/templates/base.html
+++ b/polling_stations/templates/base.html
@@ -37,13 +37,13 @@
       <p>Democracy Club is a UK based Community Interest Company that builds
         the digital infrastructure needed for a 21st century democracy</p>
       <p class="inline-list social">
-        <a href="https://www.facebook.com/democracyclub/">
+        <a href="https://www.facebook.com/democracyclub/" target="_top">
             <i class="fa fa-facebook-square" aria-hidden="true" style="color: white"></i>
         </a>
-        <a href="https://twitter.com/democlub">
+        <a href="https://twitter.com/democlub" target="_top">
             <i class="fa fa-twitter-square" aria-hidden="true" style="color: white"></i>
         </a>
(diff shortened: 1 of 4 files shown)


JoeMitchell:
>"links to external sites (google maps, electoral commission, facebook,
etc) should open in a new tab anyway"

Yep, I totally expect this. But I googled it and the internet is in
disagreement. Gah. One for the user-testing.

symroe:
I think this should happen for iFrames, as I agree it's annoying to have the page open inside a window. Maybe we should set `target="_parent"` though, to force opening in the parent window?

If we do that, we should only do it for embedded versions, not each version.

I generally think it's a bad idea to open links in a new window unless there is a clear case for doing so. For example, in the case where going back to the previous page might lose state (like when editing a document).

Changing domains doesn't feel like enough of a reason to me.

chris48s:
As noted on Slack - agree with your point. On reflection I've decided the correct target for external links should actually be "_top" rather than "_parent".
Reason for this is that the site we're embedded in will usually be implemented using a Content Management System of some kind. Sometimes (crap) CMSs will use `<iframe>`s in their internal implementation so in some cases we may actually be in an `<iframe>` inside another `<iframe>`. Using "_top" means we'll break out of N levels of iframe-inception any use the parent window/tab rather than the parent `<iframe>`.
We'll need to remember this convention when adding external links in future.

chris48s:
Also closes #523

PR #425, opened 2016-12-18
Import script for Ashfield

chris48s:
See comment at top of file for notes


symroe:
How hard would it be to convert the file types as part of the import? I guess it would add a load of dependancies to the project?

Issue #404, opened 2016-11-21
Do we need a 'last-updated' notice?


JoeMitchell:
I was emailing a council person to tell them 'not to worry, we don't assume the data is accurate for more than one election'. But in fact, the Cardiff data, e.g., is all still up there. And if there was a by-election that we're not aware of, it might not longer be accurate.

Do we need something like the attached. (But probably better designed. :D) 
![your_polling_station___wheredoivote_](https://cloud.githubusercontent.com/assets/8096305/20487675/3d186290-affc-11e6-94f6-397a95c96d22.png)

chris48s:
Would it seem reasonable to shove a similar message to the one on https://whocanivotefor.co.uk/ on the front page?

JoeMitchell:
I think so — there's a (small) risk we're getting hits from folks with
by-elections receiving data still lying around from 2016, and theoretically
the correct data could have changed versus what we have.

And we do say that we don't use data for more than one election.

andylolz:
I think the current message on https://whocanivotefor.co.uk is good. How would you change it, @chris48s?

Re. the proposed change here: can I suggest “your poll card” rather than “a poll card”, and a date e.g. “last updated on 30 May 2016” rather than “last updated for the [EU referendum]”.

chris48s:
@andylolz - Nothing too dramtic. Just substitute "elections in May 2016" for "referendum in June 2016", *possibly* remove the link to candidates site(?) but keep the blog link. I'll knock up a quick PR taking into account discussion in this issue..

andylolz:
> Nothing too dramtic. Just substitute "elections in May 2016" for "referendum in June 2016"

Well… The site doesn’t appear to contain any information. So neither of those is correct, really.

> _possibly_ remove the link to candidates site(?)

Searching for a postcode on [the candidates site](https://candidates.democracyclub.org.uk) currently takes you back to whocanivotefor, which contains no information. So yeah, either remove links to it, or change this behaviour so it is actually a May 2016 archive.

chris48s:
https://wheredoivote.co.uk/ - sorted

andylolz:
That’s https://wheredoivote.co.uk, not https://whocanivotefor.co.uk. Were we talking at cross purposes the whole time?

In any case, +1 good fix.

PR #426, opened 2016-12-19
add archived site notice

chris48s:
Based on conversation in #404 - does this seem ok?


chris48s:
Also added a note to remove it in the [pre-election tasks](https://github.com/DemocracyClub/UK-Polling-Stations/wiki/Pre-election-tasks)

symroe:
Thanks, and deployed!

PR #427, opened 2016-12-26
WIP: Crowdsourcing (3rd time lucky)

chris48s:
We're getting the band back together.. again.... again! Lets see if we can make it work this time.

This all kind of roughly builds on stuff previously done or discussed in
https://github.com/DemocracyClub/UK-Polling-Stations/issues/267#issuecomment-218831211
https://github.com/DemocracyClub/UK-Polling-Stations/pull/268
but some of it has changed because the data has changed, or because the codebase has changed or just because I came up with a better (or at least differently terrible?) idea..

# Progress:
- [ ] Scripts to clean up and import electoral roll
- [x] Script to generate CSV templates
- [x] Script to import crowdsourced data

# Notes:

## Importing Electoral Roll

I've deliberately not done anything about importing Electoral Roll data (yet). Based on the 3 example files we've got, it is clear that we're going to get a bunch of different files in different formats and there's going to be a variety of edge cases to deal with (pretty much everything we see in ResidentialAddress import scripts is going to occur somewhere.. and the rest). I think I'm going to wait until we've got the full data in and then write the cleanup/import scripts rather than trying to write them predictively.

What I **have** done is decided that we shouldn't import it straight into the ResidentialAddress table. What I actually propose is that we stage the Electoral Roll data in its' own table and then bring it over to the ResidentialAddess table as part of the import process. The web app itself never directly accesses the ResidentialAddess table. This is kind of 'wasteful'(?), but I think it simplifies things because:
* CSV templates can be generated with the database in any state - nothing needs to be done in a particular order and we don't need to know which councils will/won't provide data yet, or manually maintain that list
* Whatever mix of open data/data from councils/crowdsourced data we get is easy to accommodate at the end of the process when it is close to election day
* Once we've run all our import scripts, the ResidentialAddress table only contains addresses relevant to areas where we have some polling station coverage

## Generating CSVs

* Once we've dealt with importing electoral roll addresses, we can generate CSV templates for collecting crowdsourced data using the script provided: `python manage.py generate_csvs /path/for/export`
* Then we can:
  * Whack them on google docs
  * Mobilise wombles
  * Assemble data 

..then we can move on to...

## Importing data

* CSVs can be imported directly from a URL (probably google docs) or from local storage. Usage examples:
  * `python manage.py import_crowdsourced_csv X01000001 -f /path/to/data/X01000001.csv`
  * `python manage.py import_crowdsourced_csv X01000001 -f /path/to/data/X01000001.csv -c 'latin-1'`
  * `python manage.py import_crowdsourced_csv X01000001 -u 'https://docs.google.com/spreadsheet/ccc?key=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx&output=csv'`

* Split districts and blank rows are handled by excluding them. `WARNING` level notification is thrown in these cases.
* Grid refs are attached using postcode centroid if possible - I've assumed we won't have gridrefs for crowdsourced data. Maybe there should be optional gridref/srid columns? I think it is unlikely that PDFs will include these, but there may be some special cases..

## ...and finally

* We probably need a bit of discussion around process, but most of https://github.com/DemocracyClub/UK-Polling-Stations/issues/267#issuecomment-218831211 probably still holds.
* I had a mince pie, I promise!

### DIFF
--- a/polling_stations/apps/data_collection/base_importers.py
+++ b/polling_stations/apps/data_collection/base_importers.py
@@ -181,12 +181,24 @@ class BaseImporter(BaseCommand, PostProcessingMixin, metaclass=abc.ABCMeta):
             apps.get_app_config('pollingstations')
         ])
 
+        """
+        Shove args and kwargs into some object properties. This means if
+        another class which extends BaseImporter defines additional custom
+        args, we don't have to override handle() to access them.
+        """
+        self.args = args
+        self.kwargs = kwargs
+
         verbosity = kwargs.get('verbosity')
         self.logger = LogHelper(verbosity)
         self.batch_size = kwargs.get('batch_size')
 
         if self.council_id is None:
-            self.council_id = args[0]
+            if 'council_id' in kwargs:
+                self.council_id = kwargs['council_id']
+            else:
+                self.council_id = args[0]
+            # if args is empty, give up and let it throw an IndexError
 
         self.council = self.get_council(self.council_id)
 
--- a/polling_stations/apps/data_collection/management/commands/generate_csvs.py
+++ b/polling_stations/apps/data_collection/management/commands/generate_csvs.py
@@ -1,0 +1,74 @@
+"""
+Generate CSV templates for collecting
+crowd sourced polling station data
+"""
+
+import csv, glob, os
+from django.core.management.base import BaseCommand
+from pollingstations.models import ElectoralRoll
+
+
+class Command(BaseCommand):
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'output_path',
+            help="""Path to export CSVs to e.g:
+            '/home/user/data/elections.2017-05-04/'
+            (must be an empty dir)
+            """
+        )
+
+    def generate_files(self):
+        # grab councils we want to build a csv for
+        councils = ElectoralRoll.objects\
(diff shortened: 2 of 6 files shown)


symroe:
Not looked much, and don't want to throw a spanner in the works, but should
we think about spinning the crowd sourcing aspects off in to its own
project that feeds data in to WDIV as if it were any other data source?

Advantages: keeping WDIV simple and read only.
Disadvantages: another code base to maintain

chris48s:
Depends what the ambitions for it are..

If we want to go with the 'minimum viable crowdsourcing' google docs/CSVs based approach, I reckon its ok for a handful of django management commands to live in the main repo supporting that (which is basically what I'm doing here).

If we want to do something more elaborate and build up a website around it (like you've done with the candidates crowdsourcing), then yeah it should probably live in a separate project. I'd tend to agree that we don't want to bolt a crowdsourcing app onto wheredoivote.co.uk.

There's definitely tradeoffs to be made here, but I'm not exactly sure which side I come down on at the moment..

symroe:
Right, yeah sounds like we agree. I'd not seen that this was basically a
few management commands. I'll take a proper look tomorrow.

chris48s:

Issue #423, opened 2016-12-17
RFC: Thoughts on improving the API


chris48s:
As discussed in the hangout the other night, here's some thoughts on improving the API. For the moment, I'm just focussing on the `/postcode` endpoint flow. There's probably also stuff that needs to happen with the `/pollingstations` and `/pollingdistricts` endpoints (particularly districts - I think the `/pollingdistricts` endpoint in its current form is pretty useless), but there's already an absolute wall of text to read here, so we'll save that for another issue. 

# TL;DR version

* Should a valid response always parse into a consistent object at the top-level, or is it OK for the top-level keys of the object to vary in different cases?
* Error responses should definitely return a suitable HTTP status code, not `200 OK` as they do at the moment.
* Should valid responses include `"error": false` so that can always be checked?
* Should we add a `/address` endpoint so that we can call https://wheredoivote.co.uk/api/address/1-foo-street-bar-town.json (for example)
  OR..
* Should this be dealt with by modifying the `/pollingstations` endpoint so that we can call something like https://wheredoivote.co.uk/api/pollingstations.json?council_id=X01000001&polling_station_id=FOO ?
* Is it actually sensible to maintain a distinct API, or would it be better to simply expose the controller layer of the application as json?
* Should we version the API?

# All the gory details

OK, so at the moment, if I call https://wheredoivote.co.uk/api/postcode/CF105AJ.json (polling station known from postcode) I get back:
```json
{
  "postcode_location": {
    "latitude": "51.46518039304887",
    "longitude": "-3.1657202682046206"
  },
  "polling_station_known": true,
  "polling_station": {
    "council": "https://wheredoivote.co.uk/api/councils/W06000015/",
    "postcode": "",
    "address": "Portacabin, County Hall, County Hall, Atlantic Wharf, Butetown, Cardiff",
    "location": {
      "latitude": "51.469852462940835",
      "longitude": "-3.162477184225377"
    }
  },
  "council": {
    "council_id": "W06000015",
    "council_type": "UTA",
    "mapit_id": "2639",
    "name": "Cardiff Council",
    "email": "electoralservices@cardiff.gov.uk",
    "phone": "029 2087 2034",
    "website": "https://www.cardiff.gov.uk/ENG/Your-Council/Voting-and-elections/pages/default.aspx",
    "postcode": "CF10 4UW",
[...]

PR #368, opened 2016-10-06
Remove bootstrap_theme block

symroe:
This fixes prod API pages, as the static files finder can't find the assets linked to

Fixes #352

PR #430, opened 2017-01-02
API Improvements

chris48s:
# Summary

We need to build a better API. Based on general discussion, comments in issue #423, comments on slack and a fairly large helping of my own opinions and preferences, I've had a go at making some improvements.
Some of the changes in this PR are fairly uncontroversial and just fix stuff that was broken or add features we needed. Some of the changes are more 'opinionated' and would benefit from feedback, criticism and comment.
This PR contains quite a lot of breaking changes. Client code on WCIVF will need to be updated to account for this. As noted, while the API is labelled as alpha and we are the only people really using it, this is fine. Once we've nailed this down and it is fairly stable, I reckon we should move this from `/api/` to `/api/v1/` (or something) and version future non-BC changes to avoid pulling the rug out from under users.

# Supported calls
The api now supports the following calls:

(edited)
see https://github.com/DemocracyClub/UK-Polling-Stations/pull/430#issuecomment-272703508

# Stuff I have done

The following changes should probably be non-controversial (I think):

* Return correct HTTP response codes on error
* Make error responses consistent with standard Django Rest Framework errors
* Add `/address` endpoint returning responses matching `/postcode` + link it from `addresses` array elements in `/postcode` endpoint
* Return consistent keys in object from `/postcode` and `/address` endpoints and formalise that output spec in a `Serializer` class (this may break BC if you're checking `if 'addresses' in response` or similar)
* Add optional geojson endpoints for `/councils`, `/pollingstations`, `/pollingdistricts` (but keep non-geo outputs for performance reasons when we don't need the boundaries, just the meta-data)
* Add ability to filter `/pollingstations` and `/pollingdistricts` by council id and station/district id (trying to filter by station/district id only will throw `HTTP/1.0 400 BAD REQUEST`)
* Link everything up properly for better discoverability/HATEOAS-compliance
* Include station object in districts output using `get_polling_station()` logic (or `null` if we have an orphan district). It is worth noting I have deliberately excluded `polling_station_id` and `polling_district_id` fields as the link is implicit so we can abstract it
* Various bits of refactoring and tidying

These following changes prompt questions or require further discussion:

## Removed 'location' field from council object

Didn't get any feedback on this in issue #423, but it is just based on an example postcode somewhere in the local auth. I don't think it is a useful API output so I've got rid of it. If you disagree, let me know.

## Pagination

Calling `/pollingstations` or `/pollingdistricts` with no filters now returns a paginated response. This means we can make calls like `/api/pollingstations/?limit=100&offset=200`. When calling with a filter like `/api/pollingstations.json?council_id=X01000001`, pagination is not used - it will just return all of them. I've done this because:

* HTML outputs stay responsive/useful which helps with discoverability
* People can't tie up our server with a query that says 'give me boundaries for all polling 
districts in the country' (we definitely don't want this), but I think 'give me boundaries for all polling districts in a local authority' is perfectly reasonable for a single request to make the API actually useful.

The downside of this is that it sacrifices some consistency. A call to `/api/pollingdistricts/` returns:

```
{
    "count": n,
    "next": "http://127.0.0.1:8000/api/pollingdistricts/?limit=100&offset=100",
    "previous": null,
    "results": [
        /*array of district objects*/
    ]
}
```

whereas a call to `/api/pollingstations.json?council_id=X01000001` returns simply:

```
[
    /*array of district objects*/
]
```

Not sure if you reckon this is a problem or not? The other possible solution is just to have a call to `/api/pollingdistricts/` (with no filters) return `HTTP/1.0 400 BAD REQUEST`. This may actually be a beter solution, but in its current form, `/api` links to `/api/pollingdistricts/` and `/api/pollingstations/` so probably presents bad initial impression but this can probably be overcome with good documentation (discussed elsewhere). Let me know your thoughts.

## Representation of point objects in `/postcode` and `/address` endpoints

We've talked about this a bit on Slack already. I've chewed over a few different options and on balance, I think embedding geojson features in non-geo json (using the `PollingStationGeoSerializer` class) probably represents the right tradeoff for these 2 endpoints. Let me know if you want to discuss other options/more info.

## Representation of station objects in `/pollingdistricts/geo` call

Again, I've chewed over a few options for this and I reckon embedding just the meta-data with a link to the geo (using the `PollingStationDataSerializer` class) is the right option for this endpoint. Using the `PollingStationGoeSerializer` class to embed a geojson feature in the `'properties'` field of another geojson feature is valid geojson but that object will be parsed as a foreign member rather than a feature by most parsers, but there may be other options. Let me know if you want to discuss/more info.

## `/geo.json` routes

Ideally I'd rather have set this up so that (for example) `/pollingstations.json` (or `content-type=application/json`) is a different output to `/pollingstations.geojson` (or `content-type=application/vnd.geo+json`), but Django Rest Framework doesn't make it terribly easy to accept different content types/extensions on some endpoints but not others (note `/postcode`, `/address` do not have corresponding `/geo` routes). I think to do this you either need to define all your routes individually instead of using the built-in `DefaultRouter`/`SimpleRouter` objects or define some mess of custom `Parser`/`Renderer` classes to handle the content negotiation.
This seemed a reasonable option, but I might revisit if you think this is a particularly terrible solution.

## Other...things?

I've tried to be fairly comprehensive here, but there are bound to be other things I haven't thought of.

# Stuff I haven't done
* ~Documentation: I want to put together some users docs to go with this, but I'd like to nail down the above points first. We may be able to use tools to auto-generate a spec, but we probably also need a more bespoke user guide.~
* ~Tests: We should have tests for this, but I think it is useful to start getting some feedback on the above at this point in case I am way off the mark.~
* Fix HTML outputs (Should be handled seperately under PR #368).
* Logging for `/councils`, `/pollingdistricts`, `/pollingstations`: Hits on `/postcode` and `/address` are logged in the `LoggedPostcode` model. As noted in https://github.com/DemocracyClub/UK-Polling-Stations/issues/423#issuecomment-269655156 it would be nice to also log hits on `/councils`, `/pollingdistricts`, `/pollingstations` but they don't fit into this model as they aren't based on a postcode search.
* BBOX filter. I think it should be possible to do this (see https://github.com/djangonauts/django-rest-framework-gis#inbboxfilter ) but I'd like to try and find out a bit more about @mvl22 's project as there may be a more appropriate approach.
* ~More testing: I want to do a bit more testing to see if I can spot any more weird edge cases, but this is a good point to start getting feedback.~


chris48s:
Right.. made a few more mods to the `/pollingstations` and `/pollingdistricts` endpoints to make a call like `/api/pollingdistricts.json?council_id=<council_id>&district_id=<district_id>` behave like a 'get by id' query (which is what we're using it for) query rather than a 'filter' query.
I've also moved the api base route to `/api/v1` with the intention that the api docs will live at `/api`.

Based on the conversation we had on slack about docs, there are tradeoffs associated with all the approaches, but I think on balance API Blueprint probably solves more problems than it creates from my perspective. What I'm going to do is:
* Write the docs for the API as it stands in this PR using apibp and add them to this PR.
* Given the amount of stuff above that still requires review/feedback, there are bound to be additional changes needed to what I have done (and hence impact on the WIP docs).
* As I update the code and docs together, this will help to establish the process and how much of a problem it is to keep the docs and code in sync.

chris48s:
Added first bash at the API docs. As noted on Slack, theme needs more customisation but that can come later following updates to `dc_base_theme`.

What I've done for now is committed both the markdown source and the compiled HTML to the repo, but probably a better approach would be to just commit the source markdown here and add the command to compile the docs to the deploy process. Need to have a chat about how the deploy script(s) work though.

I'm using [aglio](https://github.com/danielgtaylor/aglio) to compile the docs. To install it `sudo npm install -g aglio` and then to compile the docs to html we run `aglio --theme-variables polling_stations/templates/api_docs/theme/dc.less -i polling_stations/templates/api_docs/blueprints/wheredoivote.apibp -o polling_stations/templates/api_docs/html/docs.html`. Aglio will throw a couple of warnings because I've described multiple `200 OK` responses for the `/postcode` endpoint because I think it is important to properly document and explain the different responses for:
* polling station found
* polling station not found
* call to `/address` needed to find polling station

in order for the docs to be meaningful.

There is a `--quiet` flag coming to supress warnings coming in a future version: https://github.com/danielgtaylor/aglio/pull/312

One note on this: I haven't bothered documenting the paginated responses for `/pollingstations` and `/pollingdistricts`. For the moment I've only documented the filtered versions (i.e: with a `council_id` parameter). The paginated responses are only really there for 'discovery' rather than because they are actually useful. Thinking more about it, I reckon if proper docs like this sit at `/api`, the 'discovery' aspect becomes less important and I'm starting to lean more towards those endpoints with no param just throwing a `400 Bad Request` and requiring at least a `council_id` filter, but I would welcome feedback on this point (as well as all the others in the first post).

Finally, it would be useful to get some review/feedback on the docs. I probably haven't got everything exactly right first time. The `/address` and `/postcode` endpoints probably need the most review as these calls require a lot more explanation than: "this endpoint slaps the data you requested in a JSON object and returns `200 OK` or you screwed up and it throws an error". Simultaneously, the fact that we can provide the necessary explanation demonstrates the usefulness of this format when compared to auto-generating the docs.

chris48s:
oh yeah.. also now I've moved the api base to `/api/v1`, the endpoints are now

* `/api/v1/postcode/<postcode>.json`
* `/api/v1/address/<slug>.json`
* `/api/v1/councils.json`
* `/api/v1/councils/<council_id>.json`
* `/api/v1/councils/<council_id>/geo.json`
* `/api/v1/pollingdistricts.json`
* `/api/v1/pollingdistricts.json?council_id=<council_id>`
* `/api/v1/pollingdistricts.json?council_id=<council_id>&district_id=<district_id>`
* `/api/v1/pollingdistricts/geo.json`
* `/api/v1/pollingdistricts/geo.json?council_id=<council_id>`
* `/api/v1/pollingdistricts/geo.json?council_id=<council_id>&district_id=<district_id>`
* `/api/v1/pollingstations.json`
* `/api/v1/pollingstations.json?council_id=<council_id>`
* `/api/v1/pollingstations.json?council_id=<council_id>&station_id=<station_id>`
* `/api/v1/pollingstations/geo.json`
* `/api/v1/pollingstations/geo.json?council_id=<council_id>`
* `/api/v1/pollingstations/geo.json?council_id=<council_id>&station_id=<station_id>`

chris48s:
now with tests

chris48s:
Removed 'WIP'. There's probably more changes required, but I think I've done everything I want to do before receiving comments at this point

symroe:
Thanks for this – it all looks excellent.

Quite hard to review, so I'll sort throw thoughts in this single comment:

* `/v1/.json` is a fairly odd URL for the root. It might be better to document that `/v1/` is the root and that clients can use content negation to change the output/

* On that – do we support any output other than JSON and HTML? What's the GeoJSON for XML or YAML or whatever? Do we not care about these? (I don't care about them).

* Not very important, but can we move al mentions of `wheredoivote.co.uk` to a setting? It looks like the markdown isn't passed through a template renderer before being compiled, so I expect not. It would be nice to have the docs use the correct domain when it's white labeled. Like I say, this is a very minor point.

* `addresses is an array of length one containing a single address object describing the requested address` – why a list of a single object, not the object?

* I've not spent a long time thinking about the `/pollingstations` and `/pollingdistricts` endpoints as we don't have current users with a need for these. Also, do you think we need to worry about people using these endpoints to clone the site and publish incorrect data (if we change things at the lst minute)?

symroe:
On the documentation, I think we should work on more of a narrative somewhere. The domain is deceptive, in that API consumers might not realise that there is more to it than a one to one between postcode <-> polling station.

For example custom finders are rare enough as to be forgotten by consumers. Also, the need to display an address picker in UIs might not be obvious at first.

I'm not sure how best to do this, maybe it's not something that should be in the API docs – maybe we should make a "reusing our data" page on the DC website that is suggested as required reading before consuming the API?

chris48s:
Cheers for providing feedback. As usual I will start off with the simple points where I can just answer your question and then move on to the questions where I am going to answer your question with a series of longer more complicated questions.

> `/v1/.json` is a fairly odd URL for the root. It might be better to document that `/v1/` is the root and that clients can use content negation to change the output/

Agreed. Let me have a go at doing something better with this, assuming we decide to keep it (see later discussion)

> Not very important, but can we move al mentions of `wheredoivote.co.uk` to a setting? It looks like the markdown isn't passed through a template renderer before being compiled, so I expect not. It would be nice to have the docs use the correct domain when it's white labeled. Like I say, this is a very minor point.

Not very easily, but from the perspective of embedding I'm not sure it is important because the `/api` route is defined in `extra_patterns` not `core_patterns` and only `core_patterns` are pulled into `whitelabel.urls`. Is there another situation other than embedding where this would be important?

> `addresses is an array of length one containing a single address object describing the requested address` – why a list of a single object, not the object?

I did it this way so that `/address` and `/postcode` return an object of the same format. See comments in #423 about consistent response format.

> On the documentation, I think we should work on more of a narrative somewhere. The domain is deceptive, in that API consumers might not realise that there is more to it than a one to one between postcode <-> polling station.

> For example custom finders are rare enough as to be forgotten by consumers. Also, the need to display an address picker in UIs might not be obvious at first.

I totally agree - this is *really* important for the 'finder' api. The ability to provide this sort of 'narrative' in the docs was one of the various reasons for picking APIBP for the docs. There's several things I've tried to do to try and clarify this:

* I've grouped the `/address` and `/postcode` endpoints into a 'Polling Station Finder' group in the docs and documented them together.
* I've made sure the opening sentence of that section is "The `/postcode` and `/address` endpoints find a user’s polling station based on their home address" to try and emphasise that both of these endpoints are necessary for a polling station search and the polling station is determined by the user's *address*, not ther *postcode*.
* The opening section of the 'Postcode Search' explanation clarifies that one of the 3 possible outcomes of a call to the `/postcode` endpoint is that we need to make a subsequent call to the `/address` endpoint

However, I obviously haven't done this well enough if it is still not clear.
Can you suggest anything else I can do here/propose any edits or additional narrative to improve this?

Finally, I'll note that if you call `/postcode` and should make a subsequent `/address` call but don't, we return `"polling_station_known": false` and provide council contact info. We don't serve the wrong polling station or multiple polling stations.

> I'm not sure how best to do this, maybe it's not something that should be in the API docs – maybe we should make a "reusing our data" page on the DC website that is suggested as required reading before consuming the API?

I think doing this as well is a good idea, but I think the docs should also perform this function as well as possible.

The next 2 points kind of address the same question in my mind.

> Do we support any output other than JSON and HTML? What's the GeoJSON for XML or YAML or whatever? Do we not care about these? (I don't care about them).

> I've not spent a long time thinking about the `/pollingstations` and `/pollingdistricts` endpoints as we don't have current users with a need for these.

The way I see it, the API endpoints (as proposed in this PR) band into 2 distinct groups:

### Polling Station Finder API

* `/address`
* `/postcode`

### Open Data API

* `/councils`
* `/pollingdistricts`
* `/pollingstations`

The Polling Station Finder API is for situations where a site/app other than wheredoivote wants to embed 'polling station finder' functionality. This covers how we are using the API on whocanivotefor, it covers (presumably) how the electoral commission are going to use it and potentially how some other data users might want to use it (media partners, councils who want to do a more custom integration than shoving our site in an iframe?). Fundamentally though, this exposes the wheredoivote 'user journey' as a **service**.

The Open Data API is more flexible, because it exposes the **data** behind wheredoivote. This opens up the data to a wider range of potential use cases.

If we stick to the 'find my polling station' use case, the `/pollingdistricts` endpoint (as I have designed it) would allow you to make a polling station finder where the user locates their house on a map (we've seen one that does this. I think it was Stroud?) and their station is determined by a point lookup rather than postcode/address.

It also potentially allows users to use the data for more esoteric purposes by using the data in bulk, for example:

* What is the largest distance a voter in council X has to travel to reach their polling station?
* What is the average distance a voter in council X has to travel to reach their polling station?
* My local council has recently conducted a review of polling districts. What impact has this had?
* How do councils currently assign codes to polling districts and what problems might need to be solved to establish a national coding system for polling districts?
* What kind of venues are used for polling stations? Are these appropriate venues?

..and probably better questions that I haven't thought of.
The finder API doesn't allow this sort of usage because it only exposes one very defined 'pathway' through the data.

I guess in order to answer your questions and take things forward, we need to answer more basic questions about what the objectives are for this API:

* Is this API fundamentally a tool that exposes our polling station finder as a reusable service?
* Is this API about publishing open data and making the data we have collected available for everyone to use (whatever form that might take) because "yay - open data and transparency"

The API I have designed addresses both cases, for 2 reasons:
* I think both objectives are valid.
* When I started working on the project, the API was branded 'Open Data API' and had 3 endpoints:
  * `/councils`
  * `/pollingdistricts`
  * `/pollingstations`
  so I've worked on making improvements from carrying forward that base position.

At the same time though, I'm not wedded to this: A lot about this project has changed over the last year or so - perhaps that should include the objectives of the API.

If what we really want to do is keep the bulk data more closed and expose the service, we can just completely bin `/councils`, `/pollingdistricts` and `/pollingstations` and strip the API down to `/postcode` and `/address`. We would also bin the root in this case as neither endpoint is 'discoverable' anyway as there is no 'list' action for either `/address` or `/postcode`.

Maybe we should recognise that these 2 groups of endpoints serve different purposes and split the api into `/api/v1/finder` and `/api/v1/opendata` or something?

> Also, do you think we need to worry about people using these endpoints to clone the site and publish incorrect data (if we change things at the lst minute)?

This is pretty much the exact justification some councils are using for not providing us with data. If we're going to adopt the position that councils should provide/publish their data for us anyway, is it legitimate to then use the same reasoning for not opening up the same data to our own users? Probably not worth addressing until we sort out the above - if we completely bin `/councils`, `/pollingdistricts` and `/pollingstations` I think it becomes and irrelevant discussion.

chris48s:
As discussed in Slack, I've moved the API to `/beta` and updated the docs accordingly.

Once this is merged, I'll rebase #368 onto these changes, make a few mods there. Then we can merge that to fix the HTML outputs and we can start piloting with some actual users to start hammering out the answers to some of these outstanding questions.

One thing to note is that once we're ready to have `/api/v1` and `/api/beta`, we'll need to restructure things a bit to support 2 parallel API codebases but we can leave that until it becomes necessary.

symroe:
💥

PR #433, opened 2017-01-06
make councils import script harvest Isles of Scilly

### LINKED ISSUE #386: Councils table is missing Isles of Scilly
https://wheredoivote.co.uk/postcode/TR220PL/
throws
`DoesNotExist: Council matching query does not exist`

http://mapit.mysociety.org/postcode/TR22%200PL.html

The mapit queries we are running in `import_councils` don't pick up `E06000053` because mapit lists Isles of Scilly under its own special type `COI` ( http://mapit.mysociety.org/areas/COI ), although I don't understand why..

Basically, we need to shove `COI` in here: https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_collection/constants.py#L10
and re-run the councils scraper on prod, but we probably want to either switch back to the MySociety mapit and add longer sleeps or fix the issues with Northern Ireland councils on the DC mapit first.

@symroe - thoughts?

chris48s:
Re-importing councils on prod with this change will close #386 but bear in mind caveat about issues with Northern Ireland councils on the DC mapit.

### DIFF
--- a/polling_stations/settings/constants/councils.py
+++ b/polling_stations/settings/constants/councils.py
@@ -8,4 +8,5 @@ COUNCIL_TYPES = [
     "MTD",
     "LGD",
     "UTA",
+    "COI",
 ]


chris48s:
refs https://sentry.io/democracy-club-gp/pollingstationsprod/issues/175174285/

PR #441, opened 2017-01-14
Fix council import

symroe:
Refs #386 

The importer is currently breaking on council id `95G` (Ballymena Borough Council). Digging about it seems that this is a NI council, that's being imported because "LGD" was in the COUNCIL_TYPES. See the comment on 2b1ff67.

Whilst I was there, I saw that we're doing some logic that just would never have worked, so I removed that, see 	ae80a17.

### DIFF
--- a/polling_stations/apps/councils/management/commands/import_councils.py
+++ b/polling_stations/apps/councils/management/commands/import_councils.py
@@ -79,9 +79,8 @@ class Command(BaseCommand):
             defaults = {
                 'name': council['name'],
             }
-            if defaults != "LGD":
-                defaults.update(
-                    self.get_contact_info_from_gov_uk(council_id))
+            defaults.update(
+                self.get_contact_info_from_gov_uk(council_id))
 
             defaults['council_type'] = council_type
             defaults['mapit_id']= mapit_id
--- a/polling_stations/settings/constants/councils.py
+++ b/polling_stations/settings/constants/councils.py
@@ -6,7 +6,6 @@ COUNCIL_TYPES = [
     "LBO",
     "DIS",
     "MTD",
-    "LGD",
     "UTA",
     "COI",
 ]


chris48s:
I've got an install where I didn't import the Northern Ireland councils to get round the issue with duplicate Northern Ireland councils on the DC mapit
e.g:
http://mapit.democracyclub.org.uk/area/23502.html
http://mapit.democracyclub.org.uk/area/24126.html
and searching for any NI postcode throws `DoesNotExist` `Council matching query does not exist`, rather than redirecting to the EONI finder, so I think there needs to be some way to get the codes and names in there or move things around to do the `CustomFinder` check before we try to look it up in the councils table.

There is already some code that accounts for the fact that the NI councils don't have any contact info added in https://github.com/DemocracyClub/UK-Polling-Stations/commit/afe119d1da90f231f7871018c16eac476cd4c550 which is what you get if you run the councils scraper using the mysociety mapit.

chris48s:
The other approach would be to fix the data on the DC mapit as I _think_ fundamentally this issue stems from the fact that there's 2 generations of NI data assigned to the same generation(?), but I'm not familiar enough with why/how that came to be or what else might depend on it to know exactly what is involved with changing that/what the consequences would be.

If http://mapit.democracyclub.org.uk/areas/LGD.html returned the same set of areas as http://mapit.mysociety.org/areas/LGD.html then the NI areas would import (albeit with no contact info - which is OK).

chris48s:
Tried re-running this with the EONI finder in the `customfinder` table and NI postcodes still fail with `Council matching query does not exist`, so definitely needs further fiddling.

symroe:
I've just demoted the old areas down to exist in generation 1 only, meaning that http://mapit.democracyclub.org.uk/areas/LGD.html is now the same as the mySociety version.

I think that fixes the issue, so I'll close this issue

JoeMitchell:
Hello! It's showing Manchester Borough Council where it should say Manchester City Council — is that related to this PR? (Assuming same problem as Coventry 'Borough' Council...)

PR #444, opened 2017-01-21
Import script for North Norfolk (+ some generalisation)

### LINKED ISSUE #443: Import E07000147-North Norfolk

chris48s:
Added import script for North Norfolk then realised I've written several import scripts recently which are basically all the same so I've added a `BaseXpressCsvImporter` class giving us a general case for data exported from Xpress electoral products. Base class might need a bit more tweaking as we see some more variations, but it is a decent base to start from.
I haven't bothered going back to the scripts from previous elections, but this should make things easier as we get more data of this form.

Closes #443


symroe:
Really nice!

Issue #372, opened 2016-10-13
don't show phone no as tel: link in browsers that don't support it


symroe:
Some browsers seem to think `tel:` URL is a relative URL rather than a protocol URL. Clicking on a `tel:` link then redirects to `/postcode/tel:12345`, raising the below error in helpers.

https://sentry.io/democracy-club-gp/pollingstationsprod/issues/168417308/

```
ValueError: Expecting value: line 4 column 1 (char 3)
(6 additional frame(s) were not displayed)
...
  File "data_finder/views.py", line 181, in get
    context = self.get_context_data(**kwargs)
  File "data_finder/views.py", line 132, in get_context_data
    l = self.get_location()
  File "data_finder/views.py", line 185, in get_location
    return geocode(self.postcode)
  File "data_finder/helpers.py", line 65, in geocode
    res_json = res.json()

ValueError: Expecting value: line 4 column 1 (char 3)
```

chris48s:
This now doesn't throw an error https://wheredoivote.co.uk/postcode/tel:12345/ .. so that's good

symroe:
Happy if you want to ignore this, but I think this issue isn't really fixed just by showing the user a better error page. The user intent is to call the council (or at least, that's the point of the `tel:` link). 

If the browser doesn't support the `tel:` link, we shouldn't show it (or, we should only add it if the browser does). I expect a good proxy for this is if the device is a mobile device.

I expect this is a JS progressive enhancement, with normal phone numbers being wrapped in a `<span class=tel>` or something and then `jQuery.browser.mobile` adding the `tel:` links)

chris48s:
Right, yep so there were/are 2 different issues here:

1. Most modern browsers know `tel:` is a protocol link and will either do something sensible with it or say something like `Firefox doesn't know how to open this address, because one of the following protocols (tel) isn't associated with any program` or whatever, but some (older) browsers don't know its a protocol link. If that happened, we were throwing `ValueError` trying to parse it, but now we're not. That is fixed.

2. Separately from that, regardless of whether the user's browser knows that `tel:` is a protocol link or not, it would be nicer to try and show/not show `tel:` based on whether we think it is going to be helpful or not on the user's device :+1:

It might be useful to do that by assuming it is on mobile and isn;t on desktop. We might be able to use JS to explicitly test if `tel:` has a registered protocol handler, which would be better.. if possible (I've not checked)

PR #447, opened 2017-01-22
Correctly handle mapit 404

chris48s:
If we try to feed something into mapit that doesn't match a route, the 404 mapit response is html not json (because it is Django's router throwing the 404 - none of the view code is ever invoked in this situation). Trying to parse it as json throws `ValueError`. This patch fixes that by re-throwing it as `PostcodeError`.

This also sort of addresses issue #372, in that if this happens as a result of requesting something like `/postcode/tel:029`, the user now sees

```
Sorry, we can't find tel:029

This doesn't appear to be a valid postcode.
```

instead of throwing an unhandled exception.

Looking at the info in Sentry ( https://sentry.io/democracy-club-gp/pollingstationsprod/issues/168417308/ ), 93% of these errors are from IE6 on Windows XP (I guess because IE6 just doesn't understand what to to with `tel:`) and I'm personally not *particularly* bothered about optimising for a 15 year old browser. Dunno if you want to say this also fixes issue #372 or not..? I'll leave it up to you

### DIFF
--- a/polling_stations/apps/data_finder/helpers.py
+++ b/polling_stations/apps/data_finder/helpers.py
@@ -58,35 +58,44 @@ def geocode(postcode):
 
     res = requests.get("%s/postcode/%s" % (settings.MAPIT_URL, postcode), headers=headers)
 
-    if res.status_code == 403:
-        # we hit MapIt's rate limit
-        raise RateLimitError("Mapit error 403: Rate limit exceeded")
+    if res.status_code != 200:
+        if res.status_code == 403:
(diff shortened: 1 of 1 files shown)


symroe:
(Sorry for the big red "changes requested" icon – I wanted to see what would happen if I pressed it!)

chris48s:
Yep good call. I hadn't accounted for other status codes with a non-json body.
This should do the job

> Sorry for the big red "changes requested" icon

No probs :x: :D

symroe:
Nice one :)

Issue #234, opened 2016-04-10
Improve directions/routing where we have property-level grid refs


chris48s:
Add an optional location column to the `ResidentialAddress` table, populate it where possible and use it as the start point for directions (in preference to the postcode centroid) for areas where we hold this data.

chris48s:
Steps involved in this:

* Add grid ref field to `ResidentialAddress` model
* Modify all the existing (applicable) `ResidentialAddress` import scripts to populate it
* Route all the `ResidentialAddress` requests through the `AddressFormView` for councils where we hold property-level grid refs (at the moment, we take a shortcut: if all the addresses in a postcode map to the same polling station, we just go straight to the polling station page. This might seem a bit silly in retrospect, but at the time we implemented this, there were no directions - they came later. This shortcut eliminated a ‘pointless’ user interaction).
* Update `DirectionsHelper` so it can use the property-level grid refs as the start point for directions.

Note:
Another shortcut I've taken with some of the `ResidentialAddress` data is I've accepted some ropey/incomplete addresses in situations where all of the addresses covered by one postcode map to one polling station (on the basis that nobody will ever need to select from that list of addresses). If/when we do this, that assumption/shortcut may need to be revised in some cases. Maybe an 'other addresses with this postcode' option that maps to the postcode centroid?

chris48s:
we do now have the data to to this in a lot of cases, but we don't currently show everyone an address picker (and may not want to)

GeoWill:
Some ROs from Scotland have asked about this. Would be a nice to have before May.

Issue #349, opened 2016-07-16
Make a better job of dealing with postcodes that sit across local authority boundaries


chris48s:
As described in
https://github.com/DemocracyClub/UK-Polling-Stations/pull/308#issuecomment-227020925
we have a bug if postcodes cross local auth boundaries in a situation where we import data for one council but not its neighbouring council.

Extrapolating that slightly further, we also sometimes show the wrong 'call your council' page in situations where we have no data for either due to the same reason. If a postcode contains addresses in 2 local auths, we just show the 'call your council' page for whichever one the postcode centroid is in.

GeoWill:
Closed in #2703

PR #448, opened 2017-01-25
Optimise ResidentialAddress queries

chris48s:
This is an optimisation that I actually thought of doing some time ago, but kind of mentally parked it because the changes in PR #359 were needed first. I've been thinking about performance if we import the electoral roll for most of the country and that has resurfaced the thought.

Here's the lowdown:
When we import a `ResidentialAddress` dataset, we currently import every address. However, in the majority of cases every address with a particular postcode is mapped to the same polling station. In these situations, we can collapse all of those addresses into a single record for that postcode and only store the individual addresses where the addresses described by a given postcode are split between >1 stations. By only storing the addresses we need, we massively reduce the search space for `ResidentialAddress` queries and improve performance.

This does add a bit more processing to the import process, but it doesn't slow it down by much.

As a happy side effect, this also means that where we have `ResidentialAddress` data, the page the user is directed to has a more meaningful slug because we are creating a postcode-level slug rather than just picking a random address with the same postcode the user entered (which is what we currently do).

As far as I can see, this doesn't create any additional problems that don't already exist but if you can think of any edge cases I haven't considered, let me know.

The only down side I can think of is it would prevent us from doing issue #234, but it isn't something we're doing at the moment so it doesn't break any existing functionality and we can always review the performance tradeoff if we decide we do want to add that.

### DIFF
--- a/polling_stations/apps/data_collection/data_types.py
+++ b/polling_stations/apps/data_collection/data_types.py
@@ -180,9 +180,58 @@ class AddressSet(CustomSet):
 
         return out_addresses
 
+    def build_postcode_lookup(self):
+        postcode_lookup = {}
+        for address in self.elements:
+            if address.postcode in postcode_lookup:
+                postcode_lookup[address.postcode].add(address.polling_station_id)
+            else:
+                postcode_lookup[address.postcode] = set([address.polling_station_id])
+
+        return postcode_lookup
+
+    def collapse_redundant_addresses(self):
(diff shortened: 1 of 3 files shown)


chris48s:
Actually now I've got into issue #349, this probably creates some edge cases around that too in its current form and may need some modification in that context.

Depending on how far we get into that before the May 2017 elections, I may re-target this at the `issue349` branch.

Park review of this PR for the moment and we'll see where we get to.

chris48s:

Issue #450, opened 2017-01-29
Import E07000152-East Northamptonshire


chris48s:
https://morph.io/wdiv-scrapers/DC-PollingStations-EastNorthamptonshire

chris48s:
https://github.com/DemocracyClub/UK-Polling-Stations/pull/452#issuecomment-286117299

chris48s:
missed the :boat:

PR #452, opened 2017-01-29
Import script for East Northamptonshire

chris48s:
As noted in slack, there are one or 2 issues for Joe to follow up with the council ;) but I think this is incorporates the right assumptions.

Refs #450 but leave it open pending feedback from East Northamptonshire council

### DIFF
--- a/polling_stations/apps/data_collection/management/commands/import_east_northants.py
+++ b/polling_stations/apps/data_collection/management/commands/import_east_northants.py
@@ -1,0 +1,78 @@
+from data_collection.morph_importer import BaseMorphApiImporter
+
+class Command(BaseMorphApiImporter):
+
+    srid = 27700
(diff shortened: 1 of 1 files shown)


JoeMitchell:
Good news I think from East Northants — sounds like they've cleared it up. I copied the notes into the Tracker for further info...

JoeMitchell:
Another attempt? A new link, at least... but possibly just the same problem as before? 

>We had to restructure our map service supplying the polling station data at East Northants  (we are in the middle of updating our GIS web services)

>The WFS will have changed due to a change in field names. Can you please check your system can still load in the data from the WFS URL below and let me know.
 
>https://mapsenc.east-northamptonshire.gov.uk/arcgis/services/elections/elections/MapServer/WFSServer?request=GetCapabilities&service=WFS

PR #456, opened 2017-02-04
Scraper report

chris48s:
Had this knocking about for a while now. There may be one or 2 kinks still to work out with scrapers that occasionally generate a false positive (i.e: reporting the data has changed when the content isn't fundamentally different) but I think at this stage it is probably useful to get it in the repo. It will provide useful info about which import scripts we need to re-run as we get closer to the elections in May.


chris48s:
Gah. This is now broken :( Hold off on review pending resolution of https://github.com/openaustralia/morph/issues/1125

chris48s:
That got sorted out fairly quickly, so this is now :+1:

PR #465, opened 2017-02-12
Deal with postcodes that cross local authority boundaries (parts 1 and 2)

chris48s:
refs #349

Important note: This PR is targeted at branch `issue349`. It should not be merged to `master` as it does not represent a completed working feature.

Doing this is going to be huge - it is a puzzle that has many pieces. The assumption that we can derive a local authority from a postcode is kind of implicitly baked into what we are doing at quite a low level. If wheredoivote is a house of cards:

```
    /\
   /\/\
  /\/\/\
 /\/\/\/\
/\/\/\/\/\
     ^ we need to change this bit
```

As such, this is going to need a lot of review/sanity checking. I'm aware that I've submitted several PRs recently which CHANGE ALL THE THINGS and were difficult to review, so I'm going to try and take a slightly different approach with this one. What I've done is created a new branch on the main repo `issue349`. I'll split the job down into some smaller bits, target a series of PRs at `issue349` and then once all the pieces of the puzzle fit together there, we can merge the completed working feature from `issue349` to `master`. Hopefully it will be easier to tackle the review in smaller chunks.

So, what are the first 2 pieces of the puzzle:

## Part 1: We need a way to attach GSS codes to UPRNs

I have handled this by adding a model to the `addressbase` app to store the [ONSAD](http://ons.maps.arcgis.com/home/search.html?q=ONS%20Address%20Directory&t=content) and written an import script to import it.

See commits 0a5a50a and ac5b986 for this.

## Part 2: Our geocoding functions need to be able to determine when a postcodes describes a set of UPRNs which exist in more than one local authority

In 103b07b, the key changes are:

* Add ability to identify council and attach GSS codes to postcodes using AddressBase/ONSAD
* Prioritise AddressBase/ONSAD as a source for geocoding/GSS codes
* Fall back to centroid-based geocoding using mapit if we can't sucessfully geocode using AddressBase/ONSAD (there are various situations where this may happen)
* Identify postcodes which cross local auth boundaries and throw `MultipleCouncilsException` if we encounter one (currently this is not handled anywhere)

In itself, this does not solve the problem - it just creates a way to identify it. As a result of this change:

* A call to `/postcode` or `/api/beta/postcode` with a postcode that crosses a local auth boundary will raise an unhandled `MultipleCouncilsException` (causing the call to throw a `500 Internal Server Error`) - unless we already hold `ResidentialAddress` data containing this postcode, in which case we don't currently call `geocode()`.
* The behaviour of all other calls should be unchanged

Handling this is going to require considering the cases where:

* Postcode maps to >=2 local auths and we currently hold data for 0 of them
* Postcode maps to >=2 local auths and we currently hold some mix of `PollingDistrict` and/or `ResidentialAddress` data for some of them, but not all
* Postcode maps to >=2 local auths and we currently hold some mix of `PollingDistrict` and/or `ResidentialAddress` data for all of them
* How we handle it in the API (+impact on docs, etc)
* How we handle it in the user-facing website
* Impact on the data import process

I may actually remove `MultipleCouncilsException` in the long-run and just modify `geocode()` so that `council_gss` is a list rather than a string (and then modify everything that consumes `geocode()` to expect a list and behave differently depending on whether there are 1 or >1 elements in the list), but before moving on to mechanisms for handling, I think this is a good point to submit this bit for review.

A few notes on what I've done:

* There are several places where I've intentionally written the code in a slightly verbose way. For example in https://github.com/chris48s/UK-Polling-Stations/blob/onsad/polling_stations/apps/data_finder/helpers.py#L203-L219, I could have written the more terse

```
    try:
        # first try addressbase
        result = addressbase.geocode()
    except MultipleCouncilsException:
        # this postcode contains uprns in multiple local authorities
        # re-raise the exception.
        raise
    except:
        # something went wrong: fall back to mapit
        result = mapit.geocode()
```

and the code at https://github.com/chris48s/UK-Polling-Stations/blob/onsad/polling_stations/apps/data_finder/helpers.py#L117-L121 could be removed completely, but I think in these cases it is valuable to sacrifice brevity and make the behaviour in each case really explicit.

* The geocoding process is now quite complicated so I have added a test suite to formalise the functionality (and done some refactoring so it is easier to mock out the api calls to mapit).
* Although I've tried to keep this easy to review, the diff is still quite large, but about 500 lines of it is adding large test fixtures - if you ignore those, hopefully this is a fairly manageable chunk.

Does this seem relatively sane so far??


chris48s:
Based on discussion in tonight's hangout about wider objectives for this, I'll carry on working on this in the context of this repo. The longer term goal will be that we should be able to spin this work out into a re-usable django app which can be pulled into both WCIVF and WDIV as a dependency as it will ultimately be applicable to both projects (and possibly other future DC projects).

In that context, I think the that what I've done here is still step 1 (and 2) so should still be reviewed and merged onto the `issue349` branch (or further changes requested) so I can work on the next step(s) using this as a base as future work depends on what I've done here. I'll carry on submitting a series of smaller PRs to be reviewed in stages.

davidmiller:
I've not reviewed the code, but the house of cards thing is super !
Nice one @chris48s :)

chris48s:
While I am pleased that you like my silly diagram, I would prefer it if 'house of cards' were a slightly less apt metaphor.

chris48s:
Pretty much everything I did in this PR has ended up getting chopped up and put into other PRs in some form now so I'm closing it

PR #482, opened 2017-02-15
Import scripts for Isle of Wight and Stockport

### LINKED ISSUE #477: Import E06000046-Isle-of-Wight
User story:

> As a CSV, I would like to be in your database

:D

### LINKED ISSUE #486: Import E08000007-Stockport

chris48s:
..and update base importer to deal with polling stations that have no postcode
Closes #477 
Closes #486

### DIFF
--- a/polling_stations/apps/data_collection/base_importers.py
+++ b/polling_stations/apps/data_collection/base_importers.py
@@ -657,9 +657,12 @@ class BaseXpressCsvImporter(BaseCsvStationsCsvAddressesImporter,
                 srid=27700)
         else:
             # otherwise, geocode using postcode
+            postcode = getattr(record, self.station_postcode_field).strip()
+            if postcode == '':
+                return None
+
             try:
-                location_data = geocode_point_only(
-                    getattr(record, self.station_postcode_field).strip())
+                location_data = geocode_point_only(postcode)
                 location = Point(
                     location_data['wgs84_lon'],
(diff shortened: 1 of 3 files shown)


chris48s:
Stockport needs the same modification

polling_stations/apps/data_collection/base_importers.py
```diff
@@ -657,9 +657,12 @@ def get_station_point(self, record):
                 srid=27700)
         else:
             # otherwise, geocode using postcode
+            postcode = getattr(record, self.station_postcode_field).strip()
+            if postcode == '':
```
symroe: Should this be `if not postcode`? It's possible that the data might parse in to `0` or `None` in some cases rather than an empty string. `not postcode` will call python's `bool` on `postcode` and catch all of these cases I think?

chris48s: If we did make it this far and this value was either `None` or `0` then the previous line just threw
`AttributeError: 'NoneType' object has no attribute 'strip'` or
`AttributeError: 'int' object has no attribute 'strip'`
so this check would not be reached.

In general the code that processes stuff sourced from `CsvHelper` assumes we are working with strings because `csv.reader` returns each row as a list of strings.

symroe: Righto, thought it was worth checking. In that case 👍 for this PR :)

PR #497, opened 2017-02-18
Import script for Hart

### LINKED ISSUE #471: Import E07000089-Hart

chris48s:
I don't think there is anything actionable to do based on this, but I want to explain what I've done just so that someone else is aware of it.

* Districts shapefile supplied by Hart was corrupt in 2 ways:
  * It was misreporting its CRS as WGS84, but the co-ordinates were actually in UK National Grid.
  * There was a feature with a null geometry (with a district code that doesn't appear in the stations file).
* I was able to import the file into QGIS (if I manually set the CRS correctly), but GDAL was only able to parse up to the null geometry. It wouldn't import anything else in the file after that.
* I tried fixing the shapefile by re-exporting the layer from QGIS - same problem
* I tried fixing the shapefile by running `ogr2ogr -skipfailures fixed_shapefile.shp corrupted_shapefile.shp` - same problem.
* I managed to fix it by exporting as GeoJSON, opening in a text editor, removing the feature with no geometry and slapping that on S3.
* Note that the GeoJson file uses CRS `urn:ogc:def:crs:EPSG::27700`. This is totally legal GeoJson, but non-WGS84 CRS is not a widely supported feature of the spec.

Resulting file now imports cleanly. As far as I can see, the resulting data checks out, but obviously I've meesed about with their data a bit, so want to make it clear that I have done that.

Closes #471

### DIFF
--- a/polling_stations/apps/data_collection/management/commands/import_hart.py
+++ b/polling_stations/apps/data_collection/management/commands/import_hart.py
@@ -1,0 +1,33 @@
+from django.contrib.gis.geos import Point
+from data_collection.management.commands import BaseCsvStationsJsonDistrictsImporter
+
+class Command(BaseCsvStationsJsonDistrictsImporter):
+    srid = 27700
+    districts_srid = 27700
+    council_id = 'E07000089'
+    districts_name = 'Hart Polling Districts-fixed.geojson'
+    stations_name = 'Hart Polling Stations 2017.csv'
+    elections = ['local.hampshire.2017-05-04']
+
+    def district_record_to_dict(self, record):
+        name = str(record['properties']['Name']).strip()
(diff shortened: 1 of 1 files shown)


JoeMitchell:
Hart DC got in touch to ask if we had any feedback. Want to feedback this to them?

chris48s:
Hmm.. not sure. I was able to import into a GIS app so may just be that the GIS libraries django uses are a bit more fussy about this sort of thing. If you're in touch with a really technical GIS person it might be worth forwarding them a link to this thread. If your contact is an elections officer it probably isn't going to go anywhere useful..

JoeMitchell:
Yep, fair enough. Will leave well enough alone!

PR #499, opened 2017-02-19
Add warning if morph api key not set

### LINKED ISSUE #467: Add MORPH_API_KEY to django's system checks
Since #456, `MORPH_API_KEY` is a requirement for `/league_table/scraper_report/` to work.

If this key is missing from the settings, that URL raises a HTTP 500, but it's not corrently included in the django system checks.

chris48s:
I've added this check as a warning as it only breaks one page if it is not set.

If you prefer, I can convert it to a fatal error and add an env var to the travis config with a junk api key so the build will pass. What do you reckon?

Closes #467


symroe:
Yeah, let's make it fail, otherwise we'll not hear about errors until the page fails in production

chris48s:
updated as requested

chris48s:
Not sure why this isn't working with the env vars, but if we can't figure it out, another option might be to add a statement that says something like `echo "MORPH_API_KEY = 'FakeApiKey'" >> polling_stations/settings/local.py` (not tested) to the build setup.

symroe:
I think adding this to `polling/stations/settings/testing.py` will work.

chris48s:
Fixed it :)
(`testing.py` is not pulled in when running `manage.py check`)

PR #501, opened 2017-02-19
Fixes and refactoring for addressbase helpers

chris48s:
One of the things we will need to do in order to deal with issue #349 is to modify the addressbase helpers to deal with situations where a postcode contains addresses that cross a local auth boundary.

I started having a look at this and realised that there were several things I wanted to change in the addressbase helpers before doing that which aren't _strictly_ related - so let's handle the groundwork first.

This PR is targetted at `master` as everything in this PR deals with improving the current behaviour or setup that will support future modifications rather than doing anything new (note there is nothing here that deals with postcodes spanning a local auth), but we will want to merge this into the `issue349` topic branch as well so I can build on it.

Stuff in this PR is explained in excruciating detail in the commit messages, so read those, but the broad themes are:

* Refactoring to make it easier to add more cases without having a hideously large `make_addresses_for_postcode` method as we add more cases (handling `PollingDistrict.DoesNotExist` will require several additional cases)
* Make `make_addresses_for_postcode` a bit faster by using `bulk_insert()` (although not by that much as most of the work is geo lookups, not database I/O)
* Make `make_addresses_for_postcode` a bit more easier to reason about/debug by removing use of `create_or_update()` and building up the data to insert before persisting to the DB
* Fix some edge cases I spotted
* Set the precedent for how we are going to deal with inserting address data in council areas where we don't hold any data - it is going to be necessary to store addresses which intentionally don't map to a polling station
* Add more tests for this

Once this and PR #465 are reviewed and merged, I can move on to the next stages of dealing with issue #349.

I'm fairly sure everything I have done here is sensible and doesn't break anything but this one could benefit from a thorough sanity check.


polling_stations/apps/addressbase/helpers.py
```diff
@@ -18,6 +19,84 @@ def centre_from_points_qs(qs):
...
 
+AddressTuple = namedtuple('Address', [
+    'address',
+    'postcode',
+    'council',
+    'polling_station_id',
+    'slug'])
+
+
+class AddressSet:
```
symroe: Could this just subclass `set`, and only define the `save` method? That would allow us to use all the built in set methods without having to access `self.elements`

polling_stations/apps/addressbase/helpers.py
```diff
@@ -18,6 +19,84 @@ def centre_from_points_qs(qs):
...
+        addresses = Address.objects.filter(postcode=postcode)
+        for address in addresses:
+            try:
+                district = PollingDistrict.objects.get(
+                    area__covers=address.location, council_id=self.target_council_id)
+            except PollingDistrict.DoesNotExist:
+                # Chances are this is on the edge of the council area, and
+                # we don't have data for the are the property is in
+                continue
+            except PollingDistrict.MultipleObjectsReturned:
```
symroe: Is it worth logging overlapping districts? It could point to fixable problems, like clipping district boundaries by council boundaries. (Also typo on the line below: s/the/that/)

chris48s:
These are both excellent suggestions :+1: - updated

PR #508, opened 2017-02-23
Import script for Ribble Valley

### LINKED ISSUE #507: Import E07000124-Ribble-Valley

chris48s:
Another un-parsable shp file here. Was getting errors like `OSError: [Errno 22] Invalid argument` and `struct.error: unpack requires a bytes object of length 8` from GEOS. Couldn't fix it so I did the same 'read it into QGIS and export as GeoJson' trick I used on Hart to get something importable. Again it is `urn:ogc:def:crs:EPSG::27700`.

We've got one orphan district so I've asked Joe to follow it up. If they get back to us with a correction, I'll update.

Closes #507

### DIFF
--- a/polling_stations/apps/data_collection/management/commands/import_ribble.py
+++ b/polling_stations/apps/data_collection/management/commands/import_ribble.py
@@ -1,0 +1,45 @@
+from django.contrib.gis.geos import Point
+from data_collection.base_importers import BaseStationsDistrictsImporter
+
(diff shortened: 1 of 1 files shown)


chris48s:
Joe followed up + received correction from nice people at Ribble Valley:

> Bashall Eaves Village Hall missing CL in Code 2 column

Since I had already exported to geojson, I just added the missing code to the geojson file on S3 so this should now import with 100% coverage.

PR #539, opened 2017-03-11
Show an error if there are no upcoming elections in the user's area

### LINKED ISSUE #431: What do we show users if they don't have an election?
> My authority [Medway] doesn’t have any scheduled elections in 2017; I wondered what message will show on the screen if anyone in Medway inputs a Medway postcode into your polling station finder?

Ideally, our EveryElection API will be able to tell us whether to display polling stations for a postcode or not. So in the above example, it would tell users that "You don't have any elections coming up."

But if that's not working to our satisfaction (i.e. it's probable we won't know about several by-elections) — what do we do?

chris48s:
Note: This currently does not work properly for all areas because YNR is still throwing up some false negatives. For example, if I call https://candidates.democracyclub.org.uk/upcoming-elections?postcode=PR32QH (Ribble Valley), it throws back `[]` but we know that `local.lancashire.2017-05-04` is happening in this area. However some areas are returning upcoming elections correctly e.g: https://candidates.democracyclub.org.uk/upcoming-elections?postcode=CF105AJ (Cardiff). As such, you might want to hold off on merging this depending on whether you think false positives or false negatives are worse at this stage.

Changes in this PR:
* Check if there are upcoming elections in the users area. If not, throw an error. I have assumed that we are going to have enough of a handle on upcoming elections that I've made the wording confident. i.e: "There are no upcoming elections in your area".
* Offer a handy 'upcoming elections' iCal feed by linking to `webcal://whocanivotefor.co.uk/elections/{yourpostcodehere}.ics`
* Add an election reminders signup form. Note that
    1. We will need to remember to check this (ping @JoeMitchell  ) and also
    2. This represents a tech debt that we need to pay off - as a longer term goal, we would like to pull this stuff out into a common/shared thing rather than implementing essentially the same functionality in multiple projects (refs https://trello.com/c/a1iGq4uG/977-think-about-how-to-share-common-components-between-dc-products )

Closes #431


chris48s:
> Oh I expect you’ll want `https://elections.democracyclub.org.uk/api/elections/?postcode={postcode}&poll_open_date__gte=` and then pass it `datetime.today()` or whatever

chris48s:
Hmm...bit of a head-scratcher this one.

Tests pass locally but build is failing. I don't think it is related to the changes in this PR though as I've tried restarting the build a few times and it isn't failing in the same place each time. In particular, one time it failed trying to call `world.browser.quit()` rather than running any of the actual tests.
I think this may actually be a travis issue. I'm going to leave it for the minute and try running the build again tomorrow and see what happens.

chris48s:
Rebased to resolve conflicts

chris48s:
promise it is actually ready for review this time

polling_stations/apps/data_finder/views.py
```diff
@@ -146,6 +159,10 @@ def get_context_data(self, **context):
         self.station = self.get_station()
         self.directions = self.get_directions()
 
+        context['has_election'] = self.has_election()
+        if not context['has_election']:
```
chris48s: sounds like a plan. If we do it that way in future, would it be sensible for that app to be a dependency shared with WDIV (so we aren't solving the same problem twice in 2 different places), or does WDIV necessarily obtain election info from YNR so it can be referenced against other the other YNR data?

symroe: Yeah, I think WDIVF should aim to use the same system. Longer term still we should think about how useful it really is to have different public facing sites…(that's for talking about another day though!)

polling_stations/templates/postcode_view.html
```diff
@@ -41,25 +41,53 @@
...
+        {% else %}
+          {% blocktrans with council.name as council_name %}
+          <p>
+            We're not aware of any upcoming elections in your area. If you think there may be elections in your area, please call {{ council_name }} to check and then <a href="https://democracyclub.org.uk/contact/">let us know</a>.*
+          </p>
+          {% endblocktrans %}
+        {% endif %}
+        <p>
+          <small>
+            * Why are we not certain? Because there is no canonical source of UK elections.
```
symroe: I asked some people from Bournemouth Council about this (they happened to email, so I showed them the wording we wanted to use). They said:

> Two colleagues of mine did comment on the use of the word 'canonical' though, not knowing what it meant and thinking electors might not either.  We thought something like 'definitive' might be more likely to be understood?

@JoeMitchell do you have reckons?

chris48s:
Wording amended. Still links to EE

PR #547, opened 2017-03-12
fix Leeds and Eden import scripts

chris48s:
Looks like we moved a class and didn't update the `import` statements.
Not tested as morph is [currently down](https://github.com/openaustralia/morph/issues/1129) but they will at least load now.


chris48s:
I've now tested these and they both run

Issue #546, opened 2017-03-12
Set up test coverage reports


chris48s:
The project has a fairly decent suite of unit tests, but I have completely lost track of what we do and don't have tests for. It would be useful if we had coverage reports so we know where we need to improve the test suite.

There a couple of minor wrinkles we need to consider with this project:
* We have a mix of unit tests implemented with `unittest` and acceptance tests implemented with aloe/selenium and we need to aggregate the coverage stats from both types of tests into a single coverage report.
* We need to exclude import scripts from the coverage reports

I did try setting this up with `django-nose` at some point in the past (and failed). I remember hitting a few problems:
* The command to delete all the data from the DB (we run this to clean the old data out of the DB before importing all the data for a new election) is called `teardown`. Nose picks this up as something it wants to run after every test, which we don't really want. This can be fixed this by renaming it to something else.
* Nose doesn't like something about the way we're doing our `import` statements. It is some time since I did anything with this, but I think it is a problem if you import the same module using `from . import foo` if you're importing within the same app and then elsewhere import the same module using `from my.absolute.namespace import foo` that confuses it? I think this was ultimately the thing that proved the major sticking point.
* There may have been others too.. it was a while ago

@symroe - I notice that you've started using `tox` as a test runner on the [DemocracyClub/Website](https://github.com/DemocracyClub/Website) and [DemocracyClub/EveryElection](https://github.com/DemocracyClub/EveryElection/) repos in the last week or so. I wonder if this might be a better approach than `nose` for this project too?

symroe:
So tox is more of a test runer-runner if you see what I mean, in that it really just manages the commands you want to run, and the test matrix (python and django versions). In the case of a website, I don't think the matrix is as useful (it's designed for libraries I think), but the interface is nice anyway, so I've been using it.

The _actual_ test runner I've been using is `pytest`, and that seems _really_ nice. 

I've just copied over the config files I used for WCIVF and renamed a couple of bits, and everything seems to work nicely. See #548.

Things to note: 

1. I've excluded all the `import_*` management commands, as the coverage goes down to 30% with them in ;)

2. In the other projects I've added `pytest --flake` as the 2nd test (after `manage.py check`). I've commented it out for this PR because it breaks quite a lot, but I do like being flake8 compliant in my projects, if only because it can catch really simple mistakes better. You'll note that I've commented out the pep8 check in all the projects! 

3. The coverage of `harvest` is appended to the `pytest` coverage.

[...]

PR #548, opened 2017-03-12
Add pytest and tox

symroe:
See #546 for conversation


chris48s:
See comments in #546 - will post if I spot any config changes needed but can we also have a README badge, please! :)

.coveragerc
```diff
@@ -0,0 +1,9 @@
+[run]
+omit =
+    */tests/*
+    */apps.py
+    */migrations/*
+    */migrations/*
```
chris48s: duplicate line

tox.ini
```diff
@@ -0,0 +1,17 @@
+[tox]
+envlist = py34-django110
```
chris48s: This project uses Django 1.8 - it might be useful to start looking at upgrading, but for the moment lets target the version in the requirements

.coveragerc
```diff
@@ -0,0 +1,9 @@
...
+    polling_stations/wsgi.py
+    */lib/python3.4/site-packages/*
+    polling_stations/apps/data_collection/management/commands/import_*
```
chris48s: minor: can we also exclude data_collection/management/commands/misc_fixes.py

PR #550, opened 2017-03-12
Correctly handle NI councils not being in councils table

### LINKED ISSUE #531: Update API docs to clarify council is nullable
Even though it will currently never be null

chris48s:
This PR goes hand-in-hand with the changes in #441 (it probably doesn't make much sense without that). It adds a variety of special cases so that we can direct NI users to the EONI site without having to have the NI councils in the councils table (and not throw 500 errors). Note that we are now relying on a `CustomFinder` record for the EONI finder being present (although we fail slightly more cleanly without it). I've added a fixture and a note in the `README.md`. I did also want to add a system check for it, but Django doesn't seem to like you trying to load a model in the system checks (I think it hasn't fully bootstrapped the app yet at the point it runs the checks) so I don't think that is an option. Dunno if you've got any other ideas on that?

Additionally, this probably creates a merge conflict with the acceptance tests in PR #501 so one or other will probably need rebasing in order to merge both.

Refs #441
Also closes #531 as it became relevant here too.

Btw, cheers for sorting coverage reports out today. Having it has already provided incentive for me to add tests that I probably wouldn't have bothered to write if I had done this yesterday :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,14 @@ python manage.py migrate
 python manage.py import_councils
 ```
 
+#### Load CustomFinder for Northern Ireland
+
+The Electoral Office for Northern Ireland run their own finder. We redirect our users to that.
+
+```
+python manage.py loaddata polling_stations/apps/data_finder/fixtures/northern_ireland.json
+```
+
 #### Import some Polling District/Station data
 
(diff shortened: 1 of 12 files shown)


chris48s:
rebased to resolve merge conflict

chris48s:
Pointless now #441 is closed

PR #552, opened 2017-03-13
ignore files called tests.py in coverage reports

chris48s:
ronseal


symroe:
This is the best way to increase coverage! ;)

PR #574, opened 2017-03-17
Fix build matrix

chris48s:
While puzzling over travis logs trying to work out what I've done to cause the odd build failures in PR #539, I noticed that the tests were always running on python 3.4 even when they were supposed to be running on 3.5 and a couple of minor things - so I fixed those while I chew the problem over.


chris48s:
top tip: improve your test coverage by actually run all your tests :D

PR #607, opened 2017-03-24
Council importer fixes

symroe:


chris48s:
Have updated my `councils` table with this - all looks fine.

The only comment I've got is that the YVM site returns all the council websites in the form `www.crawley.gov.uk/pw/index.htm` as opposed to `http://www.crawley.gov.uk/pw/index.htm` or `https://www.crawley.gov.uk/pw/index.htm`. I don't think this _breaks_ anything as we're not actually linking to them anywhere, but we do surface this data point in the API outputs. Do you think there's any value in appending `http://` to them all in the import process to make it easier for data users who might want to link to councils? Given it doesn't break anything for us, I don't have strong opinions on it.

Note: I've already accounted for this in the updates to the API docs e.g:
https://github.com/chris48s/UK-Polling-Stations/blob/83768ed19e380e10521893dbb699afd3c6db7014/polling_stations/templates/api_docs/blueprints/data_structures.apibp#L23
do if you change it, let me know and I'll update that PR.

Apart from that, all fine.

symroe:
`http://` and tests added

polling_stations/apps/councils/management/commands/import_councils.py
```diff
@@ -1,4 +1,5 @@
 import time
+import json
 
 import requests
 from bs4 import BeautifulSoup
```
chris48s: This import is now not used. In fact, can we completely remove BS4 from the project requirements?

polling_stations/apps/councils/management/commands/import_councils.py
```diff
@@ -21,6 +22,21 @@ class Command(BaseCommand):
...
+        headers['User-Agent'] = settings.MAPIT_UA
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '-n',
+            '--nosleep',
+            default=False,
+            action='store_true',
+            required=False,
+            help='Sleep between requests'
```
chris48s: Help text and param name disagree with each other

symroe:
Thanks, some fixup commits pushed that I'll rebase before merging

chris48s:
cool - I'll leave it to you to merge after you've squashed to tidy up :)

PR #631, opened 2017-03-30
Address base import improvements

symroe:


chris48s:
Given you've already done a deploy with these changes and I can't be bothered to crunch through some-million rows of data to test it I am just going to merge this - if it falls over next time we import AddressBase, we'll worry about it then :)

symroe:
"I don't always do QA, but when I do, I do it in production" ;)

PR #632, opened 2017-03-30
Multiprocessing import

symroe:


polling_stations/apps/data_collection/management/commands/import.py
```diff
@@ -110,18 +126,44 @@ def handle(self, *args, **kwargs):
...
-                        opts = {
-                            'noclean': False,
-                            'verbosity': 1
-                        }
+                    # # Only run if
+                    # existing_data = PollingStation.objects.filter(
+                    #     council_id=cmd.council_id).exists()
+                    # if not existing_data or kwargs.get('overwrite'):
+                    #     self.summary.append(
+                    #         ('INFO', "Ran import script %s" % tail))
```
chris48s: So in terms of this block you've commented out:
1. We now have no log/summary output of which import scripts were run. I think this is quite a useful output (esp with a regex) so would like to retain it if possible. Is there a reason why it is problematic to do this when running across multiple cores? If so, would it seem reasonable to at least output a list of all the scripts it is going to try and run before it starts? That way we would at least know "I tried to run this list of scripts and they all worked" or "I tried to run this list of scripts and script X failed".
2. You've removed the code that implements the `overwrite` flag. Can we either remove the flag or implement it (I think it would be useful to retain if possible)
3. Unless there's a specific reason to keep it commented, can we just remove the commented chunk

polling_stations/apps/data_collection/management/commands/import.py
```diff
@@ -110,18 +126,44 @@ def handle(self, *args, **kwargs):
...
+                    }
+                    commands_to_run.append(f)
+                    if not kwargs['multiprocessing']:
                         cmd.handle(**opts)
             else:
                 self.summary.append(('WARNING', "%s does not contain elections property!" % tail))
 
         self.output_summary()
+
+        opts = {
```
chris48s: I don't think this declaration is actually used anywhere

chris48s:
@symroe - I'm quite keen to get this finalised so I have added some commits to address the issues I've raised and tidy up a bit. Can you review when you get a chance.

Cheers

symroe:
Yep, nice, thanks for doing all this!

PR #635, opened 2017-03-31
Fix OS X bug with DataSource

symroe:
As noted elsewhere, OS X will always fail to open and data the first
time ¯\_(ツ)_/¯


chris48s:
A fine example of best practice :+1:

PR #649, opened 2017-04-06
Multiple Councils "We Don't Know" page

### LINKED ISSUE #534: Implement static postcode backlist
* Add a table we can fill with postcodes which are known to contain properties in >1 local auth
* Website and API `/postcode` lookups should query this table before doing anything else and then...
    * API: set council to `null` + return no polling station details (even if we think we know)
    * Website: er.. a thing. Not the "we don't know - call your council" page because we don't know what council they should call. Thoughts?

Maybe we need an "achievement unlocked" page: "Well done. Your postcode has an interesting property (and is probably also a bit annoying to live at)."

Refs #532

chris48s:
This PR adds the ability to show a special "we don't know" page for users whose postcode is split across multiple local authorities. It isn't the best solution to this problem, but it represents a good stop-gap from a pragmatic perspective.

A few notes on this:
* The blacklist overrides everything - even if we hold some data for some/all of the authorities concerned. If the postcode entered is split, we just say "we don't know".
* Overall test coverage has probably decreased because I haven't written tests for the import scripts but the API and front-end changes are all covered by new unit/integration test cases.
* I've gone for a de-normalised approach because
  * The `RoutingHelper` code is called on every request and I wanted to avoid joining 2x ~30 million row tables together on every request
  * If either of the tables concerned change we basically have to rebuild the entire app anyway so the benefits of normalisation are moot
* If there is no data in the `blacklist` table, this will still work - all postcodes will just be not blacklisted so this should be deployable even if we don't populate the table with anything.
* This changes our deployment process. When we build the AddressBase image, we need to:
  * Grab the latest ONSAD from http://ons.maps.arcgis.com/home/search.html?q=ONS%20Address%20Directory&t=content
  * Stick it on S3
  * Grab it from S3 and unzip it
  * Run `./manage.py import_onsad`
  * Run `./manage.py create_blacklist`
  I will update the deploy scripts and README to account for that when I do the next deployment.
* I haven't done anything specific about AB11 5QH as it is an edge case on the edge of an edge case. We can build in a special case for it, but I doubt it is the only mistake in that data - it is just one that I know about. First step is to report it to OS and see what happens..
* That bug where the integration tests fail on travis seems to be back. I don't particularly like the "fix" in 0752be4 but it seems to do the job.. any thoughts?

Closes #534 
Refs #349


chris48s:
Apparently coverage has increased even though I've added a bunch of code that doesn't have tests..
Looks like some of the addressbase stuff is getting ignored by coveralls, but I'm not entirely sure why. Hmm

chris48s:
That's ~better~ ~worse~ accurate

PR #651, opened 2017-04-07
Prioritise AddressBase/ONSAD for geocoding

chris48s:
Now that we are importing the ONSAD (see PR #649) there is scope to reduce our dependency on mapit by prioritising Addressbase/ONSAD geocoding where possible.

This gives us several advantages:

* It will be faster to query our local database than deal with network latency on every request
* The DC mapit install becomes less of a single point of failure for all our sites
* It supports our wider objective to move away from maintaining our own mapit instance

Key changes in this PR are:

* Add ability to identify council and attach GSS codes to postcodes using AddressBase/ONSAD
* Prioritise AddressBase/ONSAD as a source for geocoding/GSS codes
* Fall back to centroid-based geocoding using mapit if we can't sucessfully geocode using AddressBase/ONSAD (there are various situations where this may happen)
* Identify postcodes which cross local auth boundaries and throw `MultipleCouncilsException` if we encounter one
* Handle `MultipleCouncilsException` in API and front-end
* Refactor geocoding stuff to improve structure and make it easier to mock out mapit responses for testing purposes
* Add a load of additional unit tests

A few notes on what I've done:

* There are several places where I've intentionally written the code in a slightly verbose way. For example in https://github.com/chris48s/UK-Polling-Stations/blob/geocode_fix/polling_stations/apps/data_finder/helpers.py#L207-L223, I could have written the more terse

```
    try:
        # first try addressbase
        result = addressbase.geocode()
    except MultipleCouncilsException:
        # this postcode contains uprns in multiple local authorities
        # re-raise the exception.
        raise
    except:
        # something went wrong: fall back to mapit
        result = mapit.geocode()
```

and the code at https://github.com/chris48s/UK-Polling-Stations/blob/geocode_fix/polling_stations/apps/data_finder/helpers.py#L121-L125 could be removed completely, but I think in these cases it is valuable to sacrifice brevity and make the behaviour in each case really explicit.

### DIFF
--- a/polling_stations/apps/data_finder/helpers.py
+++ b/polling_stations/apps/data_finder/helpers.py
@@ -191,55 +191,67 @@ class AddressBaseGeocoder(BaseGeocoder):
 
 
 def geocode_point_only(postcode, sleep=True):
-    addressbase = AddressBaseGeocoder(postcode)
-    mapit = MapitGeocoder(postcode)
+    geocoders = (AddressBaseGeocoder(postcode), MapitGeocoder(postcode))
+    for geocoder in geocoders:
+        try:
+            return geocoder.run(True)
+        except ObjectDoesNotExist:
+            # we couldn't find this postcode in AddressBase
+            # fall back to the next source
 
-    try:
-        # first try addressbase
-        result = addressbase.geocode_point_only()
-    except ObjectDoesNotExist:
-        # we couldn't find this postcode in AddressBase: fall back to mapit
+            # optional sleep to avoid hammering external services
+            if sleep:
+                time.sleep(1.3)
 
-        # optional sleep to avoid hammering mapit
-        if sleep:
-            time.sleep(1.3)
+            continue
+        except PostcodeError:
+            # we were unable to geocode this postcode using mapit
+            # re-raise the exception.
+            # Note: in future we may want to fall back to yet another source
+            raise
+        except:
+            # something else went wrong:
+            # lets give the next source a try anyway
 
-        result = mapit.geocode()
-    except:
-        # something else went wrong: lets give mapit a go anyway
+            # optional sleep to avoid hammering external services
+            if sleep:
+                time.sleep(1.3)
 
-        # optional sleep to avoid hammering mapit
-        if sleep:
-            time.sleep(1.3)
+            continue
 
-        result = mapit.geocode()
-
-    return result
+    # All of our attempts to geocode this failed. Raise a generic exception
+    raise PostcodeError('Could not geocode from any source')
 
 
 def geocode(postcode):
-    addressbase = AddressBaseGeocoder(postcode)
-    mapit = MapitGeocoder(postcode)
+    geocoders = (AddressBaseGeocoder(postcode), MapitGeocoder(postcode))
+    for geocoder in geocoders:
+        try:
+            return geocoder.run(False)
+        except ObjectDoesNotExist:
+            # we couldn't find this postcode in AddressBase
+            # fall back to the next source
+            continue
+        except CodesNotFoundException:
+            # we did find this postcode in AddressBase, but there were no
+            # corresponding codes in ONSAD: fall back to the next source
+            continue
+        except MultipleCouncilsException:
+            # this postcode contains uprns in multiple local authorities
+            # re-raise the exception.
+            raise
+        except PostcodeError:
+            # we were unable to geocode this postcode using mapit
+            # re-raise the exception.
+            # Note: in future we may want to fall back to yet another source
+            raise
+        except:
+            # something else went wrong:
+            # lets give the next source a try anyway
+            continue
 
-    try:
-        # first try addressbase
-        result = addressbase.geocode()
-    except ObjectDoesNotExist:
-        # we couldn't find this postcode in AddressBase: fall back to mapit
-        result = mapit.geocode()
-    except CodesNotFoundException:
-        # we did find this postcode in AddressBase, but there were no
-        # corresponding codes in ONSAD: fall back to mapit
-        result = mapit.geocode()
-    except MultipleCouncilsException:
-        # this postcode contains uprns in multiple local authorities
-        # re-raise the exception.
-        raise
-    except:
-        # something else went wrong: lets give mapit a go anyway
-        result = mapit.geocode()
-
-    return result
+    # All of our attempts to geocode this failed. Raise a generic exception
+    raise PostcodeError('Could not geocode from any source')
 
 
 class AddressSorter:
--- a/polling_stations/apps/data_finder/helpers.py
+++ b/polling_stations/apps/data_finder/helpers.py
@@ -91,10 +91,6 @@ class MapitWrapper:
         }
 
 
-def mapit_wrapper_factory(postcode):
-    return MapitWrapper(postcode)
-
-
 class AddressBaseWrapper:
 
     def __init__(self, postcode):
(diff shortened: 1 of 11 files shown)


polling_stations/apps/data_finder/helpers.py
```diff
@@ -7,99 +7,222 @@
...
-        'wgs84_lat': res_json['wgs84_lat'],
-        'gss_codes': gss_codes,
-        'council_gss': council_gss,
-    }
+        result = mapit.geocode()
+
+    return result
+
+
+def geocode(postcode, mapit=mapit_wrapper_factory):
```
symroe: (Comment on this line, but really about most of ad33cd9208186b436d64972031ab72a31a39755c)

Few things: 

1. I'm not sure what `mapit_wrapper_factory` is doing, or why `AddressBaseWrapper` doesn't have one. I think `mapit_wrapper_factory` can be removed and `MapitWrapper` called from `geocode` directly.

2. I think the approach here should be to have a list of classes that can do the geocoding, in the order we want. The first class that can do the geocoding is returned. This is more like the pattern I've written here: https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/elections/query_helpers.py. 

3. If you think 2. is a good idea, then I think there should be a base class that sets `self.postcode` and formats it and provides a runner something like:

```python
class BaseGeoCoder:
    def __init__(self, postcode):
        self.postcode = self.format_postcode(postcode)
        
    def run(self, point_only=False):
        if point_only:
            return self.geocode_point_only()
        else:
            return self.geocode()
```

Then `geocode` becomes something like this:

```python
def geocode(postcode, point_only=False):
    for geocoder in (AddressBaseGeocoder, MapitGeocoder):
        try:
            return geocoder.run(postcode, point_only)
        except GeocodeError:
            continue

def geocode_point_only(postcode):
    return geocode(postcode, point_only=True)
```

(assume better Exception handling here, but you get the point)

I think this is simpler, and might allow for adding more geocoders in future (for example, we can fall back from AddressBase to the S3 cache to MapIt when we're doing a point_only call.

chris48s: > I'm not sure what `mapit_wrapper_factory` is doing

The reason for `mapit_factory` is for tesing/mocking purposes - its just a way to make `MapitWrapper` an injected dependency.

By passing in a function which return an instance of `MapitWrapper` as a default parameter, when we are writing tests we can easily create a mock `MapitWrapper` class and pass in a `mock_factory` instead. This means we can more easily abstract the behaviour of `MapitWrapper` when testing the exception hadling in `geocode()` (where we don't care about the actual response data - just which class returned it) without having to use really verbose VCR fixtures. See https://github.com/DemocracyClub/UK-Polling-Stations/pull/651/files#diff-1ef176bdf330b9e534a5489f367dac51R5

I'd really like to keep the ability to do that rather than have to use VCR to fake the HTTP responses in this instance. It may be that there is a more 'pythonic' way of injecting this than using a factory though. Suggestions?

> or why `AddressBaseWrapper` doesn't have one

The reason I didn't do this for `AddressBaseWrapper` is because there are no HTTP calls being made to mock. I defined the behaviour for `AddressBaseWrapper` in those tests by just re-using the existing addressbase DB fixture that was needed for the tests in https://github.com/DemocracyClub/UK-Polling-Stations/pull/651/files#diff-98027f40280a91d59bb6094c405daeeaR8 anyway.


> I think the approach here should be to have a list of classes that can do the geocoding, in the order we want

Yes. I do like your idea in points 2 and 3 about making the geocoding sources more generic/flexible so that we can swap sources in and out more easily. I will think through the details of that a bit more this afternoon.

symroe: I reckon on 1. using [`mock`](https://docs.python.org/dev/library/unittest.mock.html) would be better. Something like:
```python
from mock import patch
class MyTest(unittest.TestCase):

    @patch('MapitWrapper.geocode')    
    def test_no_records(self, mocked_mapit_wrapper_geocode):
        mocked_mapit_wrapper_geocode.return_value = { 'source': 'mapit' }
        self.assertEqual('mapit', result['source'])
```

(Untested, but you get the idea)

chris48s: ok. I'll have a look at that as well.

chris48s:
Right. I'm totally on board with you up to 167527c

After that, I tried implementing `geocode()` and `geocode_point_only()` as you've suggested and I ended up with this:

```python
def geocode(postcode, sleep=False, point_only=False):
    geocoders = (AddressBaseGeocoder(postcode), MapitGeocoder(postcode))
    for geocoder in geocoders:
        try:
            return geocoder.run(point_only)
        except ObjectDoesNotExist:
            # we couldn't find this postcode in AddressBase
            # fall back to the next source

            # optional sleep to avoid hammering external services
            if sleep:
                time.sleep(1.3)

            continue
        except CodesNotFoundException:
            # we did find this postcode in AddressBase, but there were no
            # corresponding codes in ONSAD: fall back to the next source

            # optional sleep to avoid hammering external services
            if sleep:
                time.sleep(1.3)

            continue
        except MultipleCouncilsException:
            # this postcode contains uprns in multiple local authorities
            # re-raise the exception.
            raise
        except PostcodeError:
            # we were unable to geocode this postcode using mapit
            # re-raise the exception.
            # Note: in future we may want to fall back to yet another source
            raise
        except:
            # something else went wrong:
            # lets give the next source a try anyway

            # optional sleep to avoid hammering external services
            if sleep:
                time.sleep(1.3)

            continue
    # All of our attempts to geocode this failed. Raise a generic exception
    raise PostcodeError('Could not geocode from any source')


def geocode_point_only(postcode, sleep=True):
    return geocode(postcode, sleep=sleep, point_only=True)
```

There's a couple of reasons why I'm not massively keen on that implementation:

1. All the exception handling code gets mashed together regardless of which sort of `geocode` we are doing, but a `CodesNotFoundException` or a `MultipleCouncilsException` are irrelevant if we're doing a `geocode_point_only()`. I think combining the handlers makes it less clear what conditions we might expect to happen in each case. For example, it implies `geocode_point_only()` can raise `MultipleCouncilsException`.
2. It forces `geocode()` to implement `sleep`. In order to retain the existing functionality I've had to give `geocode_point_only()` a `sleep` param which defaults to `True` but `geocode()` a `sleep` param which defaults to `False` which I think makes things confusing/inconsistent.

For this reason I have chosen to follow your suggested structure but keep `geocode()` and `geocode_point_only()` as 2 seperate functions in 821ac33

That said, it is a more verbose solution and I'm totally willing to accept that there is another consideration I haven't thought of.. comments?

polling_stations/apps/data_finder/helpers.py
```diff
@@ -191,55 +191,67 @@ def geocode_point_only(self):
...
+            return geocoder.run(True)
+        except ObjectDoesNotExist:
+            # we couldn't find this postcode in AddressBase
+            # fall back to the next source
 
-        # optional sleep to avoid hammering mapit
-        if sleep:
-            time.sleep(1.3)
+            # optional sleep to avoid hammering external services
+            if sleep:
```
symroe: Just a small thing: is it worth passing `sleep` on to the classes and letting them sleep if they need to? Otherwise we might end up in the situation where we sleep in between AddressBase calls. Basically, each class should be able to ignore sleep if it wants to.

chris48s: ..or set a different sleep interval based on that API's usage policy, I guess.
Sounds like a plan

chris48s:
merging this for now - we'll improve this independently as issue #675

PR #682, opened 2017-04-13
fail cleanly if post data doesn't contain 'postcode'

chris48s:
this will stop the `MultiValueDictKeyError` emails


polling_stations/apps/data_finder/views.py
```diff
@@ -97,7 +97,10 @@ def form_valid(self, form):
 
     def form_invalid(self, form):
         context = self.get_context_data(form=form)
-        context['postcode'] = form.data['postcode']
+        if 'postcode' in form.data:
```
symroe: `form.data.get('postcode')`?

PR #699, opened 2017-04-15
Consume Election Explainers data from Every Election

chris48s:
This PR goes hand-in-hand with https://github.com/DemocracyClub/EveryElection/pull/28 It is designed such that if EE is not returning explainers, it won't fall over but it is still probably worth co-ordinating the two. If I make changes to one they may need to be reflected in the other. I'll leave it WIP pending review of https://github.com/DemocracyClub/EveryElection/pull/28

A couple of notes/questions:
1. I've chosen to only show the explainers card when:
    * Polling station is found
    * Brand is DemocracyClub (i.e: not in the embedded version)
      Are these the correct decisions?
2. Here's an example of what the output might look like for an area with multiple elections if `get_explanations()` were to return

```
[
"""
### County Council Elections

On 4 May 2017 you will elect your county councillor in a [first past the post](http://www.parliament.uk/about/how/elections-and-voting/voting-systems/) election. Your county council is responsible for services across the whole county, including:

* education
* transport
* planning
* fire and public safety
* social care
* libraries
* waste management
* trading standards

Separately your local authority is responsible for services delivered at a local level, including:

* rubbish collection
* recycling
* Council Tax collections
* housing
* planning applications
Your local councillors are elected in separate elections.
""",

"""
### Combined Authority Mayoral Elections

On 4 May 2017 you will elect a regional mayor in a [supplementary vote](http://www.parliament.uk/about/how/elections-and-voting/voting-systems/) election. Read more about [combined authority mayors](http://www.local.gov.uk/devolution/directly-elected-mayors).
"""
]
```

screenshot at 2017-04-15 23 05 27

Obviously there's still some work we would like to do on the actual explainer text (outside scope of this PR), but does the layout, placement, etc seem ok?

Once we get something we're happy with here, we can look at doing something similar on WCIVF where this content is equally relevant.

### DIFF
--- a/polling_stations/apps/data_finder/helpers.py
+++ b/polling_stations/apps/data_finder/helpers.py
@@ -286,17 +286,32 @@ class AddressSorter:
         return sorted(lst, key = self.swap_fields)
 
 
-def has_election(postcode):
-    res = requests.get("%sapi/elections.json?postcode=%s&future=1" % (
-        settings.EE_BASE, postcode))
+class EveryElectionWrapper:
 
-    if res.status_code != 200:
-        res.raise_for_status()
+    def __init__(self, postcode):
+        try:
+            self.elections = self.get_data(postcode)
+            self.request_success = True
+        except:
+            self.request_success = False
 
-    res_json = res.json()
-    if len(res_json['results']) > 0:
-        return True
-    return False
+    def get_data(self, postcode):
+        res = requests.get("%sapi/elections.json?postcode=%s&future=1" % (
+            settings.EE_BASE, postcode))
+        res_json = res.json()
+        return res_json['results']
+
+    def has_election(self):
+
+        if not self.request_success:
+            # if the request was unsucessful for some reason,
+            # assume there *is* an upcoming election
+            return True
+
+        if len(self.elections) > 0:
+            return True
+
+        return False
 
 
 class OrsDirectionsApiError(Exception):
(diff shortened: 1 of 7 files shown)


JoeMitchell:
Looks very cool! Gradually collapsing the difference between Where and Who
and What and How :D

I wonder if the different elections should be on different cards?

chris48s:
Yeah, so here's a few other alternative layouts I tried (and discarded):

### Inception cards
screenshot at 2017-04-16 11 33 01

### Different cards, no title

(maybe the title is not needed)

screenshot at 2017-04-16 11 33 59

### Different cards, disconnected title
screenshot at 2017-04-16 11 35 08

I think we should go with either the original layout proposed in the PR (i.e: one card titled "Elections in your area" with the details of all/any applicable elections on a single card) or the "different cards, no title" layout. I don't really like the other 2. The reason I've come down on the side of putting all relevant explainers on a single card is because I think it is very slightly clearer on mobile devices (where there is a lot of vertical scrolling) that there's more than one explanation to read.. but it is a very subjective thing. If you guys prefer multiple cards, I don't have strong views on it.

One other thing to consider is that there are *some* areas where we will need to show >1 explainers but in most areas we only need to show one. Our layout needs to make sense in both cases.

JoeMitchell:
Ooh, nice, I personally prefer Inception cards (!) — this is what was in my head when I made my earlier comment — but maybe share in the Slack and see what people think? Or I can put on the blog today?

chris48s:
`0f68462` does the same thing as #722. If we merge #722 first, this will need rebasing to resolve MC. If we end up merging this one first, we can just close #722 as it becomes pointless

chris48s:
rebased to resolve conflict

symroe:
Lovely stuff – as for the card layout, let's deploy and get wider feedback, it's not the hard part of this change :)

PR #703, opened 2017-04-17
Improve import script performance

chris48s:
I ran another `imported-db` build earlier today. It took ages. This will not be improved once I start working through all the Scotland data next week (it is all shp files), so I decided to see if I could improve efficiency some more. The slowest part of the import script is finding all the postcodes which contain UPRNs in >1 polling district and assigning each UPRN a polling station, so I've focussed my efforts there.

The main problem was that `EdgeCaseFixer.make_addresses_for_postcode()` was running a minimum of 2 DB queries per address - sometimes more. I've modified this so that now we perform one giant DB query per postcode which returns all the information we need for all the addresses in that postcode. This massively reduces the amount of time each script/thread spends performing database I/O in this phase of the import.

This results in a huge performance improvement (I just ran 117 import scripts on a quad core machine in 15 mins!). The tradeoff is that we sacrifice some readability for the sake of performance (it is not a pretty query).

Obviously not breaking the tests in `test_addressbase.py` goes some way to verifying the functionality has not changed, but for belt and braces I ran all the import scripts (117 of them) twice - once before this commit and once after, exported the resulting DBs and diffed the outputs against each other. They were binary equal so I'm confident this change does not introduce any new bugs.

It would be useful to get this merged before the next time I do an `imported-db` build.

TL;DR:
pretty:  :x:
fast:  :white_check_mark:
:D

### DIFF
--- a/polling_stations/apps/addressbase/helpers.py
+++ b/polling_stations/apps/addressbase/helpers.py
@@ -1,6 +1,7 @@
 import logging
 import re
 from collections import namedtuple
+from django.db import connection
 from pollingstations.models import (PollingDistrict, ResidentialAddress,
                                     PollingStation)
 from addressbase.models import Address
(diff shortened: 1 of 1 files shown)


symroe:
Excellent! I've not tried to understand the SQL ;)

PR #722, opened 2017-04-24
explicitly throw exception if EE returns non-200 status code

chris48s:


symroe:
Ooh, I'd not seen `raise_for_status` before :)

Issue #707, opened 2017-04-20
API Keys


chris48s:
Implement authentication/API keys so we know who is using our API/limit access if appropriate
Various considerations:
* How do people sign up for an API Key? MVP = 'email us for a key'
* We need to provide electoral commission/squiz notification to migrate - we can't just pull the rug
* Do we need to establish some kind of terms of service?
* Do we want to provide any kind of open/anonymous access? (e.g: open API is rate limited, unlimited access with a key)

chris48s:
next jobs following #730:

* Allow people to sign up for api keys themselves, somehow
* Formalise out a "don't be a douche" policy attached to provision of keys

PR #730, opened 2017-04-25
api keys

chris48s:
First things first:
* This requires proper review as it has the potential to break everything horrendously if I've done it wrong.
* We want to be ablsolutely sure we want to deploy this before it goes on `master`, so this might want to sit open for a bit

This PR does several things..

* Enables token-based authentication (can be in a header e.g: `Authorization: Token blabla` or a URL param (e.g: `?auth_token=blabla`)
* Adds a management command we can use to create users and tokens (not sure this is actually the best way of doing this, maybe this would be better done through `/admin`.. which we [just removed](https://github.com/DemocracyClub/UK-Polling-Stations/pull/658)! Aagh)
* Fiddles about with the DB routing (basically, if we create an account for someone we want it to still be there next time we deploy, so users and tokens need to live in the logger DB)
* Logs the API user in the `LoggedPostcode` table on calls to `/postcode` and `/address`
* BUT.. it makes no changes to permissions/access controls/rate limiting/etc: anonymous access will continue to work just fine

This is going to be a little bit fiddly to deploy in the context of https://github.com/DemocracyClub/polling_deploy/issues/13

Suggested deploy process is:
* Apply the migrations from this PR (explicit in `data_finder` app and implicit in `rest_framework.authtoken` app) to the prod logger DB
(We can apply these migrations and the existing code will still work.)
* Add users and keys for any users who we have given an API key to in advance
* Deploy a new server image with these code mods.

It is important that we make sure that any keys we have released to real users are in the DB at the point we deploy this because once we do, the API will accept anonymous calls and calls with a valid API key, but *invalid* API keys will throw `{"detail":"Invalid token."}`. So if we haven't got the accounts set up, any users pre-emptively using an API key will start getting `{"detail":"Invalid token."}`.

So.. where next:
Deploying this will get us to the stage where API keys are _actually doing stuff_ - next things we might want to think about include:
* Notifying users: how do we communicate to people who are already using the API that we will introduce keys?
* Adding stuff to docs about how to get/use API keys
* Actually _doing stuff_ with api keys (other than logging) e.g:
  * require auth/remove anon access
  * rate limit/restrict anon access
  * something else?
* How do users get a key? (MVP is email us, better would be signup form)
* 'Terms of Service'/'Acceptable use policy'
* If we require API keys, how do the docs/examples work?
...what else have I missed?


chris48s:
refs #707

chris48s:
So in response to the points at the end of my last post...

> Notifying users: how do we communicate to people who are already using the API that we will introduce keys?

usage is negligible - nobody seems to be doing anything major without an api key

> Adding stuff to docs about how to get/use API keys

:white_check_mark:


> rate limit/restrict anon access

:white_check_mark:

> How do users get a key? (MVP is email us, better would be signup form)

:white_check_mark:

> 'Terms of Service'/'Acceptable use policy'

not done..

> If we require API keys, how do the docs/examples work?

Alleviated by allowing anon usage with a rate limit


This needs to co-ordinate with some deploy changes though..

chris48s:
definitely leave this un-merged while we sort out redis/elasticache..

chris48s:
refs https://github.com/DemocracyClub/polling_deploy/pull/39

PR #738, opened 2017-04-27
add note saying stations may change between May and June

### LINKED ISSUE #734: Add note about May/June stations
Add a bit of copy that says something to the effect of

> This is your polling station for the local elections in May. It may change for the GE in June. Remember to check back

only better-worded

chris48s:
Closes #734


JoeMitchell:
Ello, this looks good. Does saying `check here again after 5 May` make it sound like we'll have an immediate update after the locals? 

Perhaps we could say `before 8 June` to give us more leeway?

PR #739, opened 2017-04-27
Deal with points not contained by any council boundary

chris48s:
If we hit a point which is not contained by the boundary of any council object:
- Try to look up the UPRN in the ONSAD
- If that also fails, handle it by skipping the address and logging a warning (instead of raising)

...because piers and stuff

screenshot at 2017-04-27 18 24 13


symroe:
I JUST PIER REVIEWED THIS
screenshot_116

Issue #743, opened 2017-04-27
MOAR performance improvements in import scripts


chris48s:
Full import was so fast.. then I fixed  #721 and now is *crazy* slow again - needs additional optimization or I wil go insane :(

### idea 1:
In `create_address_records_for_council()`, create a list of all the local authorities which touch the target local auth then limit the search space for the spatial join where we work out the local auth to a subquery of just those areas.
For example, if the target auth is Vale of Glamorgan our search space only needs to be Vale of Glamorgan, Bridgend, Cardiff and RCT, not the entire country!

### idea 2:
Try and do the lookup using ONSAD first instead of using the spatial join and then fall back to a spatial lookup only if the UPRN is not in ONSAD

chris48s:
Also more sophisticated approach to #720 - running these in series is also a bottleneck

In descending order of solution desirability:
* Find a way to let them all share the same files without treading on each others' toes so they can all run in parallel
* Download the files for each importer to a different location to they can all run in parallel and just accept you need to store the files N times
* The scripts don't actually need to run in complete isolation. You can 'interleave' Scotland scripts with England/Wales scripts, just not other Scotland scripts

chris48s:
Can you use `__slots__` to improve performance in import script classes?
https://www.python-course.eu/python3_slots.php
https://www.chrisbarra.xyz/posts/let-me-introduce-slots/

GeoWill:
WE don't do so much parallel importing anymore

PR #746, opened 2017-04-29
Improve bulk import performance

chris48s:
Use ONSAD to determine local auth in preference to `ST_COVERS()`. Only fall back to geo lookup if UPRN is not in the ONSAD.

Refs #743 but I'm not going to close it because there's probably more we can do to improve performance. This one change is probably the largest gain we can make here though.


symroe:
Not really going to pretend I properly understand this, but it seems like a reasonable approach, based on what you said on the hangout the other day!

PR #750, opened 2017-04-30
RFC: make choices in address select form more obvious (1)

chris48s:
So over in issue #367, @mtsk3000 has helpfully given us some feedback on the address select page from a UX perspective. I've had a go at implementing his idea pretty much exactly as suggested.

On balance, I reckon this is an improvement, but the one thing I'm not 100% sure on is:

If you hit the form for the first time and see something like this:

screenshot at 2017-04-30 20 00 10

does it make it look too much like 'My address is not in the list' is the default option?

@symroe (and anyone else who is interested) - can you pull it down and see what you think. Is this better or worse than it was before?

As a note, I've done this such that if the user hits the page and doesn't have javascript enabled, both buttons ar active (i.e: it works pretty much exactly how it does now) - so the javascript 'enhances' the functionality but it is not essential to the form's operation. I think the shorter button text also looks better on mobile.

### DIFF
--- a/polling_stations/templates/address_select.html
+++ b/polling_stations/templates/address_select.html
@@ -15,11 +15,22 @@
         {% csrf_token %}
         <div class="form-group">
             {{ form.as_p }}
-            <button type="submit" class="button" id="submit-address">
-              {% trans "I've selected my address. Find my Polling Station" %}
+
+            <p class="text-right">&nbsp;
+              <span id="clear-span" style="display:none;">
+                <a href="#" id="clear-button">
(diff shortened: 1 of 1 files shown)


mtsk3000:
"does it make it look too much like 'My address is not in the list' is the default option?"
Good point - further improvement could be making this a secondary button style (e.g. make that into a link style?), so that there is only one main CTA on the page.

symroe:
Could we add "my address is not listed" to the list of addresses?

chris48s:
Went with #800 instead

Issue #767, opened 2017-05-04
Declare a custom user agent string when calling APIs


chris48s:
e.g: Every Election, Mapit, etc

chris48s:
EE done in #773 This one is the most important because it allows us to track our own internal API usage.

Mapit needs its own custom UA

Other services we call are:
* YVM
* Google directions API
* Google geocoding API
* Morph

If we want to call them all with a user agent string which identifies us, it might be worth making a small wrapper over requests or mixin with a function `get_headers()` or something so we don't just end up repeating the block 

```python
headers = {}
if settings.CUSTOM_UA:
    headers['User-Agent'] = settings.CUSTOM_UA
```

all over the shop..

PR #773, opened 2017-05-07
Use timeout and custom UA when calling EE

chris48s:
Hopefully 3e21015 is sort of shutting the door after the horse has bolted, but this will ensure that if Every Election dies, WhereDIV can move on.

9d0cade makes it easier to for us to track our own API usage internally.

Refs #767

### DIFF
--- a/polling_stations/apps/data_finder/helpers.py
+++ b/polling_stations/apps/data_finder/helpers.py
@@ -303,7 +303,7 @@ class EveryElectionWrapper:
 
     def get_data(self, postcode):
         res = requests.get("%sapi/elections.json?postcode=%s&future=1" % (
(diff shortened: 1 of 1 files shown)


polling_stations/apps/data_finder/helpers.py
```diff
@@ -302,8 +302,12 @@ def __init__(self, postcode):
             self.request_success = False
 
     def get_data(self, postcode):
+        headers = {}
+        if settings.CUSTOM_UA:
```
symroe: `hasattr` here? As you're not adding `CUSTOM_UA` to the default settings…

chris48s: fixed

symroe:
👍

PR #800, opened 2017-05-10
RFC: make choices in address select form more obvious (2)

chris48s:
Alternative approach to #367 based on comment in https://github.com/DemocracyClub/UK-Polling-Stations/pull/750#issuecomment-298619070

Obviously this is a case of either #750 or this one, not both.


chris48s:
adding screenshot:

screenshot at 2017-05-18 19 38 41

PR #926, opened 2017-05-26
ID info for Northern Ireland - important!

BobHarper1:
In Northern Ireland, registered voters *do* need photographic identification to vote.

Therefore, I've added a message to alert NI voters to this, with a link to the Electoral Office guidance on acceptable forms of ID http://www.eoni.org.uk/Vote/Voting-at-a-polling-place

A better way to handle this might be to identify an NI voter through postcode (simple as all NI postcodes start with 'BT') but this is a quick and necessary fix.

### DIFF
--- a/polling_stations/templates/sidebar/you_dont_need_poll_card.html
+++ b/polling_stations/templates/sidebar/you_dont_need_poll_card.html
@@ -4,7 +4,7 @@
         <h3>You don't need your poll card to vote</h3>
         <p>If you don't have your poll card, you can go to the polling
             station and give them your name and address.</p>
-        <p>You don't need any other form of ID.</p>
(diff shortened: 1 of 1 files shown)


chris48s:
This is very useful information - thanks for raising it. I'd rather show the right information to the right users based on their location (like you've suggested) though. I'll have a look at doing something slightly more sophisticated later today.

BobHarper1:
☝️ I added a link to voting in northern ireland https://www.yourvotematters.co.uk/how-do-i-vote

The link to 'voting in the UK' is really only specific to GB - e.g. councils don't run elections in NI, only the Electoral Office does. The linked gov.uk page doesn't make this clear so confusing for user.

BobHarper1:
Thanks Chris, I would do but not running the site locally

chris48s:
Closing in favour of #928

PR #1009, opened 2017-06-05
Add link for cycling directions; fixes #1007

### LINKED ISSUE #1007: Add cycling directions (not just walking)
Would be cool to add cycling directions, not just walking. Ours uses open data, unlike the current walking provider :p

Either write out a link:
https://www.cyclestreets.net/link/formats/

or can draw on the map using the API:
https://www.cyclestreets.net/api/v1/journey/

Hoping to be able to use the DemocracyClub data in future for projects as previously discussed with @symroe at puntcon I think it was :)

mvl22:
Think this should work - it's a simple addition. I'm afraid I don't have a chance to set up a Django installation right now, so can't add the translation stuff, but hopefully this will be a quick change you might be able to slip in.


chris48s:
Thanks @mvl22. I've slightly tweaked the copy in 3cf4d3c and merged that.
I'll deploy it later or tomorrow.

PR #1029, opened 2017-06-28
assemble directions URL in nicer way

chris48s:
Tidy up the hack in #737 a bit

### DIFF
--- a/polling_stations/apps/data_finder/helpers.py
+++ b/polling_stations/apps/data_finder/helpers.py
@@ -399,9 +399,15 @@ class DirectionsHelper():
 
         return self.Directions(walk_time, walk_dist, ps)
 
+    def get_base_google_url(self):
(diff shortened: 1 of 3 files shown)


polling_stations/apps/data_finder/helpers.py
```diff
@@ -399,9 +399,15 @@ def get_ors_route(self, longlat_from, longlat_to):
 
         return self.Directions(walk_time, walk_dist, ps)
 
+    def get_base_google_url(self):
+        return "{base}&key={key}".format(
```
symroe: Will the call work if the key is `""`?

chris48s: Yep - you can call with `&key=&foo=bar` in the URL and it is the same as not passing `key`

PR #1030, opened 2017-06-28
Revert "Remove calls to EveryElection"

chris48s:
I removed the API call to EE for fear of it becoming a performance bottleneck.

Now that there aren't elections happening, this probably does serve a useful function so lets put it back..

More generally we probably want to work out how to scale this properly

### DIFF
--- a/polling_stations/apps/data_finder/views.py
+++ b/polling_stations/apps/data_finder/views.py
@@ -155,13 +155,11 @@ class BasePollingStationView(
         self.station = self.get_station()
         self.directions = self.get_directions()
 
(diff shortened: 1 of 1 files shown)


symroe:
> More generally we probably want to work out how to scale this properly

For Who I embedded the whole EE app on the image that span the servers up. At boot time they would fetch the latest DB and import it. Seeing as the content in EE is slow moving it's reasonable to do this on a cron each night. (and set the `EE_BASE_URL` to `localhost:8000` or whatever)

This method basically uses EE like any other software dependency and seemed to work quite well.

PR #1031, opened 2017-06-28
use consistent postcode formatting functions

### LINKED ISSUE #678: Consistent postcode formatting functions
There are a number of places in the code base where we are stripping whitespace out of postcodes, adding whitespace into postcodes, converting to uppercase, etc to deal with sources that use different formats. This leads to bugs e.g: PR #677

We need to rationalise this/deal with it more consistently.

chris48s:
Lets see if we can put an end to the postcode formatting madness

I've tested this but this change touches so many important things I bet it breaks something that I have not spotted. Hopefully if we do this now I will work out what that is when we're not in the grip of a really important election :)

Closes #678

### DIFF
--- a/polling_stations/apps/pollingstations/helpers.py
+++ b/polling_stations/apps/pollingstations/helpers.py
@@ -1,0 +1,10 @@
+import re
+
+
+def format_postcode_no_space(postcode):
+    return re.sub('[^A-Z0-9]', '', postcode.upper())
+
+
+def format_postcode_with_space(postcode):
+    pc = format_postcode_no_space(postcode)
+    return pc[:-3] + ' ' + pc[-3:]
--- a/polling_stations/apps/addressbase/helpers.py
+++ b/polling_stations/apps/addressbase/helpers.py
@@ -1,10 +1,10 @@
 import logging
-import re
 from collections import namedtuple
 from django.db import connection
 from councils.models import Council
 from pollingstations.models import (PollingDistrict, ResidentialAddress,
                                     PollingStation)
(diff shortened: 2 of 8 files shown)


polling_stations/apps/pollingstations/helpers.py
```diff
@@ -0,0 +1,10 @@
+import re
+
+
+def format_postcode_no_space(postcode):
+    return re.sub('[^A-Z0-9]', '', postcode.upper())
+
+
+def format_postcode_with_space(postcode):
+    pc = format_postcode_no_space(postcode)
+    return pc[:-3] + ' ' + pc[-3:]
```
symroe: I've been meaning to ask this for a while: is the outcode _always_ 3 chars? Seems like an odd bit of consistency!

symroe:
Might be worth adding some test cases what touch this? Or some dedicated cases for odd postcode inputs?

polling_stations/apps/pollingstations/helpers.py
chris48s: The bit before the space can be 2, 3 or 4 characters. As far as I'm aware, the last bit is always 3. Here's some links that agree with me:
https://www.getthedata.com/images/postcode_format.png
https://www.mrs.org.uk/pdf/postcodeformat.pdf

chris48s:
Yeah the 2 new functions are implicitly covered by various existing tests (hence the coverage has increased even though I have not added tests) but I will update the PR with some explicit test cases for the 2 new functions tomorrow.

Issue #540, opened 2017-03-11
tidy up postcode_view


chris48s:
In general, our templates are _fairly_ tidy, but over time as we have added this feature and that to [postcode_view](https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/templates/postcode_view.html) it has slowly evolved into a big old plate of spaghetti.

We should make a bit of time to make better use of blocks/includes to get it under control before it destroys us all.

chris48s:
https://github.com/DemocracyClub/UK-Polling-Stations/pull/1051#discussion_r137800902

PR #1037, opened 2017-07-14
[WIP] improve directions

chris48s:
It feels like quite a long time since I've written one of my pull request novels :) So what's goin on here then...

We have several directions-related problems that need solving:

* Our approach to directions just draws lines connecting points - we want curvy lines
* Google maps won't let us buy more than 100,000 requests per day
* The Open Route Service fallback code was broken so we had no fallback provider

To that end I have:

* Removed the broken Open Route Service code
* Modified the google maps code to use the `overview_polyline` (so we get smoothed shapes if we're using google maps)
* Added a directions client for [mapzen](https://mapzen.com/) (according to [mapzen's pricing](https://mapzen.com/pricing/#turn-by-turn) we should be able to buy as many requests as we need but it is worth checking)
* Made mapzen the default with fallback to google directions (if no mapzen key is set it will go straight to google maps without making a http request to mapzen) - this gives us a high-volume fallback and cuts down on the number of API keys you need to be able to start contributing because you can use GMaps with no key
* Refactored the `DirectionsHelper` class so it is easier to add or switch directions providers (using the same model as we've used for geocoding services)
* Added a whole bunch of unit tests (previously the `DirectionsHelper` was kind of covered by proxy in the integration tests but this is a lot more rigorous)

so whereas before, we were showing routes like this:

before

now we can show lovely curvy ones like this (mapzen):

mapzen

..and this (google):

google

woo :)

In order to test the mapzen stuff locally, add `MAPZEN_API_KEY = 'our-api-key'` to `local.py`. @symroe - I've sent you the login details on slack.

There are several additional things we need to do alongside merging this:

* Update [deploy repo](https://github.com/DemocracyClub/polling_deploy) to add a `MAPZEN_API_KEY` to the config file
* Add DC credit card details to mapzen account
* Contact mapzen about usage - are we definitely 100% ok to smash through >100,000 requests in a single day as long as we are giving them cashmonies?
* The other thing I would also like to address in this PR before merging (and with an accompanything PR to https://github.com/DemocracyClub/WhereDoIVote-Widget ) is that we've now got a bit of a random mashup of direction links going on. On the main site we link out to google maps for directions, but in the widget we're linking to OSM for directions and also a google maps link.. I'd like to standardise all that to either use google maps or OSM consistently (I'm happy for CycleStreets to be the odd one out given you're of the opinion that it is superior to GMaps or OSM for that particular purpose). I'm leaning towards changing all the links to OSM instead of google maps because "yay - open data!!" but google maps is probably more familiar to people. Thoughts on this??

Additionally, there are several issues arising from this PR which are out of scope and should be addressed elsewhere but are worth acknowledging here:

* As noted in e8ad46a our approach to JS dependency management in this project is a bit rubbish and there's a mix of stuff we've just commited to the repo and some NPM dependencies (PhantomJS, Aglio) which aren't properly managed. Its worth looking at whether we can make a better job of this (probably with `yarn` but needs a bit of thought)
* All [this](https://github.com/chris48s/UK-Polling-Stations/commit/caa748f7b9161d65ccdfb01164fca753f2eeb1d7#diff-2163c5f843c83763dbbb8f664d402cd2) is a bit of a mess but should be addressed under #540
* Both of the above points tie into looking at our javascript pipeline for this project, but I think that fits under https://github.com/DemocracyClub/polling_deploy/issues/6 or maybe with upgrading the theme?
* Both of the directions APIs we're using do return the info we would need to actually embed directions on the page. If you look at https://maps.googleapis.com/maps/api/directions/json?mode=walking&units=imperial&key=&origin=53.52608322352942,-2.288074417647058&destination=53.521097882352954,-2.292880429411764 for example, there are instructions like `Head <b>south</b> on <b>Lowther Rd</b> toward <b>Lowther Cl</b>` in the JSON and Mapzen does the same. I think for the improvements in this PR are enough for now, but maybe we should look at using these to provide inline directions in future?

### DIFF
--- a/polling_stations/apps/data_finder/directions_clients.py
+++ b/polling_stations/apps/data_finder/directions_clients.py
@@ -7,7 +7,8 @@ from django.conf import settings
 from django.utils.translation import ugettext as _
 
 
-Directions = namedtuple('Directions', ['walk_time', 'walk_dist', 'route', 'precision'])
+Directions = namedtuple('Directions', [
+    'walk_time', 'walk_dist', 'route', 'precision', 'source'])
 
 
 class DirectionsException(Exception):
@@ -62,7 +63,8 @@ class GoogleDirectionsClient(DirectionsClient):
             directions['routes'][0]['legs'][0]['distance']['text']
         ).replace('mi', _('miles'))
 
-        return Directions(walk_time, walk_dist, json.dumps(route), self.precision)
+        return Directions(
+            walk_time, walk_dist, json.dumps(route), self.precision, 'Google')
 
 
 class MapzenDirectionsClient(DirectionsClient):
@@ -120,4 +122,5 @@ class MapzenDirectionsClient(DirectionsClient):
             round(directions['trip']['summary']['length'],1)
         ) + _(" miles")
 
-        return Directions(walk_time, walk_dist, json.dumps(route), self.precision)
+        return Directions(
+            walk_time, walk_dist, json.dumps(route), self.precision, 'MapZen')
--- a/polling_stations/apps/data_finder/directions_clients.py
+++ b/polling_stations/apps/data_finder/directions_clients.py
@@ -1,0 +1,117 @@
+import abc
+import json
+import requests
+import urllib
+from collections import namedtuple
+from django.conf import settings
+from django.utils.translation import ugettext as _
+
+
+Directions = namedtuple('Directions', ['walk_time', 'walk_dist', 'route', 'precision'])
+
+
+class DirectionsException(Exception):
+    pass
+
+
+class DirectionsClient(metaclass=abc.ABCMeta):
+
+    def __init__(self):
+        pass
+
+    @abc.abstractmethod
+    def get_route(self, start, end):
+        pass
+
+
+class GoogleDirectionsClient(DirectionsClient):
+
+    precision = 5
+
+    def get_base_url(self):
+        return "{base}&key={key}".format(
+            base=settings.BASE_GOOGLE_URL,
+            key=settings.GOOGLE_API_KEY
+        )
+
+    def get_route(self, start, end):
+        url = "{base_url}&origin={origin}&destination={destination}".format(
+                base_url=self.get_base_url(),
+                origin="{0},{1}".format(start.y, start.x),
+                destination="{0},{1}".format(end.y, end.x),
+            )
+
+        resp = requests.get(url)
+        if resp.status_code != 200:
+            raise DirectionsException("Google Directions API error: HTTP status code %i" % resp.status_code)
+        directions = resp.json()
+
+        if directions['status'] != 'OK':
+            raise DirectionsException("Google Directions API error: {}".format(directions['status']))
+
+        route = directions['routes'][0]['overview_polyline']['points']
+
+        walk_time = str(
+            directions['routes'][0]['legs'][0]['duration']['text']
+        ).replace('mins', _('minute'))
+
+        walk_dist = str(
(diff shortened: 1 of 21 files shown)


symroe:
[Improve directions API](https://trello.com/c/S3lzzf3S/879-improve-directions-api)

symroe:
> I'm leaning towards changing all the links to OSM instead of google maps because "yay - open data!!" but google maps is probably more familiar to people. Thoughts on this??

Ideally I'd like to be able to use something like `mailto:`: `navigateto:` or `map:` would be nice, but this isn't really the place to riase HTML feature requests ;)

[This SO question](https://stackoverflow.com/questions/9688607/how-to-open-a-mobile-devices-map-app-when-a-user-clicks-on-a-link) seems to have some good ideas. Maybe we look at doing some device snooping and change the URL in JavaScript depending on the user's device? (this is a suggestion for a future issue, not this one, and something that @mrwilson might be interested in doing?)

My reason for suggesting this: I think we should link to _the most native mapping tool_ for the user. That is, we should get the users default navigation app opened quickly. OSM is amazing, but I expect it's not the default mapping service for most people.

symroe:
> maybe we should look at using these to provide inline directions in future

Yes, I love this, as it would be more usable by people with sight problems (I expect – untested)

polling_stations/apps/data_finder/directions_clients.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class DirectionsException(Exception):
+    pass
+
+
+class DirectionsClient(metaclass=abc.ABCMeta):
+
+    def __init__(self):
+        pass
```
symroe: Still don't have my head around `abc`, but is this correct? Should it call `super`?

polling_stations/apps/data_finder/directions_clients.py
```diff
@@ -0,0 +1,129 @@
...
+import json
+import requests
+import urllib
+from collections import namedtuple
+from django.conf import settings
+from django.utils.translation import ugettext as _
+
+
+Directions = namedtuple('Directions', [
+    'walk_time', 'walk_dist', 'route', 'precision', 'source'])
```
symroe: Future proofing question: should we add the transport method here, and rename `walk_[time|distance]` to `[time|distance]`? If we want to add driving times in the future (for example, when the `walk_time` is greater than ~15 minutes) then this namedtuple isn't useful.

symroe: (Applies below too of course)

polling_stations/apps/data_finder/helpers.py
```diff
@@ -348,115 +349,19 @@ def get_explanations(self):
...
-                except OrsDirectionsApiError as e2:
-                    # Should log error here
-
-                    directions = None
-            else:
-                directions = None
-
-        return directions
+                    return client.get_route(kwargs['start_location'], kwargs['end_location'])
+                except DirectionsException as e:
```
symroe: Small point but I think this will make flake8 complain about an unused var

chris48s:
I've added commits which address your inline comments. I've raised a trello card for 'Add DC credit card details to mapzen account' and I'll submit a PR to https://github.com/DemocracyClub/WhereDoIVote-Widget to change the directions link to google maps over there. Assuming you're happy to merge this, I'll go through and raise issues for all the other ideas/future work identified above.

https://github.com/DemocracyClub/polling_deploy/pull/61 is a pretty trivial change.

symroe:
Nice one :)

Issue #1047, opened 2017-08-13
postcode_location isn't a valid GeoJSON feature


chris48s:
```
{
  "type": "Feature",
  "geometry": {
    "point": {
      "type": "Point",
      "coordinates": [...]
    }
  }
}
```

isn't valid GeoJSON

this should be

```
{
  "type": "Feature",
  "properties": {},
  "geometry": {
    "type": "Point",
    "coordinates": [...]
  }
}
```

chris48s:
remember impact on docs and downstream consumers

Issue #1036, opened 2017-07-14
Error code for invalid postcode


chris48s:
At the moment a `400` can be:

* postcode is not valid or
* postcode is valid but some other error happened

That means API consumers (including us) have to parse the error string, like this:

https://github.com/DemocracyClub/WhereDoIVote-Widget/pull/21/commits/c56cdc81150d72d15d1aa5b5d0b5b3e8cb04ce0b

to show sensible error messages.

Invalid postcode should have its own error code to make this a bit less nasty.

chris48s:
remember impact on docs and downstream consumers

chris48s:
This was actually an error message that was getting passed through from mapit. Now that we've removed mapit, both `api/beta/postcode/foo.json` and `api/beta/postcode/SA981AA.json` just throw `400 Bad Request` `{"detail":"Could not geocode from any source"}`.

The front-end uses `django-localflavour` to picture check user input first.. need to have a think about whether the API should do that too for consistency or whether it is the client app's responsibility to validate if it wants to. Hmm :thinking:

GeoWill:
Should this checking happen in aggregator api?

PR #1048, opened 2017-08-14
Use django-apiblueprint-view for API docs instead of aglio

### LINKED ISSUE #963: API should behave consistently with front-end where we can not geocode
e.g:
https://wheredoivote.co.uk/postcode/M15JB
https://wheredoivote.co.uk/api/beta/postcode/M15JB.json

chris48s:
I decided a while ago that I'm not really happy with aglio as a renderer for API blueprints:

* There's a few features it doesn't have that would be useful (e.g: support for default `null` in MSON schema) which I've had to work around
* In order to style/theme it, we have to use Jade for templates and LESS for the CSS which completely different to the tech we're using for our theme
* We have to compile the docs rather than just rendering them on-the-fly (hence the [cumbersome update process](https://github.com/DemocracyClub/UK-Polling-Stations/blob/ec1e2a41a3e4d172a4f271758282d95eaf0d4a17/polling_stations/templates/api_docs/README.md) )
* It doesn't seem to be very actively maintained/developed

so I looked at various other solutions, decided that none of them really tick the right boxes for me - all of them seem to be missing something important.. several weekends/evenings of yak shaving later I've ended up writing my own API blueprint rendering library which does all the things I want it to. It renders on-the-fly natively using django templates and is framework-agnostic about how you style/theme the outputs.

This PR removes the dependency on aglio and replaces it with [django-apiblueprint-view](https://github.com/chris48s/django-apiblueprint-view) and adds a template to style the API docs consistently with the rest of the website.

The slight downside is that [drafter](https://github.com/apiaryio/drafter/) (the parser library I'm using) isn't packaged as a .deb so we need to compile it from source on ubuntu but you should be able to just do `brew install --HEAD https://raw.github.com/apiaryio/drafter/master/tools/homebrew/drafter.rb` on Mac OS to review this PR. I'll submit a PR to https://github.com/DemocracyClub/polling_deploy to install drafter in the server build process.

The other thing I want to do is rewrite the finder API examples using MSON schema so they are generated from [data structures](https://github.com/DemocracyClub/UK-Polling-Stations/blob/ec1e2a41a3e4d172a4f271758282d95eaf0d4a17/polling_stations/templates/api_docs/blueprints/data_structures.apibp) rather than being defined in this [more verbose fashion](https://github.com/DemocracyClub/UK-Polling-Stations/tree/ec1e2a41a3e4d172a4f271758282d95eaf0d4a17/polling_stations/templates/api_docs/blueprints/finder_responses), but I think I'm going to leave them as they are for now and then update all the examples in the docs when I've fixed issues #963, #1036 and #1047 which will change the API outputs a bit anyway and require documentation updates.


.travis.yml
```diff
@@ -20,9 +26,20 @@ cache:
 env:
   - MORPH_API_KEY=FakeApiKey # set a fake morph api key so the system checks pass
 install:
-  - pip install pip==8.1.2
+  - sudo unlink /usr/bin/g++ && sudo ln -s /usr/bin/g++-5 /usr/bin/g++
+  - pip install --upgrade pip
   - pip install tox coveralls tox-travis
   - npm install
+  - wget https://github.com/apiaryio/drafter/releases/download/v3.2.7/drafter-v3.2.7.tar.gz
```
symroe: It's annoying there isn't a `.deb` for this, as compiling for each build will take up a lot of time. 

Is there a way to cache the build? Maybe adding `~/drafter-v3.2.7/build` to the cache dir in travis and checking for `build/out/Release/lib.target/libdrafter.so` before running `configure` and `make`?

Alternatively, how hard is it to make a `.deb` and host it on S3? Or just get the project to manage one after a PR from us?

chris48s: Yes - it is a bit of a pain. They do produce [an `aur` package for Arch Linux ](https://aur.archlinux.org/packages/drafter/) but no `deb`. I suspect that packaging for Brew and Arch reflects the preferences of the developers who maintain it. I have had a cursory look at docs for debian packaging at one point.. based on that, I think build caching is probs the first thing I'll look at ;)

chris48s:
Good suggestion on caching - I think on balance [this](https://github.com/DemocracyClub/UK-Polling-Stations/pull/1048/commits/4602054f213be285f97764d54fe6d28a54dae30a) is easier than trying to package our own `.deb` :). I did that on a dev branch, builds took 7:01 and 7:37. Pushed a junk commit and the next 2 took 4:55 and 5:40 so it seems to work.

I'm going to need to rebase this again to account for #1062 so merge that one first

chris48s:
Updated

Its a good job fixing merge conflicts in yaml files is so much fun :D

symroe:
Oh I merged this forgetting the conversation – I think we can speed up Travis more by not building or configuring at all if it's been cached:

```
[ -e 'build/out/Release/lib.target/libdrafter.so' ] || /usr/bin/python2.7 configure --shared
[ -e 'build/out/Release/lib.target/libdrafter.so' ] || make libdrafter
```

(Untested)

chris48s:
Hmm.. I tried changing it to `test -e build/out/Release/lib.target/libdrafter.so || make libdrafter` and it still runs `make`, which is really odd because adding `$HOME/drafter-v3.2.7/build/` to the cache dirs does seem to have made a meaningful performance improvement :confused:

PR #1051, opened 2017-09-05
improve handling of javascript and css

### LINKED ISSUE #412: Latest dc_base_theme breaks stuff

### LINKED ISSUE #1041: Sort out JS dependency management
There's a couple of node dependencies (PhantomJS, Aglio) and various bits of client-side JS (note e8ad46a ) which we should be doing a better job of managing.

Have a look at using yarn or npm to tidy this up

### LINKED ISSUE #1042: Re-evaluate javascript pipelining
Refs #1040 and https://github.com/DemocracyClub/polling_deploy/issues/6

Is the current approach right?
Should we be bundling?
Is there JS we can move out of templates and into .js files?

chris48s:
Top-level summary:

* Move from `django-static-precompiler` and `ManifestStaticFilesStorage` to `django-pipeline` and `PipelineCachedStorage`
* Upgrade base theme to `v0.2.2`
* Remove fallback.js and stop using CDNs
* Split css/js into 2 pipelines for performance reasons
* Use `npm` for js/css dependency management
* Delete loads of stuff we don't need + various bits of tidying
* Move some javascript out of template and into js file (Refs #540 but there is more to do on that)

more details in the commit messages. Diff is a huge, but most of it is just removing third party code from the repo in af03468 , 06726a2 and a025718

Closes #1042
Closes #1041
Closes #412

Note:
I haven't bothered to move `aglio` into `package.json` and update the docs for that as I am assuming that we're going to merge https://github.com/chris48s/django-apiblueprint-view/pull/10 and https://github.com/DemocracyClub/UK-Polling-Stations/pull/1048 at some point (not top priority) which will ultimately just remove that as a dependency.

Further work needed to finish this PR:
- [x] Fix the integration tests. I've tried shoving various combinations of `STATICFILES_STORAGE = 'pipeline.storage.NonPackagingPipelineStorage'`, `STATICFILES_STORAGE = 'pipeline.storage.PipelineStorage'`, `PIPELINE_ENABLED = False` etc into `testing.py`. None of it works. Not sure quite where to go with it but I want to get these fixed as they are useful.
- [x] Upgrade theme to `v0.2.3` and remove https://github.com/chris48s/UK-Polling-Stations/blob/e9c698526ea2c5e1dac2b2614b29e1cf0e774278/polling_stations/settings/static_files.py#L5-L7 once https://github.com/DemocracyClub/dc_base_theme/pull/20 merged.
- [x] Submit a complimentary PR to https://github.com/DemocracyClub/polling_deploy updating the static files config and adding code to install js/css dependencies with `npm` in the setup process.
- [x] Run `collectstatic` in the travis build somewhere? I think making sure it doesn't throw an exception is a useful check.
- [x] Test on a staging deploy ( @symroe - need to have a chat about how to do this without screwing up prod, related to https://github.com/DemocracyClub/polling_deploy/issues/6 )


chris48s:
refs https://github.com/DemocracyClub/polling_deploy/pull/78

polling_stations/templates/postcode_view.html
```diff
@@ -5,7 +5,11 @@
 {% load pipeline %}
 
 {% block extra_page_css %}
-{% stylesheet 'map' %}
+
+{% if we_know_where_you_should_vote and station.location %}
```
symroe: Lovely! Really only loading the JS if we need it.

Not for now, but I wonder if the view should give this logic to the template with something like `can_show_map` in the context? Anyway, this is great!

symroe:
Lovely! Good to go!

Also by the way, this was great to review with commits nicely sized/scoped :)

PR #1054, opened 2017-09-12
Replace example postcode with static example view

### LINKED ISSUE #575: Dynamic example postcodes
The problem:

We have a hard coded example postcode on the home page. This only works if we have data imported for the current election for that postcode.

Also, in the white-label or embedded versions, it would be nice to have a postcode for that area.

Suggestions:

1. Pull the postcode in to settings and set that when deploying.
2. Maintain a list of postcodes we know we have data for (by grabbing a postcode from the addresses, or picking a random postcode in a district for shapes). Then we can have a list of postcodes, one per area. We can then pick them randomly on the home page (or pick once per server load, e.g. keep it in memory).
3. For embedded, allow passing an example postcode in the query string of the URL
4. For white-label, configure this per white label site

chris48s:
Closes #575 (by making it irrelevant)

TODO:

* ~Remove example postcode stuff from deploy repo~ https://github.com/DemocracyClub/polling_deploy/pull/81
* Link to `/example` from project page on DC website somehow


symroe:
Really nice that the view is just a subclass of the 'real' view :)

PR #1061, opened 2017-09-25
don't log to google analytics when running integration tests

### LINKED ISSUE #1053: Running integration tests logs hits to google analytics
I was briefly confused about why the most visited URLs on our site are `/address_select/BB11BB/`, `/multiple_councils/DD11DD/` and so on..

chris48s:
Closes #1053

### DIFF
--- a/polling_stations/apps/pollingstations/context_processors.py
+++ b/polling_stations/apps/pollingstations/context_processors.py
@@ -1,0 +1,7 @@
+from django.conf import settings
+
+
+def google_analytics(request):
+    return {
+        'disable_ga': getattr(settings, 'DISABLE_GA', False)
+    }
--- a/polling_stations/settings/base.py
+++ b/polling_stations/settings/base.py
@@ -122,6 +122,7 @@ TEMPLATE_CONTEXT_PROCESSORS = (
     'dc_theme.context_processors.dc_theme_context',
     'dc_signup_form.context_processors.signup_form',
     'feedback.context_processors.feedback_form',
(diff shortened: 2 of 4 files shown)


symroe:
Happy with this, but just a thought: it is worth only having one condition in the template (`{% if not disable_ga %}`) and then doing `DISABLE_GA = DEBUG` in the same way as `TEMPLATE_DEBUG` does in core Django?

chris48s:
`aloe_django` (or something in that stack) automatically sets `DEBUG=False` off as part of the test bootstrap, which is a sensible thing to do for an end-to-end test but its not very explicit that it does it.. I'm struggling to find an explicit reference to it in the docs (`aloe_django` is a bit thin on docs - its kind of obliquely mentioned in passing here http://aloe.readthedocs.io/projects/aloe-django/en/latest/porting.html ) but that's why I had to do [this](https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_finder/features/index.py#L23-L27) in the integration tests when we switched to pipleined storage.

symroe:
Right, so the line in `testing.py` stays the same, but the template code gets a little simpler/explicit.

chris48s:
I think it stands as it is:
* If `DEBUG=True` (i.e: we're developing locally), we don't want to log to GA
* If `DEBUG=False` and `DISABLE_GA=True` (i.e: we're running an integration test), we don't want to log to GA

symroe:
👍

Issue #820, opened 2017-05-13
Update terminology ONSAD --> ONSUD


chris48s:
ONS seem to have rebranded their ONS Address Directory product as ONS UPRN Directory:
http://ons.maps.arcgis.com/home/item.html?id=5a0d38f408ae426eb7f309bb0a9d452d
http://ons.maps.arcgis.com/home/item.html?id=c1db046e367d497c92ffbe2f2a23e22f

This is probably a more accurate name but it means all our terminology is now all wrong. Also affects:
https://github.com/DemocracyClub/polling_deploy
~https://github.com/chris48s/ONS-Geography-Releases-Scraper~

chris48s:
Finished off in https://github.com/DemocracyClub/polling_deploy/pull/93

PR #1063, opened 2017-09-29
Split up AddressBase app

chris48s:
This piece of work supports several high-level objectives:
* Make UPRNs a first-class citizen in WhereDIV
* Retire the DC mapit
* Develop shared components for use in multiple DC projects

One of the first steps towards these is to take the existing code in the AddressBase app in this project and separate the bits that are specific to WhereDIV from the bits that are generic/re-usable. That's the main focus of this PR, but I've also taken the opportunity to make a few other improvements while I'm working on it..

Quick summary of changes
---

* Remove some stuff we can do without
* Split code from `addressbase` app into `uk_geo_utils` (generic) and `addressbase` (WhereDIV-specific)
* Rename ONSAD model to ONSUD (refs #820)
* Implement `uk_geo_utils` models as abstract models so we can extend them in client apps to avoid using one-to-one relationships (and some fiddling with tooling to support this)
* Improve address formatting code
* Refactoring and tidying up
* Improve automated testing

More details in the commit messages

At this point, all of the code in the `uk_geo_utils` (better name suggestions welcome, btw) app is now generic and potentially applicable to other projects. Everything that is left in the `addressbase` app is specific to WhereDIV.

Next steps for `uk_geo_utils`
---

* Before we spin this out into a separate package, there are some other bits of this project that should sensibly move into the `uk_geo_utils` app, like: [this](https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/pollingstations/helpers.py), [this](https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_finder/helpers.py#L258-L261), [this](https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_finder/helpers.py#L283-L317)<sup>1</sup>, and maybe some other things.. I'd like to do this "in place" as a PR to this repo, but this PR is already quite large, so I reckon it is best to review this and then deal with those as the next chunk of work.
* Spin the `uk_geo_utils` app (again, better name suggestions welcome) out into a standalone package and pull it into this projects as a dependency (in much the same way as I've done with https://github.com/DemocracyClub/dc_signup_form )
* I also want to add in an ONSPD model/import script, rewrite the [geocoders](https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_finder/helpers.py#L39-L255) to work [like this](https://gist.github.com/chris48s/3fc6b354dec4de6ae7d85b029f7ef5d1) and move them into the `uk_geo_utils` package. I'm not sure whether I want to also do that as a PR to this repo, or whether it makes more sense to consider that bit of work in the wider context of other client apps (taking into account the conversations we've had about EE's use-case).. I'll think about that in more detail when I get to it..

Next steps for `addressbase`
---

All of the code left in `addressbase` is code I plan to remove as part of the "Making UPRNs a first-class citizen in WhereDIV" task, so what remains in this app is not long for this world:

* The blacklist model will become irrelevant - we won't need to blacklist postcodes because we'll be able to deal with postcodes containing UPRNs in >1 local auth just fine, so [this](https://github.com/chris48s/UK-Polling-Stations/blob/improve_addressbase/polling_stations/apps/addressbase/models.py#L21-L32) and [this](https://github.com/chris48s/UK-Polling-Stations/blob/improve_addressbase/polling_stations/apps/addressbase/management/commands/create_blacklist.py) will be deleted.
* All of the code in [here](https://github.com/chris48s/UK-Polling-Stations/blob/improve_addressbase/polling_stations/apps/addressbase/helpers.py) and [here](https://github.com/chris48s/UK-Polling-Stations/blob/improve_addressbase/polling_stations/apps/addressbase/models.py#L5-L14) addresses edge cases that we will handle in a simpler way, so all of that will be removed.
* Address will become a core model, implemeted as

```python
from uk_geo_utils.models import AbstractAddress

class Address(AbstractAddress):
    council = models.ForeignKey(Council, blank=True, null=True, related_name='+')
    polling_station_id = models.CharField(blank=True, max_length=255)
```

in the `pollingstations` app, so we'll remove [this](https://github.com/chris48s/UK-Polling-Stations/blob/improve_addressbase/polling_stations/apps/addressbase/models.py#L17-L18)

Footnotes
---

<sup>1</sup> I'm actually in 2 minds about this.. I considered changing the AddressBase model so that it stores the addresses in the database like this:

```python
organisation_name = models.CharField(blank=True, max_length=60)
department_name = models.CharField(blank=True, max_length=60)
po_box_number = models.CharField(blank=True, max_length=6)
sub_building_name = models.CharField(blank=True, max_length=30)
building_name = models.CharField(blank=True, max_length=80)
building_number = models.CharField(blank=True, max_length=5)
dependent_thoroughfare = models.CharField(blank=True, max_length=80)
thoroughfare = models.CharField(blank=True, max_length=80)
post_town = models.CharField(blank=True, max_length=30)
double_dependent_locality = models.CharField(blank=True, max_length=35)
dependent_locality = models.CharField(blank=True, max_length=35)
```

and then uses the `AddressFormatter` class to concatenate it all together on-the-fly to display on the basis that storing it like that might make it easier to sort, but because all the 'number' fields might be "32A" or "89 to 91", the sorting is still gnarly even if you do that. I've decided to ignore this for now and just carry on with our approach of mashing it all into one `address` field then breaking the addresses apart with regex, sorting them and sticking them back together again but I may revist this idea one day..


polling_stations/apps/addressbase/management/commands/import_cleaned_addresses.py
```diff
@@ -6,17 +6,18 @@
 
 class Command(BaseCommand):
 
+    table_name = 'addressbase_address'
```
symroe: Still reading this PR so you might address this later. Are you thinking about implementing something like the `auth.User` model whereby we can set the `Address` model in settings?

I'm thinking something like this in `settings.py`:

`address_model = 'polling_stations.models.AddressWithPollingStation`

If you think that's a good pattern then this line should be something like:

`get_address_model_somehow().objects.model._meta.db_table`

polling_stations/apps/addressbase/helpers.py
```diff
@@ -232,3 +232,114 @@ def create_address_records_for_council(council, batch_size, logger):
...
+
+        i) First and last characters of the Building Name are numeric
+          (eg '1to1' or '100:1')
+        ii) First and penultimate characters are numeric, last character is
+          alphabetic (eg '12A')
+        iii) Building Name has only one character (eg 'A')
+        """
+        if element[0].isdigit() and element[-1].isdigit():
+            return True
+        if element[0].isdigit() and element[-2].isdigit() and element[-1].isalpha():
```
symroe: Might `element[-2]` raise an `IndexError`?

polling_stations/apps/addressbase/models.py
```diff
@@ -19,14 +19,13 @@ class AbstractAddress(models.Model):
...
     location = models.PointField(null=True, blank=True)
-    objects = AddressManager()
 
     class Meta:
         abstract = True
 
 
 class Address(AbstractAddress):
-    pass
+    objects = AddressManager()
```
symroe: Can you explain the reason for this? Does this mean that all sub-classes of `AbstractAddress` will need to do this?

polling_stations/apps/addressbase/management/commands/import_onsud.py
```diff
@@ -12,6 +12,8 @@
 """
 class Command(BaseCommand):
 
+    table_name = 'addressbase_onsud'
```
symroe: See the comment about this pattern with the address model.

polling_stations/apps/addressbase/management/commands/import_cleaned_addresses.py
chris48s: Yeah - I like where you're going with this. I did try to think of a way to infer the table name from the model, but wasn't sure how to specify it. Having an app setting which defaults to `Address` but can be overridden with `AddressWithPollingStation` or whatever seems obvious now you say it - this is why code review is good :). That way we can set this once rather than having to pass it as a param in multiple places.

The issue I can see is that I think at the time the settings are parsed, Django hasn't loaded the ORM/models yet so I guess the setting will have to be something like `ADDRESS_MODEL = 'uk_geo_utils.models.Address'` as opposed to `ADDRESS_MODEL = Address()` and then I'll have to go from the string representation to a class (if you see what I mean). I'm not 100% sure what the implementation looks like at the moment but I'll look into it as I think that a better approach.

polling_stations/apps/addressbase/helpers.py
chris48s: Good question/point - I probably need to add a `len()` check here.

polling_stations/apps/addressbase/models.py
chris48s: The reason for moving the manager declaration on to the implementation in that commit is because the methods `postcodes_for_district()` and `points_for_postcode()` are specific to WhereDoIVote (they're also marked for removal in future work - the `AddressWithPollingStation` model won't use them) so I didn't want to transfer them to the `uk_geo_utils` app. You would need to add that declaration to any implementation of `AbstractAddress` which should inherit those methods, but `AbstractAddress` itself shouldn't have them.

In a [later commit](https://github.com/DemocracyClub/UK-Polling-Stations/pull/1063/commits/98603881f176ae426d9589bfd82f5e9df620cf31) I [added a `uk_geo_utils.models.AbstractAddressManager` class](https://github.com/DemocracyClub/UK-Polling-Stations/pull/1063/commits/98603881f176ae426d9589bfd82f5e9df620cf31#diff-9ab660099325f43819c4b7cc51ee6780) with a method that all `Address` models should have and [modified `addressbase.models.AddressManager` to extend it](https://github.com/DemocracyClub/UK-Polling-Stations/pull/1063/commits/98603881f176ae426d9589bfd82f5e9df620cf31#diff-2a693a38b300d697ef456543e04467c2)

polling_stations/apps/addressbase/management/commands/import_onsud.py
chris48s: :+1:

polling_stations/apps/addressbase/management/commands/import_cleaned_addresses.py
symroe: It might be worth researching 'swappable' https://code.djangoproject.com/ticket/19103 https://github.com/wq/django-swappable-models, but IDK. Django does this for `auth.User` and `Profile` but I've seen other apps do this sort of thing too (I'll try to remember some examples)

polling_stations/apps/addressbase/models.py
symroe: 👍

chris48s:
Updated as requested. Turns out this was not too tricky.

PR #1065, opened 2017-10-05
uk_geo_utils app part 2

chris48s:
* Move `AddressSorter` class to `uk_geo_utils` app (straightforward lift and shift)
* Move postcode functions into `uk_geo_utils` app and encapsulate in a `Postcode` data type (more involved).

Obviously moving from functions to a `Postcode` class involves quite a lot of code changes to accommodate it, but I think on balance it is quite a neat abstraction and gives us some nicer notations.

I've also consciously given the class [this characteristic](https://github.com/chris48s/UK-Polling-Stations/blob/92df1e890e5adc5474a2276c3862e9b65ae84e36/polling_stations/apps/uk_geo_utils/tests/test_postcode_helper.py#L55-L73), partly because it made it easier to port stuff, but also because I think its quite useful that we can call `RoutingHelper()`, `geocode()` or `EveryElectionWrapper()` with either a string or an existing `Postcode` object and they'll behave the same way.

The other thing this abstraction gives us is the ability to add additional properties/methods on postcode and move some common application logic that appears in several of our sites on to this component. For example, we might add methods like:

```python
@property
def is_valid(self):
    # validate postcode against a regex e.g:
    # https://github.com/django/django-localflavor
    # https://github.com/mysociety/uk-postcode-utils
    return True/False

@property
def exists(self):
    # check if postcode is in addressbase/onspd table
    return True/False
```

and we might also modify our constructor to look something like

```python
def __init__(self, postcode, validate=False):
    if len(str(postcode)) < 5:
        raise ValueError("Postcode must have at least 5 characters")
    self.postcode = re.sub('[^A-Z0-9]', '', str(postcode).upper())
    if validate and not self.is_valid:
        raise ValueError("Postcode is not valid")
```

and then that again allows some quite nice succinct notations. I've not done any of this yet because I think its better to add these things as we actually need them rather than guess what we might need (YAGNI), but I think its useful to create the right structure to slot these things into as I refactor.

The other thing that's probably worth thinking about at this point is the format/naming of `.with_space` and `.without_space`. Do you think its right for them to be decorated as properties, or is it better to think of them as `.format_with_space()` and `.format_without_space()`? I know its a minor point, but I reckon it is worth getting the public surface of this class right because we're going to end up using this in a lot of places.


polling_stations/apps/uk_geo_utils/helpers.py
```diff
@@ -16,6 +16,29 @@ def get_onsud_model():
...
+class Postcode:
+
+    def __init__(self, postcode):
+        self.postcode = re.sub('[^A-Z0-9]', '', str(postcode).upper())
+
+    def __str__(self):
+        return self.without_space
+
+    @property
+    def territory(self):
```
symroe: FYI, there's a Ruby Gem that works out the territory (and does other things) here:

https://github.com/threedaymonk/uk_postcode/blob/master/lib/uk_postcode/country_lookup.rb

Might be worth taking a look at it's API for ideas.

I say this because I'm not sure 'GB' is a territory technically – I've used it in EE to mean a 'sub-national area'. Maybe it doesn't matter too much, but I was assuming that `.territory` used elsewhere in this commit would return one of ENG, SCT, WLS, NIR.

polling_stations/apps/uk_geo_utils/helpers.py
```diff
@@ -16,6 +16,29 @@ def get_onsud_model():
     return get_model('ONSUD_MODEL', 'uk_geo_utils.Onsud')
 
 
+class Postcode:
```
symroe: I have some general questions about the scope of `Postcode`. Obviously what it does at the moment is fairly simple, but should it work out if an input is valid or not? Should it raise on instantiation with an invalid postcode?

If it attempts to validate, should it just work out the format, or existence?

symroe: Ah, I see in a later commit you've introduced one check that raises an generic `ValueError`.

polling_stations/apps/uk_geo_utils/helpers.py
```diff
@@ -19,6 +19,8 @@ def get_onsud_model():
 class Postcode:
 
     def __init__(self, postcode):
+        if len(str(postcode)) < 5:
```
symroe: Shouldn't this be _after_ stripping spaces and other non-alphanumerics?

chris48s: Yes - good spot

polling_stations/apps/uk_geo_utils/helpers.py
symroe: Sorry, ignore this – I've seen you wrote about this in the description :/

chris48s: That's exactly what I'm getting at in the first post of this PR in the section

> The other thing this abstraction gives us
...
> but I think its useful to create the right structure to slot these things into as I refactor

polling_stations/apps/uk_geo_utils/helpers.py
chris48s: If we want to actually work out what country (or local authority, or ward..) a postcode is in, that should be handled in the `Geocoder` class(es) rather than trying to overload everything into the `Postcode` class.

I've put this in here because there's a convenient shortcut/picture check we can do but maybe the best way is to say this check should just go back into the WhereDIV specific code ( https://github.com/DemocracyClub/UK-Polling-Stations/pull/1065/files#diff-0e31cd62c86a0e4133b1d5916504b644L258 ) and it doesn't belong in the `Postcode` class.

Trying to use a regex to work out if a postcode is in Wales (or whatever) is not a good idea for several reasons.

1. It doesn't work: For example, here are some postcodes which contain UPRNs which are in both England and Wales:
SY156AU
SY156DF
SY156DS
SY156EB

2. Even if it did work, you've got to attempt to "maintain" hundreds of characters of regex or generate it each time a new addressbase comes out or whatever. This is what the `Geocoder` classes are there to save us from.

symroe:
[WhereDIV new data model](https://trello.com/c/5ysTspzL/981-wherediv-new-data-model)

PR #1070, opened 2017-10-30
uk_geo_utils app part 4

chris48s:
* Model for ONSPD
* Import script for ONSPD
* OnspdGeocoder class (and temporary Adapter class)
* Remove MapitGeocoder
* Update tests

Although I've swapped the ONSPD geocoder in place of Mapit for the moment, once we update the data model to key off UPRN I think I'm going to end up removing this from WhereDIV. At this point, this is more about finishing off the `uk_geo_utils` app so we can spin it out into a standalone app for wider use.

There's some further questions I've got about edge case handling which I think are worth considering at this point. These are more applicable to the `uk_geo_utils` app in the abstract rather than specifically about our usage in WhereDIV (where the range of geography layers we care about is very constrained).

1. In both the ONSPD and ONSUD there are some pseudo-codes that are used to indicate that a layer of geography is not applicable to a given UPRN or postcode. For example, if you request the county field for a postcode/UPRN in Wales (where there are no counties), the value is set to `W99999999` + various varaitions on this theme for the various supported geographies. For the moment I've chosen not to handle this in any way (e.g: by wrapping it in an exception) in the geocoders. I'm just returning the pseudo-code and leaving the client to handle it.

2. In the ONSPD _only_ (this isn't the case with the ONSUD because every record corresponds to a currently valid UPRN in Great Britain) there are various situations where the postcode has a `NULL` grid ref or code: Some terminated postcodes, postcodes that are for some reason non-geographic, postcodes which apply to territories where certain data is not available (Hello my old friend https://mapit.democracyclub.org.uk/postcode/JE37DW.json - we meet again) etc. For the moment I've taken the decision to apply no special handling in the `OnspdGeocoder` and just pass through whatever is in the DB, leaving the calling code to handle it as appropriate (e.g: https://github.com/DemocracyClub/UK-Polling-Stations/pull/1070/commits/26ad9fc40b2bc39eccde8a580d33ba40cef0a156#diff-0e31cd62c86a0e4133b1d5916504b644R144 ).

Leaving the `OnspdGeocoder` as a very thin wrapper over the ONSPD is simpler and also more flexible (because different client code can choose to handle these conditions as appropriate for each application), but is also complicates calling code meaning that we (and potentially other users, but one step at a time...) have to implement some edge case handling in code that consumes this API which is maybe not immediately obvious.

@symroe - any thoughts on these design decisions? Maybe we can discuss it on Monday. I'll chew it over a bit more over the weekend myself..


chris48s:
[Where Do I Vote? new data model](https://trello.com/c/5ysTspzL/981-where-do-i-vote-new-data-model)

polling_stations/apps/uk_geo_utils/models.py
```diff
@@ -98,3 +98,114 @@ class Meta:
...
+    def _get_hlthau(self):
+        return self.oshlthau
+
+    def _get_rgn(self):
+        return self.gor
+
+    def _get_ruc11(self):
+        return self.ru11ind
+
+    cty = property(_get_cty)
```
symroe: No problems here, but what's the thinking behind this pattern rather than just using `@property` on `def cty(self):`?

polling_stations/apps/data_finder/helpers.py
```diff
@@ -118,6 +122,61 @@ def geocode_point_only(self):
...
+    """
+
+    def geocode(self):
+        geocoder = OnspdGeocoder(self.postcode)
+        centre = geocoder.centroid
+        if not centre:
+            raise PostcodeError("No location information")
+
+        local_auth = geocoder.get_code('lad')
+        error_values = [
```
symroe: Hide Easter eggs here :)

polling_stations/apps/uk_geo_utils/models.py
chris48s: Now you raise it, I think I just looked up how to alias fields on a django model and found a code example that used the function form of `property` rather than the decorator form of `property`. These could probably be rewritten using the decorator form.

symroe: Interesting they've done it that way! I wonder if this is because there's issues with sub-classing a property?

PR #1071, opened 2017-10-31
Import councils using ONS data instead of mapit

chris48s:
This is a step towards removing our dependencies on mapit and retiring the DC mapit install. It also makes the councils import a lot quicker because we can fetch all the boundaries we need with 2 HTTP requests instead of having to make 391 requests.

~One question on this: For now I have declared the URLs to get the boundaries from as app settings [here](https://github.com/chris48s/UK-Polling-Stations/blob/e5c5347265f46f4d6dabc20c73cfec7972259423/polling_stations/settings/constants/councils.py#L4-L5) but I could set it up so that we set them in the deploy scripts (like [this](https://github.com/DemocracyClub/polling_deploy/blob/master/vars.yml#L8-L9)) for consistency if you prefer. I think it is probably more reasonable to just define these as constants and update them occasionally because local auth boundaries change relatively infrequently whereas AddressBase, ONSAD etc change much more often but I don't have strong opinions on it.~

Actually I think I've answered my own question in d3e492b: because the field names aren't necessarily stable across releases it is better to declare as app constants because we may need to modify the import script when we update the URLs

Note: deploying this will require us to apply a migration to the logger DB.


chris48s:
[Retire the DC Mapit, Where Do I Vote? new data model](https://trello.com/c/5ysTspzL/981-retire-the-dc-mapit-where-do-i-vote-new-data-model)

symroe:
Nice one!

Issue #1092, opened 2017-11-27
Upgrade to django 2.0


chris48s:
Django 1.8 is an LTS release and still gets security patches but it is EOL in April 2018 - see https://www.djangoproject.com/download/

We need to sort out how to migrate at some point but we are definitely relying on some functionality which is deprecated in more recent versions so it is a non-trivial task.

There is also the question of whether we should target 1.11 or wait a bit and try to jump straight to 2.0?

@symroe - probs worth having a chat about this one at some point.

party98:
Is this up for grabs?

party98:
It seems that 2.0 is available now

chris48s:
Hi @party98 Thanks for showing interest in this. You're right - Django 2.0 was released yeterday.

In terms of this project, we've yet to decide on the migration path.

I've not reviewd the release notes for 2.0 yet, but at the moment I'm kind of leaning towards 1.11. It is a LTS release and it will probably be a lower bar to migrate the actual application to 1.11 (i.e: less non-BC changes).

Additionally we rely on a number of django-based dependencies (either third party, or which we maintain ourselves - see https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/requirements/base.txt ). Most or all of them are tested under 1.11. If we want to go to 2.0 we would need to be in a place where we know that all of our dependencies are Django 2.0-ready and we are definitely not there right now.

In terms of the work needed to migrate our application, its not just a case of bumping the version and running the tests. As a minimum, we'll need to:

* Review all our dependencies for compatibility - if there are compatibility blockers, we may need to update packages we maintain ourselves (e.g: https://github.com/DemocracyClub/dc_base_theme , https://github.com/DemocracyClub/dc_signup_form , https://github.com/chris48s/django-apiblueprint-view ) or wait for/submit PRs to/switch out third-party packages with compatibility issues.
* Make code updates based on known deprecations/BC-breaks (e.g: `from django.core.urlresolvers import reverse` --> `from django.urls import reverse`, update `SessionMiddleware` + so on)
* The way we are providing the embedded and whitelabel versions of our site rely on a deprecated use of `django.conf.urls.patterns()` which is removed in Django 1.10 onwards. We will need to work out how to implement this functionality in a newer version of the framework.
* While we do have good test coverage, there is a bunch of code that is not under test (see https://coveralls.io/github/DemocracyClub/UK-Polling-Stations ) which will need manual review for compatibility.

[...]

PR #1093, opened 2017-11-27
update ALL THE PACKAGES

chris48s:
I've been intending to

* prune unnecessary dependencies from our requirements files and
* do a better job of keeping our dependencies up-to-date

for some time. It occurrs to me that upgrading loads of packages now is probably a good idea - as opposed to, for example, in April. We're not in 'critial' deployment and if I'm doing a load of work on the site locally, I should hopefully pick up anything subtle that has been missed in the tests. To keep us on track moving forwards, I've pinned everything to the latest version (or latest version we can use with django 1.8 in some cases*) and I want to add [PyUpBot](https://pyup.io/) to the repo. PyUpBot will monitor our dependencies and every time a new version of a library we are using is release it will submit a PR [like this](https://github.com/chris48s/UK-Polling-Stations/pull/32) to the repo updating the requirements file with the changelog in the body of the PR. This will make this easier to manage an review moving forwards so we can keep on top of updates (this is one of the reason for trying to get more code under test). I've included the propsed PyUp config in this PR, but I won't add the integration to the repo until we've merged this - otherwise it will just go mental and submit 30 PRs telling us everything is out of date.

I've pre-emptively rebased this onto the changes in PR #1091 so that

* more tests get run
* they actually pass

but review that one first. If you want changes done, I'll do those first and rebase this again.

*Side issue: Django 1.8 is an LTS release but it is EOL in April 2018 so we need to sort out how to migrate at some point but I'm going to say that is out of scope for this PR as that is a large issue. I've raised it as #1092

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -1,0 +1,18 @@
+update: all
+pin: True
+schedule: "every day"
+search: True
+
+requirements:
+  - requirements/base.txt:
+      update: all
+      pin: True
+  - requirements/local.txt:
+      update: False
+      pin: False
+  - requirements/production.txt:
+      update: all
+      pin: False
+  - requirements/testing.txt:
+      update: False
+      pin: False
--- a/polling_stations/apps/data_collection/views.py
+++ b/polling_stations/apps/data_collection/views.py
@@ -1,4 +1,3 @@
-import requests
 from datetime import datetime
 from django.conf import settings
 from django.db.models import Case, IntegerField, Q, Value, When
--- a/polling_stations/apps/data_finder/features/index.py
+++ b/polling_stations/apps/data_finder/features/index.py
@@ -9,8 +9,27 @@ from django.conf import settings
 from django.core.management import call_command
 from aloe import before, after, around, step, world
 import aloe_webdriver.django
-from selenium import webdriver
+from selenium.webdriver import PhantomJS
+from selenium.common.exceptions import TimeoutException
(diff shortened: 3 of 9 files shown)


chris48s:
oh, and I did clear the pip cache before submitting this PR :)

symroe:
Great stuff, and good shout on PyUpBot – I'd not heard of it before. Worth adding it to our other repos in time too.

chris48s:
> PyUpBot – I'd not heard of it before

Yeah the thing that seems quite handy about this approach is that you can pin your dependencies so you have deterministic builds but the bot will automate the process of keeping track of new releases and testing new dependencies before upgrading. Whereas if you specify `requests >=2,<3` or something in your requirements, you'll always be on the latest 2.x version but there's always a risk you'll deploy and something will break because you've unexpectedly upgraded a dependency and there's a bug in the latest release or whatever.

> Worth adding it to our other repos

We should work towards that, but in order to keep on top of it you do need quite good tests in place to make sure you're not breaking stuff as you update, which is why I've gone for this repo first as I think WhereDIV has the most extensive test coverage of our repos. Also it has the potential to increase noise and testing overhead, so you want to make sure that every dependency it is monitoring as actually used for something (hence removing all the redundant/transitive dependencies first) - so there's a bit of setup to do first.

This will be the first repo where I've used it in anger anyway, so we'll see how it goes.

PR #1101, opened 2017-11-30
Use station UPRN in Xpress importers

### LINKED ISSUE #1039: Use Polling_Place_UPRN in BaseXpressDemocracyClubCsvImporter
I'm not sure if I've ever seen a situation where `polling_place_uprn` is populated and `polling_place_easting`/`polling_place_northing` aren't, but in principle, `get_station_point()` should prioritise geocoding based on UPRN above postcode centroid if that case occurs.

chris48s:
It is relatively unusual, but if we've got a station UPRN but no grid ref, try UPRN to geocode in preference to postcode so we have a more accurate point if possible.

Closes #1039

### DIFF
--- a/polling_stations/apps/data_collection/ems_importers.py
+++ b/polling_stations/apps/data_collection/ems_importers.py
@@ -3,10 +3,12 @@ Specialised base import classes for handling data exported from
 popular Electoral Management Software packages
 """
 import abc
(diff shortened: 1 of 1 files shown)


chris48s:
re: coverage drop, I've raised #1099 about adding tests for the EMS importers. The Xpress/Halarose import classes are very frequently used so should be under test.

Issue #1045, opened 2017-07-27
Add more automated checks to import process


chris48s:
A lot of the overhead of this project near to elections is around basic consistency checks on the data. We should automate as many of these as possible. We should throw some kind of error or warning when:

- [x] ~A station point is outside target local auth~ (done in PR #1034 )
- [x] A polling district wholly or mostly falls outside the target local auth boundary (refs #666, #667) (PR #1203 )
- [x] A `ResidentialAddress` falls outside the target local auth boundary (how do we do this quickly? Checking every address one-by-one is going to be really slow)
- [ ] The data we've imported only covers part of the local authority (need a proxy for this e.g: wards with 0 addresses in them)
- [x] ~Number of records 'looks right' based on total number of addresses in local auth (bit 'finger in the air' but should be able to work out a plausible-ish percentage and warn if below)~ #1122

..anything else?

symroe:
> A ResidentialAddress falls outside the target local auth boundary 

Maybe with http://postgis.net/docs/ST_Disjoint.html?

chris48s:
It would also be useful to be able to detect/fix polling distirct boundaries which don't clip precisely to local auth boundaries, but this is a soft error - sometimes it matters more than others.
e.g: if a boundary in a coastal local auth clips to the coastline but we've got a full extent boundary then that is a different class of error than a polling district for a London borough grabbing a couple of streets from its' neighbouring local auth.

chris48s:
* `ResidentialAddress` record has incorrect postcode attached (can we check input postcodes against AddressBase where input record has UPRN?) This will at least help to catch some cases.

chris48s:
> A polling district wholly or mostly falls outside the target local auth boundary

Use https://postgis.net/docs/ST_PointOnSurface.html + check if point is inside or outside (centroid is not a good plan because C-shaped or U-shaped polygons may have a centroid outside the polygon).

chris48s:
> A polling district wholly or mostly falls outside the target local auth boundary

given 2 geometries `pollingdistrict`, `localauth`: if `localauth` doesn't contain or intersect, `pollingdistrict` its wholly outside. If it contains its wholly inside. If it intersects, `intersection = pollingdistrict.intersection(localauth)` gives you the bit of `pollingdistrict` inside `localauth` and you can compare `pollingdistrict.area` to `intersection.area` to work out the size of the overlap and report it.

chris48s:
[...]

PR #1122, opened 2018-01-03
Show info about expected dwellings in local auth before import 

chris48s:
Refs #1045

This PR adds the ability to print some contextual information which gives us some idea how many addresses we would expect to find in the input file if it is complete. Because this adds some overhead to the start of the script, I've set it so it doesn't run if we pass the `--nochecks` flags (this means we won't run this code when we do a bulk import to build an image and we can pass it over in unit tests, etc).

I don't really have a good handle yet on which of these numbers is the most useful comparison point or what a sensible warning range is going to be. I think 'Total Dwellings from 2011 Census' is actually going to be the closest number mostly (but a bit on the low side because its fairly out-of-date now) because both of the UPRN totals include non-residential addresses whereas the Census number is only residential dwellings. Predictably none of the numbers we can easily generate or get hold of neatly map on to the exact number of records we would expect in the input file.

Either way, having these numbers gives us a good picture check - if we've got an incomplete file or a file that contains data from >1 local auth, its should be much more obvious now. Maybe once I've written a bunch of input scripts I'll have a feel for a reasonable warning threshold.

### DIFF
--- a/polling_stations/apps/addressbase/migrations/0008_auto_20180102_1606.py
+++ b/polling_stations/apps/addressbase/migrations/0008_auto_20180102_1606.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('addressbase', '0007_delete_onsud'),
+    ]
+
+    operations = [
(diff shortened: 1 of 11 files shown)


symroe:
Nice!

And nice API find – I'd not seen that before. It would be good to use it elsewhere, like getting stats in EE.

Anyway, that's a different thing. This looks good to go.

PR #1124, opened 2018-01-03
upgrade to Django 1.11

chris48s:
Not ready for review yet..

Refs #1092

### DIFF
--- a/polling_stations/apps/pollingstations/tests/test_urls.py
+++ b/polling_stations/apps/pollingstations/tests/test_urls.py
@@ -1,0 +1,30 @@
+import json
+from django.test import TestCase
+from django_extensions.management.commands.show_urls import Command
+
+
+class UrlTests(TestCase):
+
+    def is_exception(self, url):
+        exceptions = ['.txt', '.ics']
+        for exception in exceptions:
+            if exception in url:
+                return True
+        return False
+
+    def test_trailing_slashes(self):
+        c = Command()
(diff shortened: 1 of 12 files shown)


chris48s:
Having chewed over #1092 a bit, I reckon the sensible thing to do here is to upgrade to Django 1.11 first because:

* All of our dependencies work with it today - we're not waiting on anything
* It gets us on to a LTS release
* Everything we need to do in order to upgrade to 1.11 is a step on the road to 2.0 anyway so its a helpful stepping stone

I _think_ this PR now covers everything we need to do in order to upgrade to Django 1.11, but I would like to leave the PR open for a bit before merging to `master` and I'll continue doing some more local maunal testing with this configuration as I do some dev work.

Note: deploying this requires applying some migrations to the logger DB

chris48s:
I don't want to load a bunch of unrelated changes into this PR but since we were talking about it this morning in the context of changing URL patterns in this PR, I've tacked ce1e49e on to this.

polling_stations/apps/pollingstations/tests/test_urls.py
```diff
@@ -0,0 +1,30 @@
+import json
+from django.test import TestCase
+from django_extensions.management.commands.show_urls import Command
+
+
+class UrlTests(TestCase):
```
symroe: Nice! :D We should use this everywhere.

Issue #1040, opened 2017-07-17
Update to latest DC theme


chris48s:
..to bring visual appearance into line with DC website and WhoCIVF

chris48s:
Doing this will also move us to a single column layout, at which point we should review our information hierarchy (particularly where opening times fits in)

chris48s:
Also when we do this, drop font-awesome for a lighter solution. Once we're onto the latest theme we will be serving up quite a large icon set to only use 2 icons out of it (the house and the tick in the 2 map markers for the home point and polling station point).

chris48s:
Doing this will require moving from a 2 col layout to a single col layout. Reference https://democracyclub.org.uk/wheredoivote_user_feedback/ when doing so for "information hierarchy" notes.

chris48s:
also note to self

https://github.com/DemocracyClub/UK-Polling-Stations/blob/5c17f6ae925c06b7353270641dcf85f04a1538e8/requirements/base.txt#L18-L23

chris48s:
closed in #1137, tracking further issues in #1206 and #1207

Issue #741, opened 2017-04-27
Fonts on home page


JoeMitchell:
![image](https://cloud.githubusercontent.com/assets/8096305/25502563/01e6c81a-2b8f-11e7-8c04-79caca0d3359.png)

I think all this text should be the same font. 

The two lines behave differently (see below) when squished to mobile too... they should behave the same way! 

![image](https://cloud.githubusercontent.com/assets/8096305/25502647/64113142-2b8f-11e7-9457-d9976bd9fd6e.png)

chris48s:
Given we want to redo the styling on this site so it is all a bit more


![](https://democracyclub.org.uk/static/dc_theme/images/logo-with-text-2017.png)


I'm going to suggest we park styling tweaks and revisit as part of that process..

JoeMitchell:
Could a quick fix just be to apply the same style to both lines?

chris48s:
review after #1040

chris48s:
closed in #1137

PR #1137, opened 2018-02-18
Upgrade to DC_base_theme 0.3

symroe:
This is very much the start of #1040

Still do to:

* [x] Remove Foundation `<div class="row">` type tags. These are mostly in templates that extend from `base.html`
* [x] Add the postcode form to the context so we can use `{{ postcode_form|dc_form }}` in templates rather than writing the form 'by hand'
* [x] Think about the design and information hierarchy esp of the "polling station found" (see linked issue)
* [x] …and much more including browser testing etc.
* [x] Fix the embedded Version
* [x] Remove the [uglify dependency from package.json](https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/package.json#L12)
* [x] Convert `menulist` JS to pure JS 
* [x] Do all tasks on https://github.com/DemocracyClub/dc_base_theme/projects/1?add_cards_query=is%3Aopen first
* [x] Merge https://github.com/DemocracyClub/dc_base_theme/pull/37 first


symroe:
On the information hierarchy, as is with the theme update and no other thought, the example URL looks like this:

screenshot-2018-2-18 where do i vote

chris48s:
Cheers for making a start on this.

In terms of 'information hierarchy', lets ignore moving stuff about for the moment and just deal with the 'nuts and bolts' of the theme update here and then address re-organising the content as its' own issue (even if it puts some stuff in a worse place). I reckon we maybe want to look at how to present the voter id stuff in that mix too.

Keeping those separate will hopefully keep things easier to review.

symroe:
Yep, that's reasonable. The good news is that I don't think it's actually _much worse_ as a single column – I expect more people will read the whole column and not worry about the order than will fail to see the right hand column.

Not to say that it doesn't need improving, but I think it's not actually worse.

chris48s:
btw, I suspect the reason the build is failing on this PR is probably because you haven't updated the requirements, so its pulling in the old theme using

https://github.com/DemocracyClub/UK-Polling-Stations/blob/5c17f6ae925c06b7353270641dcf85f04a1538e8/requirements/base.txt#L23

and then failing on the integration tests trying to use your new front-end code to render the page. Also when you update that remember to clear the travis cache before building.

chris48s:
Right.. done some fairly extensive testing on this. I'm going to split the review into multiple posts to attempt to break the content up into slightly more manageable chunks. There are quite few things to point out here so I think its worth prefacing the review with a couple of points:

1. In general things are looking good:
    * This layout is going to make improving the information hierarchy and mobile UX easier
    * There are a number of other things that work in a nicer way in this theme, for example I particularly prefer the styling of messages to confirm form submission
    * There were a few things on my list of checks that I expected might be pain points in the upgrade which were not an issue :)
    * Doing this also fixes some other minor design issues we've been ignoring like https://github.com/DemocracyClub/UK-Polling-Stations/issues/741
2. There might be some things on this list where we choose to say "well, its not worse then the corp site and WhoCIVF so lets merge it and raise an issue on the theme repo". While I'm picking it apart and attempting to break the design it is worth mentioning them all here even if we choose not to fix/improve all of these issues right now.

chris48s:
# Some form fields overflow parent divs

1. I like how the address picker expands/shrinks in the desktop layout depending on the line length, for example:

screenshot at 2018-03-13 13 35 31

screenshot at 2018-03-13 13 35 53


but some combinations of long address/narrow window can cause overlap e.g:

screenshot at 2018-03-13 13 34 58

2. On my (android) phone the address picker box always renders off the side of the screen regardless of how long the addresses are (this is scrolled all the way to the right - there's a horizontal scrollbar):

screenshot_20180313-150105

This always puts the arrow off the right-hand side of the screen. If you click(touch) it, the actual address picker does render sensibly though.

3. Here's some examples of text boxes rendering outside div/off screen on (simulated) lower resolution devices:

screenshot at 2018-03-13 14 31 59

screenshot at 2018-03-13 14 57 06

I think form elements should generally be constrained to rendering inside their parent div. I'd rate this group of related issues as fairly important to fix.

css_is_awesome

chris48s:
# Form Validation Issues

There's a couple of form validation issues:

1. I'm not terribly keen on the way the form displays on invalid postcode:

screenshot at 2018-03-13 13 11 50

screenshot at 2018-03-13 13 31 01


In general I'm not opposed to way the validation errors are displayed, for example I think the way they render here looks good:

screenshot at 2018-03-13 14 05 41

but in the 'invalid postcode' case, I think the way it displays feels a bit clumsy because of the centred field and repeated similar text with the same styling. I also accept that it is not worse than what we have on WhoCIVF (see https://github.com/DemocracyClub/WhoCanIVoteFor/issues/213 )

Can you think of an improvement we could make here (and also apply on WhoCIVF)?

2. This one is a bit of an edge case, but in some situations (no js turned on, client side validation passed but server side validation failed) it is possible to try and submit the feedback form with neither 'yes' or 'no' selected. The easiest way to reproduce this is to hit `/feedback/` and just press 'send feedback'. Using the v0.2 theme, this will throw a validation error on the client side using HTML validators. With the v0.3 theme, it just does nothing - it doesn't work and doesn't give any error/reason. If this is lots of effort to fix, I think it is low priority here because in most cases the conditions that lead you here are rare but it probably applies to any for where you you're using this button/radio input mashup pattern so is worth being aware of as an issue.

chris48s:
# Embedded Version

I know we are encouraging councils to use the JS version as the default embed option, but there will be local auths that do use `/embed` in an `<iframe>` because that's what they've done in the past. We should make sure this works well.

1. As noted in Slack, the content to logo ratio here is a bit much ;)

screenshot at 2018-03-13 12 58 07


suspect https://github.com/DemocracyClub/dc_base_theme/pull/29/commits/7d5a03a1d5416981a7ac3f59831b035346cc23e1#diff-07a0ed53124a587430c358056f8d4b14 is the cause.

2. More generally, looking at the embedded version it seems like you've removed the styling entirely - even stuff like div widths and link colours. Do you have specific thoughts on this, or is that oversight?

I'd rate this as fairly important to look at as I suspect that some subset of local authorities will present to their users in this way.

chris48s:
# API Documentation


I realise you haven't been able to test any of this, but there are a few things I'd like to improve in the API docs:

2. Is there a style in the v0.3 theme that I can use to differentiate the HTTP method text a bit? I'd like to style the text `GET` here so it contrasts a bit:

screenshot at 2018-03-13 13 54 41

like this:

screenshot at 2018-03-13 13 55 11


2. I think the way the inline `<code>` style in the v0.2 theme used a different background colour provided some useful visual differentiation for code snippets which is lacking with the v0.3 theme and helped break things up a bit:

screenshot at 2018-03-13 13 53 27

screenshot at 2018-03-13 13 54 05

Everything is a bit.. black and white. I'd like to do something to make the inline `<code>` blocks a bit more obvious. Would your preference here be to deal with that as an inline style in these docs or in the base theme?


3. Not sure why but the toggle links don't have a link cursor with the v0.3 theme, whereas they did with the v0.2 theme. Any thoughts on what I'd need to change here? It might be one of the styles I'm defining in the [base template](https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/templates/api_docs/api_base_template.html) that is interacting differently:

gifrecord_2018-03-13_135640

gifrecord_2018-03-13_135751


If it is painful for you to get drafter 3.2.7 installed and you want me to just deal with these issues, let me know your thoughts and I'm happy to do that work.

chris48s:
# A few more minor things

1. Could the error page for 403/500 have a header that links back to the home page like the 404 page does:

500/403 page:



404 page:



(this may have also been an issue on the old theme)

2. Feedback form needs a bigger text box - it looks really odd with big buttons and a tiny text box:

screenshot at 2018-03-13 16 09 00

I made this small because the old theme used little radio buttons here so it looked odd with tiny buttons and a huge text box :)

3. The NUS Wales themed site looks a bit strange now. Given we have now let wheredoivote.wales , bleibleidleisio.cymru etc expire my preference here is to just bin the nus_wales app/routes as part of this piece of work rather than attempt to fix anything. If we want to update the translations and revive the Welsh language version of the site, we can do it under the DC branding.

4. The league table view is much less readable using the v0.3 theme:

screenshot at 2018-03-13 13 11 49

screenshot at 2018-03-13 13 10 41

I don't think there's any point trying to improve this but I think the action points here are:

* Remove the routes for `/league_table` and `/coverage`. Neither of them provide any useful information to anyone and fixing/improving either of them is not high on my list of things to do, or yours. If we ever want to resurrect these ideas, we can also think about design.
* As an abstract point, the v0.3 theme could benefit from having better a table style. The screenshot there with the v0.2 theme is much more readable.

polling_stations/settings/static_files.py
```diff
@@ -9,8 +8,7 @@
     ],
     extra_js=[],
 )
-PIPELINE['JS_COMPRESSOR'] = 'pipeline.compressors.uglifyjs.UglifyJSCompressor'
-PIPELINE['UGLIFYJS_BINARY'] = 'node_modules/uglify-js/bin/uglifyjs'
```
chris48s: `jsmin` seems to work fine here :+1: You can also remove the uglify dependency from `package.json` here https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/package.json#L12 if we're going to use something else.

polling_stations/templates/address_select.html
```diff
@@ -1,40 +1,32 @@
...
+    </button>
+  </form>
 </div>
 
+
 {% endblock content %}
 
 {% block in_page_javascript %}
 <script type="text/javascript">
-$(document).ready(function() {
```
chris48s: This seems to work fine everywhere I tested it, but I'm just a bit confused by this change and the commit message for it. Your commit comment is "Don't require jQuery for raw JS" but one of the 4 lines of JS wrapped by this is `$('.select_multirow').css('-webkit-appearance', 'menulist');` which is using jQuery.

Also I was under the (possibly misguided) impression that anywhere you're using `$('thing').stuff()` it needs to be wrapped in a `ready()`.

This comment is a request for clarification/gap in my knowledge rather than a request for change.

requirements/base.txt
```diff
@@ -15,11 +15,9 @@ requests==2.18.4
 boto==2.48.0
 
 
-git+https://github.com/mysociety/django-pipeline-csscompressor
-
 # TODO: review this when you upgrade dc_base_theme to 0.3
-libsass==0.13.6  # pyup: >=0.13.6,<0.13.7
+libsass==0.13.7  # pyup: >=0.13.6,<0.14.0
```
chris48s: Does upgrading to libsass `>=0.14` do bad stuff with the 0.3 theme? If so, lets raise an issue on the theme repo. Whereas I wasn't particularly inclined to fix the legacy theme, I'm more inclined to care about this in the context of the theme we're actively developing/migrating projects to.

chris48s:
# ...and finally

1. As noted in Slack, we'll need to tag a 0.3.6 release of the theme and update this once https://github.com/DemocracyClub/dc_base_theme/pull/29 is merged regardless of whether there are any further changes.
2. Can you think of anything we're doing here that impacts on the deploy process? I can't think of anything, but if there is anything we need to update in the deploy repo, let me know.
3. Did you have any thoughts on https://github.com/DemocracyClub/UK-Polling-Stations/issues/1040#issuecomment-328117072 ? I'd lumped it under the theme work, but if you want to ignore it and do that as a seperate piece of work, we can do.

Sorry - I am aware I've just submitted the war & peace of code reviews there :(

chris48s:
> If it is painful for you to get drafter 3.2.7 installed

Thinking about it, is this as simple as just running
* `brew install https://raw.githubusercontent.com/apiaryio/drafter/55614cdb017370915d21ed7a2c9b1fd487297423/tools/homebrew/drafter.rb` or
* `brew install --HEAD https://raw.githubusercontent.com/apiaryio/drafter/55614cdb017370915d21ed7a2c9b1fd487297423/tools/homebrew/drafter.rb` ?

(I have no idea what the difference between those 2 would be)

I'd test it, but.. y'know :penguin:

polling_stations/templates/address_select.html
symroe: 1. Whoops, I missed that call. I've changed it. FYI this is the first part in a project to remove jQuery as a dependency from our projects, or more generally get used to not using it. It's slightly pointless on this project as the other JS we pull in is going to require it.

2. `ready()` just prevents the JS from running before the HTML has rendered / downloaded. For a lot of cases the need for this goes when you put the JS below (in the source order) the HTML it's expecting. I think this is a good call for this case as the user will want the menulist to work before all images etc have loaded. Really we want to convert the style of the menulist as early as we can, so waiting for the document to be 'ready' isn't helpful.

requirements/base.txt
symroe: Tested and it seems to work 👍

polling_stations/templates/address_select.html
```diff
@@ -1,40 +1,41 @@
...
-    var isAndroid = /(android)/i.test(navigator.userAgent);
-    if (isAndroid) {
-        $('.select_multirow').css('-webkit-appearance', 'menulist');
-    }
-});
+  var isAndroid = /(android)/i.test(navigator.userAgent);
+  if (isAndroid) {
+    var elements = document.querySelectorAll('.select_multirow');
+    var className = 'android_menu_list';
+    Array.prototype.forEach.call(elements, function(el, i){
```
chris48s: I know you've got to go back to Android <= 4.1 and IE9 for browsers that don't support `forEach()` but do you reckon it would be better to use a more backwards-compatible syntax here?

polling_stations/templates/address_select.html
chris48s: > remove jQuery as a dependency from our projects

Lets not get bogged down in it here as there is enough to discuss but can we earmark this as something to talk about in a bit more detail - would be useful to hear your thoughts on it. Maybe Tuesday would be a good time for that?

PR #1138, opened 2018-02-20
copy edit

chris48s:
There are no upcoming elections in your area --> We don't know of any upcoming elections in your area


symroe:
You're changing it in this file rather than in 

https://github.com/DemocracyClub/UK-Polling-Stations/blob/1bff228945069a14d4383753accfbcaf44e8521b/polling_stations/apps/data_finder/views.py#L158

because the error text is used elsewhere? Wouldn't it be better to change the error text at source?

PR #1145, opened 2018-02-26
Add 'report a problem' form

### LINKED ISSUE #139: Add a "report problems" feedback form
From @chris48s: 

> Each /postcode/ page should have a 'report an issue with this page' button on it which takes the user to the form where they can:
> - request removal of a boundary
> - request removal of a polling station
> - submit a corrected address for a polling station
> - provide a contact email address?
>   Submitting that form inserts a record (and probably fires off an email to alert an administrator).
>   There is a corresponding page (admin interface is probably the sensible place for this) where an admin can review these requests and either apply the correction or ignore the request if it appears spurious/invalid.

chris48s:
Closes #139
Refs https://github.com/DemocracyClub/polling_deploy/pull/110

Sorry - quite a lot going on in this..

Stuff I **have** done:

* Add bug report form ('nuts and bolts' are done, some work to do on design - see next section)
* Re-enable `/admin`
  * This gives us a nicer way to manage API keys
  * Improvements to admin console for log model
  * Get rid of stuff that makes no sense
  * Provide interface for extracting user feedback via admin console
  * Handy admin console tools for working with bug reports
* Include a 'report a problem' link in API outputs so data consumers can give users a back-channel

Stuff I **haven't** done:

Basically, designed it.

What I've done for now is just shoved a link to the bug report form with the text "Is there anything wrong with this page?" (and not particularly attempted to lay it out nicely) but I really quickly realised that this isn't an ideal default because if you want to report something, you can't see the thing while you're filling the form in. Based on the reports from last year, I think it would be useful if by default this form was displayed inline (I've added the template/context processors to do an inline form, but they're not used anywhere yet).

I think the direction I'm probably going to go with this next is to try ripping off the gov.uk style, roughly like this:

gifrecord_2018-02-26_154915

I've got a decent idea how I want to do it, but since it involves a bunch of fiddling with templates and javascript and some layout decisions which are going to be influenced by the theme I think given #1137 is in progress I'll basically end up doing the same job twice if I try and do this now. At this point it is probably best block progress on this until #1137 is done. Once merged I'll rebase on to it and do the front-end bits.

Also you may want to review the text a bit?

Not to self: deploying this requires logger DB migrations.


chris48s:
Right - Accepting that we probably want to move this whole div to a different place on the page, I think this is good for review. The solution I've arrived at works like this with javascript (which is quite nice):

gifrecord_2018-03-23_140648

If you hit the page without javascript enabled, you only get the feedback form. I think the 'breakthough' (if you can call it that) was deciding that the 'report a problem' form is an enhancement, so we can show it if you have JS enabled and just provide no button if not. If someone doesn't have JS enabled there is still some kind of back-channel there. We might get a few bug reports as feedback, but that's going to account for a small slice of the pie. I think that's probably a better tradeoff than showing 2 forms in that situation or a confusing mashup - the page still renders quite nicely with no JS:

screenshot at 2018-03-23 14 09 34

Finally there is also the small issue with getting the buttons to space nicely on a narrow display. @symroe to provide advice :)

polling_stations/apps/bug_reports/templates/bug_reports/disclaimer.html
```diff
@@ -0,0 +1,5 @@
+<p><small>
+  If you provide an email address, we may contact you to
+  request more information or notify you of improvements.
+  It will not be used for email marketing purposes.
```
chris48s: I think we should have something to this effect here, but feedback on wording would be useful. Maybe @JoeMitchell has thoughts on this?

polling_stations/templates/fragments/feedback_js.html
```diff
@@ -15,10 +15,23 @@
     });
 
     if ($('#feedback_form input[name=found_useful]:checked').val() === undefined) {
+
       $("#feedback_form .comments_container").hide();
+      $('<input type="radio" name="found_useful" value="PROBLEM" required data-toggle="button" id="report_problem" />'+
+        '<label class="link-button" for="report_problem">Report a problem with this page</label>'
```
chris48s: Also open to feedback on the text here (ping @JoeMitchell ). I think 'Report a problem with this page' is good but we could say 'Report a problem'  or 'There is something wrong with this page' (maybe that fits better as an 'answer' to 'Did you find this useful?' The context is:

screenshot at 2018-03-23 14 18 42

polling_stations/apps/bug_reports/templates/bug_reports/report_form_view.html
```diff
@@ -0,0 +1,29 @@
+{% extends "base.html" %}
+
+{% load dc_forms %}
+{% load i18n %}
+{% load static from staticfiles %}
+
+{% block content %}
+
+<div class="row">
+  <div class="columns large-8">
```
symroe: `<div class="row">` and `<div class="columns large-8">` isn't used in base theme 0.3. both can just be removed.

polling_stations/apps/bug_reports/templates/bug_reports/report_form_inline.html
```diff
@@ -0,0 +1,17 @@
...
+
+<div class="card">
+  <form id="report_form" action="{% url 'bug_report_form_view' %}" method="post">
+      {% csrf_token %}
+      <div class="button">Report a Problem</div>
+      <div class="form-group">
+        {{ bug_report_form|dc_form }}
+        <p><small>
+          If you provide an email address, we may contact you to
+          request more information or notify you of improvements.
```
symroe: "improvements related to this report"? No general improvements, right?

polling_stations/apps/bug_reports/views.py
```diff
@@ -28,7 +28,10 @@ def post(self, request, *args, **kwargs):
 
             report.status = 'OPEN'
             report.report_type = 'OTHER'
-            report.user_agent = request.META['HTTP_USER_AGENT']
+            try:
+                report.user_agent = request.META['HTTP_USER_AGENT']
```
symroe: Can this be `request.META.get('HTTP_USER_AGENT', '')`?

polling_stations/apps/bug_reports/tests/test_bug_report.py
```diff
@@ -0,0 +1,67 @@
...
+        self.assertEqual(302, response.status_code)
+        reports = BugReport.objects.all()
+        self.assertEqual(1, len(reports))
+        self.assertEqual('some text in here', reports[0].description)
+        self.assertEqual('', reports[0].source_url)
+        self.assertEqual('', reports[0].source)
+        self.assertEqual('OPEN', reports[0].status)
+        self.assertEqual('OTHER', reports[0].report_type)
+
+    def test_post_bug_report_view_valid_with_querystring(self):
```
symroe: Would you mind adding a test that checks we don't redirect to a malicious URL in the source URL (I see you've dealt with this in the code, but we don't want a regression on that)

polling_stations/apps/bug_reports/templates/bug_reports/report_form_view.html
symroe: Sorry, realised this was replaced later in the PR 👍

polling_stations/apps/bug_reports/templates/bug_reports/disclaimer.html
JoeMitchell: i think it's good!

polling_stations/templates/fragments/feedback_js.html
JoeMitchell: Quite tricky this bit, might just need to be out in the world and see what happens.

My guess is that 'There is something wrong with this page' better suits a factual error, whereas 'Report a problem' better suits a technical bug... But we're looking for both, I imagine?

Maybe 'Report a problem' because it's shorter? :)

chris48s:
Nice. Cheers - useful suggestions. Think this should be good now

PR #1173, opened 2018-02-28
update ONSPD model and import script for Feb 2018 release

### LINKED ISSUE #1084: uk_geo_utils app: update ONSPD model and import script for Feb 2018 release
The Feb 2018 ONSPD release is going to remove and rename some legacy fields which will require the ONSPD model and import script to be updated. This may possibly impact on ONSUD as well.

From the Nov 2017 release notes:

> For the February 2018 release onwards, the ONSPD will no longer contain the following fields:
> * 1991  Census  ED  (both  OGSS  and  Census  code  ranges)  and indicator  (PSED,  CENED  and  EDIND)
> * Pre-July 2006 SHA (OSHAPREV)
> * 2009 LEA (LEA)
> * Pre-2002 Health Authority (OLDHA)
> * 1991 ward (both OGSS and Census code ranges) (WARDO91 and WARDC91)
> * 1998 ward (WARD98)
> * Pre-October 2006 PCO (OLDPCT)

> The health and 1998 ward fields will still be available on the NHSPD. In addition, two fields will be changing:
> * The 2012 Pan SHA field (HRO) will be updated to the NHS England (Regions) (NHSER)
> * The GOR field will be renamed RGN to reflect the 2010 Regions (former GORs)

> The HRO field will still be available on the NHSPD.

chris48s:
Closes #1084
Refs https://github.com/DemocracyClub/polling_deploy/issues/106


symroe:
As it is this looks good to go.

For future, do we want to consider supporting historic data as well as new data? It would make things like the `COPY` statement harder to write, but it might be useful to know we can import data from the past. (Thinking especially about `uk_geo_utils` being its own package, rather than for it's current use case where we obviously don't care about the past)

PR #1176, opened 2018-03-05
uk_geo_utils has moved to its own repo

### LINKED ISSUE #1090: Package uk_geo_utils as a re-usable component
I think `uk_geo_utils` is _fairly_ close to being a django app which can move to its own repo and packaged as a re-usable app which can be shared by Every Election and WhoCanIVoteFor in future. As you're working on #1087 and #1088 review whether there is additional work to do here first. If not:

* Move it to its own repo
* Write some docs
* Package it (so it is `pip install`-able)

etc

chris48s:
Closes #1090

Expect coverage % will drop as this PR deletes a lot of code with high test coverage.


chris48s:
I'm just going to crack on and merge this as it does not change any functionality. Its just moving stuff from one place to another.

PR #1186, opened 2018-03-07
Deal with stations represented as polygons, import Calderdale

chris48s:
Uniquely, Calderdale are now representing their polling stations as a polygon of the building rather than just a point:

screenshot at 2018-03-07 13 15 59

Quite cool, but no other council does this, so I'm not particularly keen to handle it beyond reducing it to a point on import. This PR allows us to explicitly handle this case. Note it also chucks out a bunch of warnings because in 99% of cases if the station isn't a point it is more likely to indicate that I'm trying to import the wrong file or something.


symroe:
So lovely they're doing it! A shame to throw the data away, but I agree it's not worth dealing with this single case. 👍

PR #1194, opened 2018-03-12
voter ID pilot info

### LINKED ISSUE #1059: Add info on voter ID trials for relevant areas in May 2018
Woking, Gosport, Bromley, Watford and Slough will be taking part in voter ID trials (see https://www.gov.uk/government/news/voter-id-pilot-to-launch-in-local-elections ), so the block 

```django
{% if territory == 'NI' %}
  <p>{% trans "You will need photographic identification" %}</p>
{% else %}
  <p>{% trans "You don't need any other form of ID" %}</p>
{% endif %}
```

will need to be updated to present the correct info for these areas

### LINKED ISSUE #1128: Surface additional notes/meta-data to API users
In our [user feedback](https://democracyclub.org.uk/wheredoivote_user_feedback/), lots of users left positive feedback that opening times and info about whether you needed id/polling card were useful.

We currently don't serve this data up in the API, but it might be useful to make this available to API data consumers. This would be particularly useful in a situation where there are additional notes (e.g: [voter id trials](https://github.com/DemocracyClub/UK-Polling-Stations/issues/1059) )

Currently this info mostly lives in templates. In order to do this we would need to move this info out of templates do we define it in python and we can use it on the front-end and in API outputs.

We would also add some extra keys to json outputs from the 'finder' API and corresponding documentation changes.

chris48s:
Closes #1059
Closes #1128

Definitely don't review/merge this. It is based on assumptions about PRs that haven't been merged and data structures that do not yet exist which may end up not being true. Also needs rebasing onto other stuff to properly finish.

TODO:
* Haven't done the API docs as I'll end up doing it twice to sort out the merge conflict with https://github.com/DemocracyClub/UK-Polling-Stations/pull/1145/commits/ddfa9b52dde25af787a767b168e6c28d7bd44e13 but the format will be roughly that, with an object.
* Write a test for `get_metadata()`
* Once we've got https://github.com/DemocracyClub/EveryElection/pull/223 merged/deployed and started serving up some data I'll finish this off.

In current design we will show something like

screenshot at 2018-03-12 17 34 48

but that's in the old theme and we should aim to merge #1137 ahead of this and rebase on to it.

### DIFF
--- a/polling_stations/apps/data_finder/helpers.py
+++ b/polling_stations/apps/data_finder/helpers.py
@@ -237,8 +237,7 @@ class EveryElectionWrapper:
                     continue
                 if not election['metadata']:
                     continue
-                if election['group_type'] == 'organisation':
-                    return election['metadata']
+                return election['metadata']
         return None
 
 
--- a/polling_stations/apps/data_finder/helpers.py
+++ b/polling_stations/apps/data_finder/helpers.py
@@ -228,6 +228,19 @@ class EveryElectionWrapper:
                     })
         return explanations
 
+    def get_metadata(self):
+        if not self.request_success:
+            return None
+        if len(self.elections) > 0:
+            for election in self.elections:
+                if not 'metadata' in election:
+                    continue
+                if not election['metadata']:
+                    continue
+                if election['group_type'] == 'organisation':
+                    return election['metadata']
+        return None
+
 
 class DirectionsHelper():
 
--- a/polling_stations/apps/api/address.py
+++ b/polling_stations/apps/api/address.py
@@ -48,6 +48,7 @@ class PostcodeResponseSerializer(serializers.Serializer):
     polling_station = PollingStationGeoSerializer(read_only=True)
     addresses = ResidentialAddressSerializer(read_only=True, many=True)
     report_problem_url = serializers.CharField(read_only=True)
+    metadata = serializers.DictField(read_only=True)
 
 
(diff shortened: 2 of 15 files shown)


symroe:
FYI, this is the same HTML in the new theme:

screenshot 2018-03-12 17 58 36

polling_stations/apps/data_finder/helpers.py
```diff
@@ -209,6 +209,19 @@ def get_explanations(self):
...
+    def get_metadata(self):
+        if not self.request_success:
+            return None
+        if len(self.elections) > 0:
+            for election in self.elections:
+                if not 'metadata' in election:
+                    continue
+                if not election['metadata']:
+                    continue
+                if election['group_type'] == 'organisation':
```
chris48s: on reflection, I think its important that each ballot id has the meta record attached to it in EE - revisit this.

chris48s:
This is a bit less WIP, but still WIP.

* https://github.com/DemocracyClub/EveryElection/pull/223 is now deployed
* I've rebased onto `master` to account for the changes made in #1196 and #1137.

Once #1145 is reviewed and merged that unblocks the work on the API docs and I'll finish this off.

Then once that's done we can crack on with #1207 :rocket:

chris48s:
I reckon this covers everything now

symroe:
Great stuff!

Issue #1086, opened 2017-11-24
"why don't we just use AddressBase?"


chris48s:
Discussion is too long for an issue, but there's several pages of notes in the wiki about the process of evaluating, benchmarking and prototyping this new data model: https://github.com/DemocracyClub/UK-Polling-Stations/wiki/AddressBase-as-a-canonical-base-dataset

One of the useful things to come out of this is a list of [questions](https://github.com/DemocracyClub/UK-Polling-Stations/wiki/Questions) we should bear in mind if we get a chance to conduct research with an EMS supplier or Electoral Services Officer.

PR #1196, opened 2018-03-15
Query EE by point around the edges

### LINKED ISSUE #1129: Save UPRNs on ResidentialAddress records (if we have it)
* Add a column for it, allow it to be empty
* Pass it through in import scripts if we've got it
* Review uprn as slug logic (at the moment we assume if we pass a uprn we want it to be the slug, but sometimes this is.. problematic e.g: parent/child UPRNs)

We don't want to use it for anything right now, but its a useful bit of meta-data to have. Y'know - just in case.

chris48s:
Key changes in this PR:
* If we can attach a UPRN and doorstep-level grid reference to a `ResidentialAddress` record at import time with some reasonable level of confidence that it is correct, do that. Denormalising point into `ResidentialAddress` table here saves us a round-trip to the DB or a join at query time.
* If we show the user an address picker and we hold a grid reference for the target address, query Every Election by grid reference instead of postcode (if possible). In all other cases, just query by postcode.
* Check Every Election before serving an API result. If the target postcode/address does not have an election, serve up `polling_station_known: false` (i.e: the "we don't know" response). For future we should probably look throwing an error response here but no result is better than "go to the wrong place" and isn't a BC break for existing consumers.
* Add and update unit tests.

Side notes:
* Obviously there are still some fuzzy edges here, but this definitely reduces them.
* This is a giant merge conlfict with #1194 but merge this one ahead of reviewing #1194 and I'll rework that to match the new behaviour defined here. #1194 is stil blocked on review of https://github.com/DemocracyClub/EveryElection/pull/223 and this is PR good to go now.
* I'm now fairly convinced that `AddressSet` is _the wrong abstraction_ . [At the time](https://github.com/DemocracyClub/UK-Polling-Stations/pull/359#issuecomment-247840082), it seemed like a `set()` of `NamedTuple`s was conceptually the correct representation of this data structure. I've cleaned up the boilerplate for doing it a bit in this PR, but there are now 3 methods on this class which basically take the set of tuples, convert it into a list of dicts perform some transformations and then turn the list back into a set of tuples again. I think there's a good case for either just keeping it mutable (i.e: represented as a list of dicts) for 'longer' in the process before turning it into a set or thinking about processing it in a more 'functional' way (pure functions which accept a set and return a different set). Dunno.. I'm not quite ready to shave this yak yet - needs more thought, but this is defo on my list of stuff that needs a refactor now. Any input on it?
* Doing this does add some additional performance overhead to the import process. Doesn't seem to be huge, but each script is definitely going to be a bit slower.

Closes #1129
(very loosely refs #1045, #234, #1086)


polling_stations/apps/pollingstations/models.py
```diff
@@ -122,6 +122,8 @@ class ResidentialAddress(models.Model):
     council            = models.ForeignKey(Council, null=True)
     polling_station_id = models.CharField(blank=True, max_length=100)
     slug               = models.SlugField(blank=False, null=False, db_index=True, unique=True, max_length=255)
+    uprn               = models.CharField(max_length=100, blank=True)
```
symroe: Looks like we don't need it at the moment, but it might be worth chucking an index on this

symroe:
AddressSet: no particular input at the moment – my personal style would be to use mutable objects in a list until the last moment – maybe using `__slots__` if memory were an issue. I have no good reasons for this above a functional option, this is just how I would think about writing it.

Seems like a post-election refractor though, so we can chat about it at a later date.

Otherwise: this is looking great – I left one minor comment but other than that I reckon it's good to go 👍

chris48s:
> Seems like a post-election refractor though

yep - absolutely

PR #1203, opened 2018-03-19
 log warning if >=1% of polling district area is outside target local auth

chris48s:
Refs #1045

This does several useful things:

1. Report how much of a polling district is inside the target local auth at import time. If its >99%, log an `INFO`, else a `WARNING`. If there is a substantial chunk of a polling district outside the council area this indicates a possible issue. The 1% tolerance gives us a (needed) bit of leeway for "these 2 lines are supposed to be exactly the same but they aren't quite".
2. Performing transformations/calculations on the polygons at this stage has the useful side effect that issues like polygons with implicit SRID or invalid polygons will give us a more explicit warning e.g:
    * `ERROR: Calling transform() with no SRID set is not supported`
    * `ERROR: Error encountered checking Geometry returned from GEOS C function "GEOSIntersection_r"`

### DIFF
--- a/polling_stations/apps/data_collection/base_importers.py
+++ b/polling_stations/apps/data_collection/base_importers.py
@@ -14,7 +14,7 @@ from django.apps import apps
 from django.core.management.base import BaseCommand
 from django.conf import settings
 from django.contrib.gis import geos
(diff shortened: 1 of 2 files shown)


symroe:
Nice!

PR #1208, opened 2018-03-20
remove NUS Wales subsite

chris48s:
Now we have dropped wheredoivote.wales, bleibleidleisio.cymru etc, lets bin this :wave: :dragon: 
If we want to revive Welsh at a later date we can revisit under our own branding


symroe:
👋 good times man, good times. 👋

PR #1209, opened 2018-03-20
Remove pointless code

chris48s:
* Remove a helper that is no longer used
* Coverage report tells us nothing useful - lets get rid of it
* I have no intention of fixing the league table between now and May, so lets get rid of the route so we at least don't have misleading content on the site


symroe:
…and the tables looked so nice…

This is good – one thing though: do we care about any links in the wild? I think we link to the league table from one of the yearly reports on the DC website so we should remove that. Is it worth putting a redirect in place?

chris48s:
So we did.. what should it redirect to? Home page?

symroe:
There's nowhere good to link to, but maybe the API, as at least _some_ people could get coverage if they really wanted to? Otherwise I don't really mind, as long as it's not a 404

PR #1217, opened 2018-03-22
add copyright notices

chris48s:
There are a bunch of notices and attributions we are supposed to have on our site where we use postcode lookups, boundaries, etc (see https://www.ons.gov.uk/methodology/geography/licences ). Lets bung them at the bottom of the footer.

### DIFF
--- a/polling_stations/templates/base.html
+++ b/polling_stations/templates/base.html
@@ -49,7 +49,13 @@
       <li><a href="https://facebook.com/DemocracyClub">Facebook</a></li>
       <li><a href="https://github.com/DemocracyClub">GitHub</a></li>
     </ul>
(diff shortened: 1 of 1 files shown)


symroe:
Do you think they have to be on every page, or can they be on an about page?

chris48s:
Ts & Cs say "whenever the owners' data is used". My interpretation is they do need to be on say `/postcode`, `/address` but not on every single page Shoving them at the bottom of the footer is a way to make sure they do appear everywhere they need to be.

The "about" link redirects to https://democracyclub.org.uk/projects/polling-stations/ - would you prefer to put them somewhere in there?

symroe:
This smells like a policy written by someone who doesn't know how APIs or data works, but fair enough, this is the quickest best option :+1:

chris48s:
Agreed. Simultaneously, let he whose data attribution policy is without sin cast the first stone:

screenshot at 2018-03-26 11 18 01

Perhaps this would be a useful consideration to bear in mind if we review our own attribution policy for candidate data.

PR #1229, opened 2018-03-28
re-order results page content

### LINKED ISSUE #1207: review information hierarchy
Now that #1137 is merged, we need to review the order we present content in, esp referencing https://democracyclub.org.uk/wheredoivote_user_feedback/

chris48s:
Closes #1207

* Move the opening hours on to the same card as the address/map/etc
* Move polling card/voter id info much higher up the page
* Move 'register to vote' card higher up the page (we will remove this card after the deadline on 17th April)
* Move 'information on your candidates' card lower down the page (very few people wanted this or said it was useful)

Although it is useful to look at the diff, it might also be easier to look at screenshots to review this. Rather than pasting lots of large screenshots into a GH issue, I've uploaded them to: https://drive.google.com/drive/folders/1SrJ_CZXJEORrpxwMrvrSxeiOEx28iZkY

There are screenshots for each of the following cases:

* Station not known, without voter ID pilot
* Station known, with map (without voter ID pilot)
* Station known, no map (without voter ID pilot)
* Station not known, voter ID pilot
* Station known, (with map) voter ID pilot
* No Election
* Multiple Councils

I have intentionally done the screenshots with a narrow resolution as this is the experience we're trying to optimise for.

### DIFF
--- a/polling_stations/templates/fragments/polling_station_known.html
+++ b/polling_stations/templates/fragments/polling_station_known.html
@@ -34,6 +34,8 @@
   </div>
 {% endif %}
 
+<p><strong>Polling stations are open from 7am to 10pm on Thursday 3 May.</strong></p>
+
 {% if station.location %}
   <div id="area_map" class="card_inset"></div>
 {% endif %}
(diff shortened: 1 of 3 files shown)


JoeMitchell:
Looks really good Chris. 👍 

My only thought would be about the top card — should the 'It's about a 0.6m walk...' and 'Get walking directions via Google' be below the map? 

But if it didn't come up in the feedback, it's perhaps not a problem.

PR #1256, opened 2018-04-01
replace links to geojson.org with links to RFC 7946 in api docs

chris48s:
RFC 7946 replaces the old 2008 spec


chris48s:
Just going to go rogue and merge this. Its pretty clear cut..

PR #1348, opened 2018-04-27
misc fixes

chris48s:
doing this one like this as an excuse to test deploying an ad-hoc fix


symroe:
Does this script get run on a new image build? Worth doing that, or fixing it in the importer too?

chris48s:
> Does this script get run on a new image build?

yep:

https://github.com/DemocracyClub/polling_deploy/blob/e7a6413cdf61f7700856779c41ff4d494cdff4a5/deploy.yml#L62-L66

I'll probs tidy this particular one to the importer once I've finished noodling though

symroe:
Nice, just checking :-)

PR #1391, opened 2018-06-15
Move integration tests to chromedriver

### LINKED ISSUE #1141: migrate the integration tests to...something else
Phantom JS isn't really actively developed anymore and Selenium are dropping support for it:

```
site-packages/selenium/webdriver/phantomjs/webdriver.py:49: UserWarning: Selenium support for PhantomJS has been deprecated, please use headless versions of Chrome or Firefox instead
```

it is also the flaky thing that makes the tests intermittently fail.

I've had the tests running on Firefox at one point, but I seem to remember there are a couple of useful things that headless Firefox doesn't support (or didn't at the time) that we would have had to remove. In javascript land the simple solution to this is electron, but I'm not sure that is an option for python/selenium. I'll poke headless Firefox/Chrome again at some point..

chris48s:
* Migrate E2E tests from PhantomJS to Chrome/Chromium and chromedriver. I had a look at Firefox and geckodriver as well, but I prefer this setup because geckodriver doesn't provide an interface to the javascript log and I'm keen to keep the tests which check the inline js didn't throw any errors. They have proved useful on more than one occasion. Closes #1141 . Hopefully this should be more stable than PhantomJS too. :crossed_fingers: 

* Update docs to reflect new testing setup (plus a couple of other bits I spotted while I was there)

* Ignore port number when generating fixture names


README.md
```diff
@@ -69,20 +67,29 @@ pip install -r requirements/base.txt
...
+sudo ln -s /usr/lib/chromium-browser/chromedriver /usr/local/bin/chromedriver
 ```
 
-On Ubunutu, run:
+On Mac OSX, run:
 ```
-sudo npm install -g phantomjs-prebuilt
+brew install brew-cask
+brew cask install google-chrome
+brew cask install chromedriver
```
chris48s: @symroe - I've attempted to fill in something sensible here, but obviously I haven't tested this. Can you let me know if this actually works/gets you to a stage where the tests will run. If this is wrong or we need extra steps, let me know and I'll correct.

symroe: First line needs to be 

```bash
brew tap homebrew/cask
```

`brew cask install google-chrome` isn't needed if the user already has Google Chrome installed using the conventional (non-brew) method of installing apps.

symroe:
This works for me, apart from `harvest` fails because I can't build the API docs. 

15 other tests worked though, so I expect this works fine.

chris48s:
> `brew cask install google-chrome` isn't needed if the user already has Google Chrome installed using the conventional (non-brew) method of installing apps.

Sure. Similarly, you wouldn't need to `brew install postgresql` if you've already installed Postgres via one of the other supported methods for MacOS: https://www.postgresql.org/download/macosx/ (I do get that Chrome is a more widely used application though).

If you are a mac user and you have already installed Chrome via another medium, what happens if you `brew cask install google-chrome`? Does it do nothing, or do you end up with 2 different incompatible versions installed or something?

If nothing happens, I reckon its fine to leave it as-is (if I already have python 3 set up I don't need to run `sudo apt-get install python3-dev` but its not going to do me any harm, right?).. if its an instruction that has a high chance of leaving someone in a wierd state, maybe we need a clarification there?


> because I can't build the API docs

Bit of a tangent, but did `brew install https://raw.githubusercontent.com/apiaryio/drafter/v3.2.7/tools/homebrew/drafter.rb` not get you a working install of drafter 3.2.7? I thought it did, but maybe I misremembered? If not I probably shouldn't claim it does in the docs. Also it would be useful to work out what the correct version of that instruction is. I did start looking at whether I can add 4.x support but currently the python bindings for Drafter support drafter 3 only - see https://github.com/kylef/draughtsman/issues/1, so using the 4.x pre-release isn't an option yet.

symroe:
> what happens if you brew cask install google-chrome?

You get an error saying that Chrome is already installed – no inconsistent state, but it does download the lot before telling me that. The reason I think this is different is that it's not obvious to me that `chromedriver` will work with _any_ (compatible) version of Chrome (i.e, I thought that I was installing a `-dev` version, not just the normal version)

> Bit of a tangent, but did brew install https://raw.githubusercontent.com/apiaryio/drafter/v3.2.7/tools/homebrew/drafter.rb not get you a working install of drafter 3.2.7? 

Nope! I'll shave that yak one day, but I wasn't about to sharpen the razor when reviewing this PR… :)

chris48s:
Hrm. In an attempt to debug this I shoved [linuxbrew](http://linuxbrew.sh/) on a VM (who knew, right?).

I couldn't get any version to install with that, including `4.0.0-pre.0`.. so that wasn't terribly helpful. However, `brew install --HEAD "https://raw.githubusercontent.com/apiaryio/drafter/v3.2.7/tools/homebrew/drafter.rb"` appeared to at least fail in the same place as `brew install --HEAD https://raw.github.com/apiaryio/drafter/master/tools/homebrew/drafter.rb` did.. maybe give that a go?

chris48s:
Given I've made all the requested changes which were actually related to the original point of this PR (moving the integration tests to headless chrome), I've put the line about installing drafter back how it was before I started and squashed the history into something sensible.

I'm going to merge this and I'll raise another issue about fixing the drafter install instructions for mac as it is only tangentially related.

PR #1419, opened 2018-09-06
add status check endpoint

chris48s:
This adds a status check endpoint for use with ELB status checking. It checks for a "clean" file (if applicable) and makes sure it can connect to any defined DB connections before returning a `200 OK`. I miiiight look to add some more handy checks to this at some point, but lets ignore that for now.

I've tried deploying this off a branch and it does the job. I'm quite keen to get this deployment work finished off this sprint if possible, so would be handy to get a quick :+1:  on this if you get a chance @symroe

### DIFF
--- a/polling_stations/apps/pollingstations/views.py
+++ b/polling_stations/apps/pollingstations/views.py
@@ -1,3 +1,22 @@
-from django.shortcuts import render
+import os
+from django.conf import settings
+from django.db import connections
+from django.db.utils import OperationalError
+from django.http import HttpResponse
 
-# Create your views here.
+
+def status_check(request):
+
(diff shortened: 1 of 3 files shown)


symroe:
Looks good, only question is about the logic inversion from Who. There I have a dirty file that I remove, here you're adding a clean file. Any particular reason?

chris48s:
> Any particular reason?

No, not really. I started doing it one way and then realised I had done it the opposite way round to WhoCIVF after I already had stuff done/built that was slow to change. I don't think either one is necessarily better/worse. If it is a pattern we end up using a lot, it is probably worth picking one approach and using it consistently.

PR #1432, opened 2018-10-23
remove some deprecated calls

chris48s:
This fixes the following deprecation warnings:

* `RemovedInDjango20Warning: Importing from django.core.urlresolvers is deprecated in favor of django.urls.`
* `RemovedInDjango20Warning: Specifying a namespace in django.conf.urls.include() without providing an app_name is deprecated. Set the app_name attribute in the included module, or pass a 2-tuple containing the list of patterns and app_name instead.`

### DIFF
--- a/polling_stations/apps/data_collection/urls.py
+++ b/polling_stations/apps/data_collection/urls.py
@@ -1,6 +1,6 @@
 from django.conf.urls import include, url
 from django.conf import settings
-from django.core.urlresolvers import reverse_lazy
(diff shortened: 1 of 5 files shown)


chris48s:
Got a bit enthusiastic trying to remove some stuff that's deprecated in DRF 1.9 before we've actually applied that upgrade. I've made some notes in https://github.com/DemocracyClub/UK-Polling-Stations/pull/1428#issuecomment-432198061

Thanks, Continuous Integration :)

Issue #1373, opened 2018-05-02
API should serve election info


chris48s:
* User has an election - yes/no
* Countermanded elections

refs https://github.com/DemocracyClub/WhereDoIVote-Widget/issues/37
refs https://github.com/DemocracyClub/EveryElection/issues/33

chris48s:
this is done in #1440 but I need to update the docs

chris48s:
I reckon #1446 mostly covers this, but it probably also need some comms for existing API consumers

chris48s:
merging this into #1459

PR #1433, opened 2018-10-25
Provide information about cancelled elections

chris48s:
* This turned out to be more complicated than planned
* 1a90087 is just re-organising. There are no functional changes in that commit.
* Because I don't really know exactly how we're going to handle this until we have to deal with real examples, I've intentionally constructed this in quite a flexible way so we can provide extra info if we have it but fall back to something generic if we don't. This is convenient in some ways but increases the amount of faffing about we have to deal with.
* For the moment the API behaviour for a cancelled election is the same as when there is just no election at all: we just don't return a polling station result. I decided not to add any special handling in the API (yet) for 2 reasons:
  * There is quite a lot going on in this PR already
  * Dealing with cancelled elections fits into a wider bucket of work around serving election info via WDIV API https://github.com/DemocracyClub/UK-Polling-Stations/issues/1373 which requires a bit more thinking. I should probably work on that in the near future though.


chris48s:
While working on https://github.com/DemocracyClub/WhoCanIVoteFor/pull/305 I thought ab out this slightly more and made a small wording change before merging :see_no_evil:

PR #1455, opened 2019-01-08
deal with upcoming changes in local government orgs

chris48s:
There are 4 data sources we care about that give us data on local authorities:

* ONSUD (UPRN --> GSS lookup)
* ONSPD (fallback postcode --> GSS lookup)
* Local Auth boundaries (polygons)
* Your Vote Matters (contact details for Electoral Services teams)

Each one has its own release schedule and at various points between now and May somewhere between 0 and 4 of these data sources will be updated with new codes/names/boundaries. This PR makes some changes which will smooth this transition a bit so that we hopefully don't break or need to make a code change every time one of those things changes.. hopefully.

I think we might want to hold off on merging/deploying this until we're a bit closer to the time, but it would be useful to get it reviewed soon so its sanity-checked and ready-to-go once we can fill in some more contact details.

---

e3f8695 is a very small commit that requires quite a lot of explanation:

The way we've got this set up now, the `LoggedPostcode` table lives on the logger DB in production and has a FK to `Councils`. This means that:

a) We have to maintain a `Councils` table on both the local DBs and one on the logger DB so the FK relationship works.
b) They need to be in sync with each other, or we start throwing `IntegrityError`s about the place trying to log stuff.
c) If some councils were to cease to exist and we want to delete stuff out of the councils table on logger (e.g: by running `import_councils` with the `--teardown` option, or by manually deleting old councils out of it), that will (unexpectedly) cascase the delete and delete corresponding records from the `LoggedPostcode` table.

I did think about a few different options here, including:

* Maintain a 'mongrel' `Councils` table on logger with both new and old councils in it plus any temp codes we've used.
* Take a backup/archive of our historic logs, clear `LoggedPostcode` and rebuild the councils table.

..but having thought it through, I reckon it makes most sense to just drop the FK constraint on `LoggedPostcode.council_id` and move to a position where we store a GSS code (or temp code) but we don't care if that GSS code currently exists in the(/a) `Councils` table or not. That means we can just stop storing the `Councils` table on logger at all or caring if it is in sync with the other DBs and then if we find out that `BPC-tempid` is now `E06000057` and decide to change it a week before polling day, then its easy to deal with and we won't be surprised by data loss. We don't ever really use the property that `LoggedPostcode.council_id` is a FK (we never query it within the application - we only ever dump the logs and analyse them offline), so it makes more sense that it just contains whatever codes were relevant at the time we created the logged record.

I think that makes sense. Does that sound good? I'll probably give it a test run on a real DB dump before I apply it on the real logger DB.


polling_stations/settings/constants/councils.py
```diff
@@ -2,5 +2,37 @@
...
+    "E07000053": "DST-tempid",
+    # East Suffolk
+    "E07000205": "ESK-tempid",
+    "E07000206": "ESK-tempid",
+    # West Suffolk
+    "E07000201": "WSK-tempid",
+    "E07000204": "WSK-tempid",
+}
+
+NEW_COUNCILS = ["SWT-tempid", "BPC-tempid", "DST-tempid", "ESK-tempid", "WSK-tempid"]
```
chris48s: I've made up some fake codes here based on the GDS register codes, but I'm hoping we'll at least get a Code History DB release telling us what the new codes are going to be in the next couple of months :crossed-fingers: These are the same temp codes I've assigned these `OrganisationGeography` objects in Every Election.

polling_stations/apps/councils/management/commands/import_councils.py
```diff
@@ -1,13 +1,66 @@
...
 from councils.models import Council
 
 
+def union_areas(a1, a2):
+    if not a1:
+        return a2
+    return MultiPolygon(a1.union(a2))
+
+
+TEMP_CONTACT_DETAILS = {
```
chris48s: Hopefully we'll be able to fill some more of this stuff in fairly soon. I'll follow up with the Electoral Commission again when we meet them next month.

chris48s:
I split a chunk of this out into #1468 merged it and rebased the other half to resolve conflicts

chris48s:
Wrote a couple of importers for new areas and realised I'd missed a few things. I think we're now at a point where it makes sense to merge this on Monday so we can crack on with that. @symroe - if you're about on Monday, I might run a couple of assumptions past you on before going ahead with that..

chris48s:
Going to merge this as it stands as we've now got a couple of areas from this list ready to go.

If it turns out that we need to deal with either Dorset or East Suffolk as >1 organisation I'll cross that bridge when we come to it.

PR #1458, opened 2019-01-11
present a single date-level elections metadata object

### LINKED ISSUE #1456: Improve election metadata handling
https://github.com/DemocracyClub/UK-Polling-Stations/blob/697437a02b87959116e45a3e41accd34a56184a4/polling_stations/apps/data_finder/helpers/every_election.py#L173-L176

https://github.com/DemocracyClub/UK-Polling-Stations/blob/ea0fc102341f50d72765ff8043f429e2021f504a/polling_stations/apps/api/address.py#L126

https://github.com/DemocracyClub/UK-Polling-Stations/blob/ea0fc102341f50d72765ff8043f429e2021f504a/polling_stations/apps/api/postcode.py#L118

chris48s:
Having thought through https://github.com/DemocracyClub/UK-Polling-Stations/issues/1456 I think the right decisions to take here are:

Data model:
* Special-case voter id pilots and accept that we also need to special-case any other category of spurious election event we add
* In Every Election we must store this stuff at ballot-level not group (by convention). We seem to be standardising on the ballot object as the canonical "unit of election" we expose via our various APIs

WDIV Front-end:
* Assume we won't have >1 pilots with different requirements on the same day at the same polling station :pray:  - if there's one ballot happening with a pilot, just pick one and show that info

WDIV API:
* Continue to present a single top-level `metadata` object aggregating info from all applicable ballot objects.
* Instead of trying mash _everything_ into a consistent meta-data object (which is hard if there are >1 ballots), be slightly "editorial" about it. If there is a cancelled election, that trumps everything else. If there are >=1 ballots with an id trial, give the user one info object (again assuming we won't have >1 pilots with different requirements at the same polling station at the same time) instead of expecting them to disambiguate >1 objects.

This changes how we assemble the `metadata` object, but avoids making a backwards-incompatible API change on WDIV. Its easier for us to explain to API consumers and its easier for them to consume than having to extract this from the ballots array in their own app.

Happy to discuss this if necessary @symroe as it took me about a day to absorb all the context and get my head back into the problem-space in the end.

closes #1456

### DIFF
--- a/polling_stations/apps/data_finder/tests/test_ee_wrapper.py
+++ b/polling_stations/apps/data_finder/tests/test_ee_wrapper.py
@@ -58,6 +58,7 @@ def get_data_with_elections(self, query_url):
             "election_id": "foo.bar.date",
             "election_title": "some election",
             "group_type": "organisation",
+            "metadata": None,
             "cancelled": False,
             "poll_open_date": datetime.now().strftime("%Y-%m-%d"),
         },  # no explanation or metadata keys
@@ -76,7 +77,7 @@ def get_data_with_elections(self, query_url):
             "group_type": None,
             "explanation": "some text",  # explanation key contains text
             "metadata": {
-                "this election": "has some metadata"
+                "2019-05-02-id-pilot": {"this election": "has an ID pilot"}
             },  # metadata key is an object
             "cancelled": False,
             "poll_open_date": datetime.now().strftime("%Y-%m-%d"),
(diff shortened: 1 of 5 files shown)


chris48s:
Since Friday I've slightly modified this so that we'll pull info for both voter id pilots in England and standard Northern Ireland ID requirement info into a single `voter_id` key. This adds that additional assumption/requirement that we attach that info to all the NI ballots in Every Election.

polling_stations/apps/data_finder/tests/test_ee_wrapper.py
```diff
@@ -76,7 +77,7 @@ def get_data_with_elections(self, query_url):
             "group_type": None,
             "explanation": "some text",  # explanation key contains text
             "metadata": {
-                "this election": "has some metadata"
+                "2019-05-02-id-pilot": {"this election": "has an id pilot"}
```
symroe: Caps "ID"

chris48s: all about the grammatically correct test fixtures :D

PR #1469, opened 2019-01-30
Don't import address record if postcode fails picture check

### LINKED ISSUE #1321: Picture-check postcodes on ResidentialAddress import
If a 'postcode' in the input file doesn't look like a valid postcode, flag it.

The specific example I've just spotted is a group of properties where the 'postcode' column variously contains `UB4 0SE` (with a zero) and `UB4 OSE` (with an uppercase O). `UB4 OSE` should fail a basic validation check (it ends with 3 letters) and be flagged at import.

chris48s:
Closes #1321

This performs a regex check on every postcode to ensure its valid before we import it. If it isn't, discard it (it won't be reachable from the front-end anyway)  and log it (if we know its happened, we might be able to correct it).

### DIFF
--- a/polling_stations/apps/data_collection/data_types.py
+++ b/polling_stations/apps/data_collection/data_types.py
@@ -5,7 +5,11 @@ Data type classes used by base importers
 import abc
 import logging
 from collections import namedtuple
+
 from django.db import connection
+from django.forms import ValidationError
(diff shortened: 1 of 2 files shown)


chris48s:
this is now done in #1476

PR #1470, opened 2019-01-30
store has_election in LoggedPostcode table

### LINKED ISSUE #1377: Store has_election in LoggedPostcode table

chris48s:
closes #1377

This makes it easier to produce stats on searches in areas with an election/without an election happening


chris48s:
note to self: logger migration in this one

PR #1475, opened 2019-01-31
more lenient station UPRN validation in Xpress importer, flag on Xpress base importer to control fallback to postcode

### LINKED ISSUE #1463: Add a flag on Xpress base importer to control fallback to postcode
Sometimes if we've got UPRNs for stations, we want to only attach a station point if we can do so by UPRN - if we can't, we want to show no map instead of falling back to geocoding by postcode. At the moment, to do this we have to override `get_station_point()` e.g: https://github.com/DemocracyClub/UK-Polling-Stations/blob/bdb9dc6237c1cc9875b1f08174099319615d6bfe/polling_stations/apps/data_collection/management/commands/import_kirklees.py#L37

Switch this so we can control postcode fallback Y/N by setting a boolean.

### LINKED ISSUE #1465: Loosen up station UPRN validation in Xpress importer
At the moment, if we've got a station UPRN in Xpress data, we filter the UPRNs on postcode before attempting to match (essentially requiring the supplied station postcode and postcode in addressbase to match in order to geocode):

https://github.com/DemocracyClub/UK-Polling-Stations/blob/bdb9dc6237c1cc9875b1f08174099319615d6bfe/polling_stations/apps/data_collection/ems_importers.py#L115-L116

In some cases, it may be more useful to assume the UPRN match is more important and accept mismatched postcodes if the UPRNs match and then address text "agrees". Either make this behaviour configurable, or convert this to match on UPRN first but log a warning if the postcodes don't agree so we flag it for manual check.

chris48s:
closes #1463
closes #1465

These 2 issues aren't strictly related, but they both involved changing the same function so I just loaded them into one PR

### DIFF
--- a/polling_stations/apps/data_collection/ems_importers.py
+++ b/polling_stations/apps/data_collection/ems_importers.py
@@ -82,6 +82,12 @@ class BaseXpressCsvImporter(BaseCsvStationsCsvAddressesImporter, metaclass=abc.A
         except PostcodeError:
             return None
 
+    def geocode_from_uprn(self, record):
+        uprn = getattr(record, self.station_uprn_field)
+        uprn = uprn.lstrip("0")
+        g = AddressBaseGeocoder(self.get_station_postcode(record))
+        return g.get_point(getattr(record, self.station_uprn_field))
+
     def get_station_point(self, record):
         location = None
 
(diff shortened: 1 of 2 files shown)


polling_stations/apps/data_collection/ems_importers.py
```diff
@@ -31,6 +32,10 @@
 class BaseXpressCsvImporter(BaseCsvStationsCsvAddressesImporter, metaclass=abc.ABCMeta):
     csv_delimiter = ","
 
+    # Set this to false in an import script if we want to only set a station
+    # point based on UPRN or co-ordinates (even if we've got a valid postcode)
+    allow_station_point_from_postcode = True
```
chris48s: I struggled to find a good name for this property as the behaviour it toggles is quite specific - suggestions for better name welcome if you've got any

PR #1520, opened 2019-02-17
be smarter if residential UPRN and postcode don't match

### LINKED ISSUE #1330: Make better assumptions if residential UPRN and postcode don't match
At the moment if we catch this condition, we essentially always assume `(address, postcode)` is good and discard the UPRN:

https://github.com/DemocracyClub/UK-Polling-Stations/blob/d2679c22a04233e34fb6f02e96f98d0e49f3cb50/polling_stations/apps/data_collection/data_types.py#L235-L247

but sometimes `(uprn, address)` is good and the postcode needs correcting.

Find a more sophisticated way to deal with this.

chris48s:
This is mostly explained inline in the code in some fairly verbose comments, but broadly I'm attempting to:

a) Make better assumptions in the case where UPRN and postcode don't match
b) Warn loudly about a subset of cases where we definitely want to do a manual check/follow-up

Closes #1330

### DIFF
--- a/polling_stations/apps/data_collection/data_types.py
+++ b/polling_stations/apps/data_collection/data_types.py
@@ -8,6 +8,7 @@ from collections import namedtuple
 
 from django.db import connection
 from django.forms import ValidationError
+from fuzzywuzzy import fuzz
 from localflavor.gb.forms import GBPostcodeField
 
 from data_collection.slugger import Slugger
@@ -163,7 +164,7 @@ class AddressList:
             )
 
     def get_address_lookup(self):
-        # for each address, build a lookup of address -> list of station ids
+        # for each address, build a lookup of address -> set of station ids
         address_lookup = {}
         for record in self.elements:
             address_slug = Slugger.slugify(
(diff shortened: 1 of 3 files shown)


polling_stations/apps/data_collection/data_types.py
```diff
@@ -191,6 +192,17 @@ def get_uprn_lookup(self):
 
         return uprn_lookup
 
+    def get_postcode_lookup(self):
```
symroe: Not tested this but would a `collections.Counter` object here remove the need for this method and the `is_split_postcode` function below?

```
c = Counter([r["postcode"] for r in self.elements)
c[record["postcode"]] > 1
```

Not really a functional PR message, but might reduce the number of lines and connotative load.

chris48s: I think the snippet you've given there would just count the number of times each postcode occurrs (which you would expect to be lots for every postcode, as it will occur once for each relevant address in the list) and doesn't take account of how many different polling stations its associated with (which is the thing we care about). That said, you're probably right that there is probably a niftier way of leveraging collections to do this job (and what I'm doing in `get_address_lookup()` / `get_uprn_lookup()` ). I'll have a think about it..

PR #1526, opened 2019-02-18
Revert "be smarter if residential UPRN and postcode don't match"

chris48s:
This reverts commit 1fbc213cba7f573933085376e3c770994f72f047.

Having spent a bit of time with 2 of us working with this, I've decided to back it out for now. I think there is merit in the approach, but in this exact form, although we're fixing some issues, we're also _introducing_ new errors and doing it in a way that is hard to undo. Needs more thought/testing

### DIFF
--- a/polling_stations/apps/data_collection/data_types.py
+++ b/polling_stations/apps/data_collection/data_types.py
@@ -8,7 +8,6 @@ from collections import namedtuple
 
 from django.db import connection
 from django.forms import ValidationError
-from fuzzywuzzy import fuzz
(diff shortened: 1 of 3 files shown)


symroe:
Given I reviewed that and didn't have a clue, not sure there's much point in me reviewing this. Carry on 👍

chris48s:
Cheers. I'll probably end up re-submitting a slightly tweaked version of this in the near future. I mainly just want to revert this on `master` for now so I can deploy without this change in the meantime

PR #1527, opened 2019-02-19
Import script 2019 02 19

GeoWill:


polling_stations/apps/data_collection/management/commands/import_south_norfolk.py
```diff
@@ -0,0 +1,27 @@
...
+        if record.property_urn.strip().lstrip("0") == "2630166893":
+            rec["postcode"] = "NR9 3JP"
+
+        if record.property_urn.strip().lstrip("0") == "2630159192":
+            rec["postcode"] = "NR35 2QR"
+
+        if record.property_urn.strip().lstrip("0") == "2630153843":
+            rec["postcode"] = "NR14 6RJ"
+
+        if record.addressline6 == "NR95DX":
```
chris48s: If we're comparing to the input file, there is no pre-processing done on the input file yet, so you'd want `if record.addressline6 == "NR9 5DX":` (with a space). This won't match anything because it has a space in the input file. (I can see this is just a placeholder for now so it doesn't really matter if we match it)

PR #1558, opened 2019-02-24
be smarter if residential UPRN and postcode don't match (part 2)

### LINKED ISSUE #1330: Make better assumptions if residential UPRN and postcode don't match

chris48s:
Closes #1330

Lets see if we can make a slightly better job of this. I _think_ this is an improvement.

**Key changes:**

* Strip commas before fuzzy matching
* Set default match threshold to 100
* Allow match threshold to be configured on each import script instead of set globally
* Allow fuzzy matching to be turned off if needed
* If we can't fuzzy match, delete any (problematic) leftovers

**Review notes:**

* Still doing this as a post-process:
  * Input to `address_record_to_dict()` should match CSV
  * There are some things we need to know which require iterating the whole thing
  * How do we deal with a single-record correction or exclusion?
* `INFO`s vs `WARNING`s
* ~Is `delete_unmatched_postcodes` setting useful/necessary? Would it be better as a `--flag`~

**TODO:**

* ~Test the changes against all existing import scripts~

@symroe - feel free to have opinions about this, but I think the next thing I want to do with this is review the assumptions with @GeoWill next week and we'll see how it goes/tweak it before we merge.

### DIFF
--- a/polling_stations/apps/addressbase/models.py
+++ b/polling_stations/apps/addressbase/models.py
@@ -1,4 +1,5 @@
 from django.contrib.gis.db import models
+from django.db import connection
 from uk_geo_utils.models import AbstractAddress, AbstractAddressManager, AbstractOnsud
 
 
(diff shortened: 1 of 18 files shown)


symroe:
Yep, I have no opinions about this, best to have a crack at it and see how it goes 👍

chris48s:
Will did some testing on this the other day and I've spent a couple of days updating the import scripts we've written so far and ironing out some of the kinks. Think this is going to be an improvement :crossed_fingers:

PR #1564, opened 2019-02-27
Imports 2019 02 27

GeoWill:


polling_stations/apps/data_collection/management/commands/import_knowsley.py
```diff
@@ -3,7 +3,40 @@
...
 class Command(BaseXpressDemocracyClubCsvImporter):
     council_id = "E08000011"
-    addresses_name = "local.2018-05-03/Version 1/Democracy_Club__03May2018 Knowsley.CSV"
-    stations_name = "local.2018-05-03/Version 1/Democracy_Club__03May2018 Knowsley.CSV"
-    elections = ["local.2018-05-03"]
+    addresses_name = "local.2019-05-02/Version 1/Democracy_Club__02May2019Knowsley.CSV"
+    stations_name = "local.2019-05-02/Version 1/Democracy_Club__02May2019Knowsley.CSV"
+    elections = ["local.2019-05-02"]
     csv_delimiter = ","
+    private = True
```
chris48s: This needs removing as we want to get the data from S3 by default.
I wonder if there is something cunning we can do like move this to a global setting you can set in `local.py` or add a lint rule to fail the build if `private = True` is set in an import script. I'll have a quick think tomorrow

chris48s:
I've not checked every single example, but this seems broadly sensible

There was an extra commit in here which added a bit of extra logging. One of the things I was going to ask you about was if you've got any ideas for how to make the logging a bit more useful. I like where you're going with it.. Feel free to push that commit up as another PR or something if you want

GeoWill:
Pushed as a fixup, but will squash when you give me the nod :+1:

polling_stations/apps/data_collection/management/commands/import_knowsley.py
GeoWill: I'll bung something in a pre-push hook to keep me right for now - was a classic "I'm already late for where I'm meant to be PR..."

chris48s:
Yep - squash away..

PR #1587, opened 2019-03-01
api doc updates for May 2019

### LINKED ISSUE #1459: Last few API doc updates
Now that #1458 is done, this is fairly nailed down. Provide examples of responses and a section a bit like c895a4a with some wordy stuff around edge cases relevant to May 2019:

* Voter ID pilots in England (`metadata` key)
* Northern Ireland: standard voter id requirements (`metadata` key)
* Northern Ireland: EONI Finder (`custom_finder` key)
* Cancelled ballots (`metadata` key)
* Ballots array (and inferring "does the user have an election" - Y/N)

Maybe take some cues from https://developers.democracyclub.org.uk/api/v1/#notes-and-tips-2

chris48s:
Closes #1459

@symroe - I wasn't really sure where to put this. I'm making quite a lot of reference to the `/address` and `/postcode` API endpoints and special cases in those outputs, so it seems like from the perspective of a new user it should come _after_ that section of the documentation. i.e: if you're reading the docs start-to-finish, we should explain the `/address` and `/postcode` endpoints first and then get into the special cases of those outputs. That said, its also something that I want people to read if they're using the API this May.

I think the solution is just to directly link to this section heading when I contact existing API users (this is the last task blocking that mailing) or create any new API keys.

### DIFF
--- a/polling_stations/templates/api_docs/blueprints/finder.apibp
+++ b/polling_stations/templates/api_docs/blueprints/finder.apibp
@@ -19,14 +19,16 @@ A `200 OK` response from `/postcode` or `/address` is an object containing the f
 * `metadata`: (object, nullable) - This field may be used to supply additional information about
   an election, particularly if there is some unusual condition or event for voters to be aware of.
 * `ballots`: (array) - An array of ballot objects listing the elections applicable to this request.
-  It is possible for more than one ballot to occur on the same date. For example, a user may vote in a local council election and mayoral election on the same day.
+  It is possible for more than one ballot to occur on the same date.
+  For example, a user may vote in a local council election and mayoral election on the same day.
+  If `ballots` is an empty array `[]`, there are no upcoming elections in this area.
 
 The entry point to a polling station search is a call to the `/postcode` endpoint.
 A valid postcode search may result in one of 3 outcomes:
-* We hold data for this area and all voters with this postcode vote at the same polling station.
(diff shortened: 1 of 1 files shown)


chris48s:
Once we're happy with the text on this, I'll also transplant _roughly_ the same thing into the docs on https://developers.democracyclub.org.uk but lets review it in one place first..

polling_stations/templates/api_docs/blueprints/finder.apibp
```diff
@@ -203,3 +205,76 @@ the following conditions can be observed in the response body:
...
+
+## Group Local Elections May 2019
+
+On 2 May 2019, local elections take place in England and Northern Ireland.
+
+### Ballots
+
+Local council elections will be taking place in 250 local authorities in England. Some areas also have mayoral elections.
+
+The `ballots` array can be used to determine if a user has elections scheduled. If `ballots` is an empty array, there is no scheduled election. Where elections are scheduled, `ballots` may contain multiple objects. For example, a user may vote in a local council election and mayoral election on the same day.
```
symroe: Worth adding something about no elections that we cover (we can't assert no elections). 

Also, do you want to get in to the business of recommending that users don't show data about polling location if we don't know about elections? These docs don't have a lot of opinions about _how_ to show the data, but we do have opinions about it.  I don't know how much we should talk about that here.

chris48s:
pushed a couple of caveats about parish/town/community elections :+1:

PR #1589, opened 2019-03-02
remove 'private' setting

chris48s:
If `PRIVATE_DATA_PATH` is configured locally, use that. Otherwise fetch from S3

@GeoWill - does this seem fair enough to you?


GeoWill:
@chris48s This looks good to me :slightly_smiling_face:

PR #1667, opened 2019-03-19
change "we don't know of any upcoming elections" wording

### LINKED ISSUE #1541: change "we don't know of any upcoming elections" wording
Can we tweak this a bit to clarify that there might be parish/town/community election in some places

chris48s:
closes #1541
This gives us some wording to bikeshed. I've attempted to balance the needs of 3 groups here:

* Providing useful information to members of the public
* Councils who do have parish elections happening in areas that don't have district ward elections and want us not to tell their users "you don't have an election"
* Councils who do have parish elections and don't want loads of people phoning them up to ask if they have an election or not.

Screenshot at 2019-03-19 15-00-29

For now, lets concern ourselves with the 'happy path'. There is also an edge case where there are several councils where we don't hold very good contact info, but I'm hoping to fix that at the start of April when the new councils become "official".

There is also an edge case for users in NI where the relevant thing to contact is EONI not the local council, but I'm going to ignore that one for now because this content is now 2nd May-specific and everyone in NI has an election on 2nd May, but we probably also need a NI-version of this content when we re-generic-ify this content again.

### DIFF
--- a/polling_stations/templates/fragments/no_election.html
+++ b/polling_stations/templates/fragments/no_election.html
@@ -5,17 +5,18 @@
   {% include "fragments/cancelled_election.html" %}
 {% else %}
   <h2>We don't know of any upcoming elections in your area</h2>
-  {% if council.phone %}
-    {% blocktrans with council.name as council_name and council.phone as council_phone %}
+  {% if council.phone and council.website %}
+    {% blocktrans with council.name as council_name and council.phone as council_phone and council.website as council_website %}
     <p>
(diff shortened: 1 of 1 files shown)


JoeMitchell:
Ello, did a quick http://hemingwayapp.com/. 

Could we break the first sentence into two: 

>There are no district council or mayoral elections in your area. There may be parish, town or community council elections in some areas.

This has a reading grade of 7, vs the original of 14 'very hard to read'.

PR #1700, opened 2019-03-22
Imports 2019 03 22

GeoWill:


warning.txt
```diff
@@ -0,0 +1,452 @@
+WARNING: Polling station 7562 is in Stratford on Avon District Council (E07000221) but target council is Warwick District Council (E07000222) - manual check recommended
```
chris48s: Can you remove this file from the PR :)

polling_stations/apps/data_collection/management/commands/import_warwick.py
```diff
@@ -3,7 +3,115 @@
...
+            record = record._replace(polling_place_postcode="CV35 8AN")
+        # Lillington Nursery, Lillington Nursery & Primary School, Grange Road Entrance, Off Pound Lane, Lillington, Royal Leamington Spa
+        if record.polling_place_id == "7395":
+            record = record._replace(polling_place_postcode="CV32 7AG")
+        # Lowsonford Village Hall, Lowsonford
+        if record.polling_place_id == "7280":
+            record = record._replace(polling_place_postcode="CV35 8JE")
+        # Church of Jesus Christ Of Latter Day Saints, Saltisford/Birmingham Road/A425, Warwick
+        if record.polling_place_id in ["7526", "7476", "7479"]:
+            record = record._replace(polling_place_postcode="CV34 4TT")
```
chris48s: If these are corrections, can we confirm them with the council first (stick them in a trello card for Peter to follow up). Usually our suggested corrections are right, but occasionally they'll come back and request we use a slightly different postcode to the one we suggest.

warning.txt
GeoWill: Done.

polling_stations/apps/data_collection/management/commands/import_warwick.py
GeoWill: That's now here: https://trello.com/c/EKhqevnh

PR #1741, opened 2019-04-02
WiP: Import script for Brighton and Hove (closes #1688)

### LINKED ISSUE #1688: Import E06000043-Brighton and Hove
Xpress DC

GeoWill:
https://trello.com/c/BEuzXuuy


chris48s:
Reply from Brighton & Hove:

> Thank you for enquiry. We have double checked this and I believe that it may be down to Bishop Hannington having two church halls…!

> As it is, the co-ordinates that you kindly provided relate to the church itself and one of the halls. However, the Bishop Hannington Youth & Community Centre (where the polling station actually is) is located just down the road at 38 Holmes Avenue and we are happy with the co-ordinates already provided by us.

PR #1765, opened 2019-04-09
add NEXT_CHARISMATIC_ELECTION_DATE setting

chris48s:
This will allow us to set `NEXT_CHARISMATIC_ELECTION_DATE = "2019-05-02"` in the settings on the deploy repo to temporarily ignore any elections that may or may not be happening on 2019-05-23 until we're ready to shift focus and deal with them.


symroe:
Nice. Let's revisit this when (if) we calm down a little, but this is a great win for the situation we find ourselves in.

PR #1841, opened 2019-04-24
run SpatialHub scripts in parallel

chris48s:
waiting on other things..


chris48s:
done this in #1846 now

PR #1846, opened 2019-04-25
Spatial hub imports

GeoWill:
Here's an initial bash at updating the `ScotlandSpatialHubImporter` class. It seems to behave as expected with a couple of scripts, I've attached the Aberdeen one as an example - the duplicate stations in district issue is unchanged from 2017.  
It would be good to get feedback on 71fb6f9  - I'm not sure if the `encoding` kwarg that I've added to the `ShpHelper` class is the best way to go as it's not obvious to me where one would set it in an import script. However it may be an expedient hack that gets us through to June.

### DIFF
--- a/polling_stations/apps/data_collection/management/commands/import_aberdeen.py
+++ b/polling_stations/apps/data_collection/management/commands/import_aberdeen.py
@@ -10,11 +10,11 @@ due to incomplete/poor quality data
 class Command(BaseScotlandSpatialHubImporter):
     council_id = "S12000033"
     council_name = "Aberdeen City"
-    elections = ["local.aberdeen-city.2017-05-04", "parl.2017-06-08"]
+    elections = ["europarl.2019-05-23"]
 
     def station_record_to_dict(self, record):
         # exclude duplicate/ambiguous codes
-        if record[1] in ["DG0205", "DG0206", "CS1008"]:
+        if record[0] in ["DG0205", "DG0206", "CS1008"]:
             return None
 
         return super().station_record_to_dict(record)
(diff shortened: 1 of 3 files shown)


chris48s:
I wouldn't change the default to `latin1` directly in `ShpHelper` probably breaks other things that are relying on it defaulting to `utf-8`.

If you revert that and change the `parse_string` method in `BaseScotlandSpatialHubImporter` to

```py
    def parse_string(self, text):
        try:
            return text.strip().decode("latin-1")
        except AttributeError:
            return text.strip()
```

does that solve the same problem? If not, I'd switch to passing encoding through the `options` param in `FileHelperFactory.create()` in the same way we do with CSV encoding/delim etc:

https://github.com/DemocracyClub/UK-Polling-Stations/blob/c35acc8f9a1c6485389b9ae9dfeb8f26b8989953/polling_stations/apps/data_collection/filehelpers.py#L166-L182

polling_stations/apps/data_collection/management/commands/import_aberdeen.py
```diff
@@ -10,11 +10,12 @@
 class Command(BaseScotlandSpatialHubImporter):
     council_id = "S12000033"
     council_name = "Aberdeen City"
-    elections = ["local.aberdeen-city.2017-05-04", "parl.2017-06-08"]
+    elections = ["europarl.2019-05-23"]
+    private = True
```
chris48s: This shouldn't be needed. `BaseScotlandSpatialHubImporter.data_path()` should respect the `PRIVATE_DATA_PATH` setting

GeoWill:
Implemented as suggested by using the `FileHelperFactory.create()`.

PR #1988, opened 2019-05-13
Import Durham

mtravis:


chris48s:
ta. Can you remove 19baa45  from this branch

PR #1991, opened 2019-05-13
Import Daventry & update misc_fixes 

mtravis:
UPRN added to misc_fixes


chris48s:
Can you get rid of 19baa45 and 28953e2 off this branch - they're covered in other PRs. Also are you able to resolve the merge conflict on misc_fixes (both `10014047099` and `28033460` should be in the list)

PR #1996, opened 2019-05-14
Import scripts scotland

GeoWill:


chris48s:
Cheers. Merging _most_ of this, but I've rebased and picked out the commit for Aberdeenshire into #2020 as I spotted an issue to follow up there. Can you have a look at that one when you get a chance (non-urgent).

PR #2011, opened 2019-05-16
Import scripts scotland 20190516

GeoWill:
This is ready for review, but relies on the change made in 914b1ef which is in #1996, so that should go in first.


GeoWill:
This may also end up with merge conflicts once #1996 is merged, so so I'll keep an eye to rebase if needed.

PR #2019, opened 2019-05-17
Import scripts scotland 20190517

GeoWill:
This is ready for review, but relies on the change made in 914b1ef which is in #1996, so that should go in first.
Don't think there should be any merge conflicts but bounce it back if there are.

### DIFF
--- a/polling_stations/apps/data_collection/management/commands/import_eilean_siar.py
+++ b/polling_stations/apps/data_collection/management/commands/import_eilean_siar.py
@@ -1,0 +1,22 @@
+from data_collection.management.commands import BaseScotlandSpatialHubImporter
+
+"""
+Note:
+This importer provides coverage for 45/47 districts
+due to incomplete/poor quality data
+"""
+
+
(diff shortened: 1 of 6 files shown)


chris48s:
Ta

> I confess to not being totally clear why the seen_codes hack
was needed. But it stopped the script trying to load things twice.

Its because you've got the same station in the stations file twice, so there are 2 stations (in the same place) with the slug `'albert-halls'`. This allows us to pass it over the second time because we already inserted the station with both district codes the first time :)
This is fine, as long as they're definitely the same place, and you haven't got 2 Albert Hall's in different places, at which point it becomes.. not fine.

---

One additional thing we probably could have done here is in the cases where we've got a district we can't match to any point in the stations file, but the districts file does contains a full address (as opposed to just the station name), we could import just the address with a null location so we show that with no map. Given the time constraints, what we've done here is fine and there is more to be gained by working through some of the backlog for higher population areas than trying to fill in the odd missing districts in Scotland. I'll create an issue for it as we will probably need to import data from SpatialHub again in future with similar issues.

PR #2023, opened 2019-05-18
import script for Glasgow City

chris48s:
Extracted from #2011 You should be able to push to this branch.

This one has broadly the same problem as Aberdeenshire - stations and districts files disagree for various codes e.g:

District `AN3723`

* districts file says its "Jordanhill Parish Church Hall, Munro Road (28 Woodend Drive), G13 1QT"
* stations file says its "KELVINDALE PRIMARY SCHOOL, 11 DORCHESTER PLACE, G12 0BP"

District `CC0407`

* districts file says its "City Building Depot, 32 Greenholme Street, G44 4DU"
* stations file says its "CATHCART CONGREGATIONAL CHURCH HALL, 15 GARRY STREET, G44 4AY"

I ended up binning this one in 2017 for this reason, but given Glasgow covers mor than half a million population, I reckon its probably worth chucking a few more hours at trying to get coverage on the districts where the 2 files agree. Do you reckon its worth having a go at matching them on both code and address slug and see where we get to with that..

### DIFF
--- a/polling_stations/apps/data_collection/management/commands/import_glasgow_city.py
+++ b/polling_stations/apps/data_collection/management/commands/import_glasgow_city.py
@@ -1,5 +1,11 @@
 from data_collection.management.commands import BaseScotlandSpatialHubImporter
 
+"""
+Note:
+This importer provides coverage for 245/258 districts
+due to incomplete/poor quality data
+"""
+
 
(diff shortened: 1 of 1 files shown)


chris48s:
nice one :+1:

PR #2028, opened 2019-05-21
Import scripts 20190520

GeoWill:


polling_stations/apps/data_collection/management/commands/import_barnet.py
```diff
@@ -4,21 +4,16 @@
...
-            rec = super().address_record_to_dict(record)
-            rec["polling_station_id"] = ""
-            return rec
-
-        if record.postcode == "N20 0RB":
-            return None
-
-        return super().address_record_to_dict(record)
+        if uprn == "200196535":
+            rec["accept_suggestion"] = False
```
chris48s: this function needs to return something :)

GeoWill: But I found this great way to make all the errors go away...

PR #2037, opened 2019-05-21
Europarl home page text

symroe:
Screenshot 2019-05-21 19 20 05

After making this change, I'm not any more convinced it's useful. One to chat about on the call tomorrow.

As I'd done the work of adding a card that only shows up when polls are open, I thought I'd PR it anyway.

4e229eb changes the "we don't know" page a bit – I think it's slightly better wording than before (e.g not using "data"). Not going to care too much if I'm wrong though. Another thing to ~bike shed~ talk about tomorrow.

### DIFF
--- a/polling_stations/apps/data_finder/features/index.feature
+++ b/polling_stations/apps/data_finder/features/index.feature
@@ -24,7 +24,6 @@ Feature: Check Postcodes
     Then I submit the only form
     Then The browser's URL should contain "/address/5/"
     And I should see "Contact Foo Council"
-    And I should see "We don't have data for your area."
     And No errors were thrown
 
     Scenario: Check postcode without address picker
@@ -46,7 +45,6 @@ Feature: Check Postcodes
     Then I select "My address is not in the list" from "address"
     Then I submit the only form
     Then I should see "Contact Foo Council"
-    And I should see "We don't have data for your area."
     And No errors were thrown
 
     Scenario: Check multiple councils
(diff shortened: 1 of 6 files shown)


symroe:
(Will ignore tests until / if we decide to make these changes)

chris48s:
I think the landing page change is good and we should do it.

I agree we should get rid of the wording "We don't have data for your area". I think that stems from the time when we had the email signup trying to get people to sign up to hassle their council to publish data (which we removed because nobody cared).

I reckon "We don't know your _exact_ polling station" is also a bit odd though. What do you think of:

"We don't know where you should vote" or
"We don't know where your polling station is"?

chris48s:
Once we've settled on wording, I think we just need to update the occurrences of that wording in the tests to get them passing:

https://github.com/DemocracyClub/UK-Polling-Stations/blob/076bc096bea151f170037a273231ac13ea8747ca/polling_stations/apps/data_finder/features/index.feature#L27
https://github.com/DemocracyClub/UK-Polling-Stations/blob/076bc096bea151f170037a273231ac13ea8747ca/polling_stations/apps/data_finder/features/index.feature#L49

symroe:
Agree about the "exact" wording, re-reading it. 

Maybe it's not useful telling people we don't know (It's not about us). Maybe "To find out where to vote, look at your poll card etc" is more of a useful approach – getting directly to the thing the person needs to do / know.

chris48s:
Yeah we could just remove that line completely and leave it as

> **Contact [your council name]**
> Your polling station address should be printed on your polling card...

to keep it actionable

PR #2038, opened 2019-05-22
Adding some punctuation

JoeMitchell:


symroe:
👍 

Do you reckon this line should have the comma too? https://github.com/DemocracyClub/UK-Polling-Stations/blob/84fa7110739895f73c090d28f68dd05be8e7c99f/polling_stations/templates/fragments/you_dont_need_poll_card.html#L16

JoeMitchell:
Yes, think so. 

I actually think we could remove that line for simplicity...

JoeMitchell:
Unless you think it's achieving something important? 

I think we could just go

>Polling stations are open from 7am to 10pm today. 

>You don't need your poll card or any other ID. 

>You must vote at your assigned polling station.

JoeMitchell:
(My reasoning here is that if people are hitting this page, bouncing and going back to Facebook, then short, simple text is vital. The other sentence is really long and just won't be read.)

symroe:
Agree about length, but I think, just based on talking to people about this in general, that it's better to be really verbose – people seem not to believe it.

symroe:
Folded this in to #2037 manually

PR #2077, opened 2019-08-15
add a SHOW_MAPS setting

chris48s:
We've talked a bit about whether in the event of a snap election we could increase coverage and still provide a useful service but eliminate one of the major sources of errors/checking by just showing polling station address, but no map/directions etc. This is what some of our major data consumers do (e.g: the Electoral Commission). We may have a chance to find out in the near future.

This PR gives us a global setting we can use to just stop serving maps via the front-end or any point/location data via the API. My thinking with this approach is that its still useful to import/store points in the DB and be able to use them in dev/QGIS for checking purposes, but just not serve maps to users.

The downside with this approach is its a sledgehammer, not a scalpel. We can't turn it on for some councils but off for others, for example.

### DIFF
--- a/polling_stations/apps/api/address.py
+++ b/polling_stations/apps/api/address.py
@@ -3,6 +3,7 @@ from rest_framework import serializers
 from rest_framework.permissions import IsAuthenticatedOrReadOnly
 from rest_framework.response import Response
 from rest_framework.viewsets import ViewSet
+from django.conf import settings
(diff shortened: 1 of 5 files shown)


chris48s:
Abandoning this. Instead of doing it as a global setting, we can do this on specific import scripts by overriding `get_station_point()`. That will allow us to do it in a slightly more granular way.

PR #2079, opened 2019-08-21
add every election THRESHOLD_DAYS setting

chris48s:
This is a bit obscure, but the problem I'm trying to solve here is: We would like to start creating ids for elections in May next year, but as soon as we do that, we shift to saying "we don't know" instead of "no upcoming elections" in all the areas where we do that. This allows us to set a number of days (opening gambit: 90 days) beyond which we'll ignore upcoming elections (in the front-end, for the purpose of serving a polling station result).
That means we can start creating IDs for elections we know are going to happen in May next year but ignore them in WDIV until nearer the time.

### DIFF
--- a/polling_stations/apps/data_finder/helpers/every_election.py
+++ b/polling_stations/apps/data_finder/helpers/every_election.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 import requests
 from django.conf import settings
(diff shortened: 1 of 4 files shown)


chris48s:
Revisit this: decided to deal with it by changing the definition of "current" in EE and then filtering elections where both `current=1` and `future=1` in client apps (apply to WCIVF too)

chris48s:
replaced by #2090

Issue #1401, opened 2018-06-28
Update to boto3


chris48s:
This project uses the legacy version of boto:

https://github.com/DemocracyClub/UK-Polling-Stations/blob/f218637d2d8ace4857ec467e49825e63f9bd5574/requirements/base.txt#L15

We should update it to `boto3`. This will require some changes to https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_collection/s3wrapper.py

sam-drew:
Does this still need doing? I think I could help

symroe:
Hi Sam!

Thanks!

This is, understandably, a bad day for us to focus on anything but running the service, but help would be great, say, in the next few weeks!

If you've not already, it might be worth joining our Slack and chatting to us when it's all over?

http://slack.democracyclub.org.uk/

sam-drew:
Of course, good luck!

PR #2096, opened 2019-10-03
self serve uploads

chris48s:
Right then.. this is some code that moves this idea forward. There are a number of TODOs and questions.

The very top-level summary is this provides a form which an authenticated user can use to dump some files into a S3 bucket based on the method described at https://blog.webiny.com/upload-files-to-aws-s3-using-pre-signed-post-data-and-a-lambda-function-7a9fb06d56c1

At the moment I'm assuming this will just be piloted with an internal user (or 2).

Some design decisions I have made:

* In the DC theme and elsewhere on the site, we're still using jQuery but I've opted to go jQuery-less for this feature and just use plain javascript as that's the direction we want to head in future. I've tried to keep to very compatible syntax/language features as far as possible. The biggest choice I've made here affecting client-side compatibility is the use of promises (see [compatibility](https://caniuse.com/#feat=promises) ), but I feel like using them here is buying us something here.
* I think I've decided to write files to a temp bucket first instead of writing files directly to the `pollingstations-data` bucket. I'm envisaging that the lambda hook will move files to `pollingstations-data` if some checks are passed, but we can talk about the tradeoffs of this.
* Accepting the JSON object we post to WDIV in order to obtain the pre-signed post data means we need to validate an input object which isn't completely trivial. I decided to try out [marshmallow](https://marshmallow.readthedocs.io/en/stable/) for this. I've been meaning to have a look at it for a while - its quite nice. We'll also have the same requirement to validate a JSON object once we start accepting a POST request from the lambda hook, so I'll probably use the same approach then.
* For the moment, I've assumed we're going to just accept CSV/TSV. In a future iteration, we could expand to accept zipfiles here. I'm assuming that "we want to send you shapfiles/kml/etc" is sufficiently rare that it is out-of-scope for this.

High-level TODOs and questions:

* For the purposes of writing to a temp bucket, I'm just generating a filename based on a timestamp because that's easy to work out. When we label them by hand, we group data by election id and then `/Version1`, `/Version2` etc. I need to think a bit more about how its sensible to organise this as we move data to the `pollingstations-data` bucket in the lambda hook.
* I haven't implemented any bahaviour on a sucessful upload. One of the things I hit here was that the v3 base theme has an `error-summary` style but there is not like a corresponding `success-summary` or `info-summary`. It might make sense to redirect back to a list/summary page here though.
* TODO: Put a logout link somewhere, link `/upload_files` from somewhere. Talk about menu elements in dc-base-theme.
* TODO: I haven't written any tests for any of this yet. Feels like we want to be more confident in the approach before formalising it


polling_stations/apps/file_uploads/templates/file_uploads/index.html
```diff
@@ -0,0 +1,178 @@
...
+        for (var i=0; i<data.files.length; i++) {
+          uploads.push(
+            uploadFileToS3(data.files[i], 'file'.concat(i), 'progressBar'.concat(i))
+          );
+        }
+        Promise.all(uploads).then(
+          function(data) {
+            console.log('done!');
+            console.log(data);
+            document.getElementById("submit").disabled = false;
```
chris48s: TODO: implement some kind of success behaviour.. not sure what that looks like yet

polling_stations/apps/file_uploads/templates/file_uploads/index.html
```diff
@@ -0,0 +1,178 @@
...
+          },
+          handleError
```
chris48s: TODO: there is an edge case here. If we upload 2 files and the first one fails I don't want to re-enable the upload button until after the second upload completes. I need to handle the rejections as described in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#Promise.all_fail-fast_behaviour to do that.

polling_stations/apps/file_uploads/templates/file_uploads/index.html
```diff
@@ -0,0 +1,178 @@
...
+        <button type="submit" class="button" id="submit">Upload</button>
+
+        </div>
+      </form>
+
+</div>
+{% endblock content %}
+
+{% block in_page_javascript %}
+<script>
```
chris48s: TODO: I might move all this out to a seperate .js file

polling_stations/apps/file_uploads/views.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Do some basic picture checks on request body and files.
+        Obviously these values can be spoofed, so checks on
+        file extension and MIME type etc are just to "fail fast".
+        We will check the files _properly_ in the lambda hook.
+        """
+        try:
+            UploadRequestSchema().load(body)
+        except MarshmallowValidationError:
+            raise DjangoValidationError("Request body did not match schema")
```
chris48s: TODO: it would be really nice here to report a more meaningful error but its also a pain to parse `e.messages` in a general way. I'm going to leave it for now. We'll swallow it on the front-end anyway.

polling_stations/apps/file_uploads/views.py
```diff
@@ -0,0 +1,137 @@
...
+from boto.pyami.config import Config
+from botocore.exceptions import ClientError
+from marshmallow import Schema, fields, validate
+from marshmallow import ValidationError as MarshmallowValidationError
+
+from councils.models import Council
+
+
+MAX_FILE_SIZE = 150000000  # 150Mb
+S3_BUCKET_NAME = "pollingstations-data-dev"
```
chris48s: TODO: set up a real bucket

polling_stations/apps/file_uploads/views.py
```diff
@@ -0,0 +1,137 @@
...
+        validate=validate.Length(min=1, max=2),
+    )
+
+
+def get_s3_client():
+    config = Config()
+    access_key = config.get_value(settings.BOTO_SECTION, "aws_access_key_id")
+    secret_key = config.get_value(settings.BOTO_SECTION, "aws_secret_access_key")
+    return boto3.client(
+        "s3", aws_access_key_id=access_key, aws_secret_access_key=secret_key
```
chris48s: I'm aware this is a bit wierd. We need to use `boto3` as `generate_presigned_post` isn't implemented in `boto`. Simultaneously, I don't want to address #1401 or make the same credentials available in 2 ways, so I am mixing `boto.pyami.config` to get the auth credentials in the boto style with `boto3.client` to make the connection. I think its better to do this for now and review how we make S3 credentials available to WDIV under #1401 rather than scope creep here.

polling_stations/apps/file_uploads/templates/file_uploads/index.html
```diff
@@ -0,0 +1,178 @@
...
+
+{% load i18n %}
+{% load static %}
+
+{% block extra_page_css %}
+<style>
+  button:disabled {
+    background-color: lightgrey;
+    border-color: darkgrey;
+  }
```
chris48s: It would be quite nice to have a disabled button style in the base theme. In general there are a few "trying to build a CRUD app with the v3 theme" issues I hit with this PR. I hacked this one, but it really tested the limits of my CSS skills ;)

chris48s:
Gah I meant to create this as a draft PR and forgot, but you can't go backwards :( oh well

chris48s:
The bit of info missing from here was how this fits into a high-level view of the entire process. The full process I'm envisaging looks like:

1. Client submits request to WDIV to upload files
2. WDIV issues an authorization to upload to S3 bucket
3. Client uploads file to S3 bucket
4. Lambda hook picks up file from S3, performs some checks on it, raises a GH issue, etc
5. Once its run, lambda hook POSTs a report to an endpoint on WDIV reporting the outcome of the checks

This PR addresses steps 1-3. The code for 4 will live in a separate repo (not started yet) and be deployed to lambda independently. The endpoint for 5 will live in this repo, but I've not attempted to do anything with it in this PR.

symroe:
I've started collecting the use cases for base theme designs in https://github.com/DemocracyClub/dc_base_theme/issues/49

polling_stations/apps/file_uploads/templates/file_uploads/index.html
```diff
@@ -0,0 +1,178 @@
...
+
+{% block content %}
+<div class="card">
+
+    <noscript>
+      <div class="card error-summary">
+          <h2>Javascript disabled</h2>
+          <p>To upload files, please enable javascript.</p>
+      </div>
+    </noscript>
```
symroe: Given this is for a tiny set of users that we know, then feel free to ignore this comment, but given what you say about specific JS requirements, it might be worth adding some sort of feature check in future that tells the user they are using an unsupported browser, rather than just a boolean JS enabled/disabled.

Also, this feels like something that could live in the base theme as a JS utility function.

polling_stations/apps/file_uploads/views.py
```diff
@@ -0,0 +1,137 @@
...
+from marshmallow import ValidationError as MarshmallowValidationError
+
+from councils.models import Council
+
+
+MAX_FILE_SIZE = 150000000  # 150Mb
+S3_BUCKET_NAME = "pollingstations-data-dev"
+
+
+class FileSchema(Schema):
```
symroe: Fine for MVP, but I reckon `FileSchema`, `UploadRequestSchema` and `get_s3_client` could be moved out of the view file at some point.

polling_stations/apps/file_uploads/views.py
```diff
@@ -0,0 +1,137 @@
...
+    return boto3.client(
+        "s3", aws_access_key_id=access_key, aws_secret_access_key=secret_key
+    )
+
+
+class FileUploadView(LoginRequiredMixin, TemplateView):
+    template_name = "file_uploads/index.html"
+
+    @method_decorator(staff_member_required)
+    @method_decorator(ensure_csrf_cookie)
```
symroe: Could these be on `dispatch` to ensure all (HTTP) methods get covered?

polling_stations/apps/file_uploads/views.py
```diff
@@ -0,0 +1,137 @@
...
+    def get(self, request, *args, **kwargs):
+        return super().get(self, request, *args, **kwargs)
+
+    def get_context_data(self, **context):
+        context["councils"] = (
+            Council.objects.all()
+            .filter(~Q(council_id__startswith="N09"))
```
symroe: This is the same as `.exclude(council_id__startswith="N09")`, no?

polling_stations/apps/file_uploads/views.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @method_decorator(staff_member_required)
+    @method_decorator(csrf_protect)
+    def post(self, request, *args, **kwargs):
+        body = json.loads(request.body)
+
+        try:
+            self.validate_body(body)
+        except DjangoValidationError as e:
+            return JsonResponse({"error": e.message}, status=400)
```
symroe: I know this isn't the file content, so should be small in normal usage, but is it worth us validating the upload size here, before parsing it as JSON? Either in nginx or in this view, or both?

polling_stations/apps/file_uploads/templates/file_uploads/index.html
```diff
@@ -0,0 +1,178 @@
...
+
+    <form id="file_upload_form" method="POST">
+
+        <div class="card error-summary" id="error" hidden></div>
+
+        <p>
+          <label for="council_id">Council</label>
+          <select id="council_id" name="council_id">
+            {% for council in councils %}
+            <option value="{{ council.council_id }}">{{ council.name }}</option>
```
symroe: Two stretch goals here, post MVP:

1. select2 for filtering / type ahead
2. selecting a default from a GET param, for deep linking from the tracker sheet / inbox / whereever

polling_stations/apps/file_uploads/views.py
chris48s: In 80e9bdf I've moved the auth behaviour to a shared mixin (which does apply to `.dispatch()`. The CSRF decorators are necessarily different for each method: `.get()` uses `@method_decorator(ensure_csrf_cookie)` whereas `.post()` needs `@method_decorator(csrf_protect)`

polling_stations/apps/file_uploads/views.py
chris48s: Yeah `.filter(~Q(condition))` and `.exclude(condition)` are the same thing. I've switched the occurrences of `.filter(~Q(condition))` to `.exclude(condition)` in 6e9097e.

polling_stations/apps/file_uploads/views.py
chris48s: There's already a hard limit on the POST body size in the NGINX cofig: https://github.com/DemocracyClub/polling_deploy/blob/c819ad898ddeae06ae251eb8204a3e0adddd9e53/files/conf/nginx.conf#L61
I reckon that probably covers it tbh.

polling_stations/apps/file_uploads/templates/file_uploads/index.html
```diff
@@ -0,0 +1,182 @@
...
+          </select>
+        </p>
+
+        <p>
+          <label for="file0">File 1 (required)</label>
+          <input type="file" id="file0" name="file0" accept=".csv,.tsv" required />
+          <progress id="progressBar0" value="0" max="100"></progress>
+        </p>
+        <p>
+          <label for="file1">File 2 (optional)</label>
```
chris48s: TODO: allow >2 files to be uploaded

polling_stations/apps/file_uploads/templates/file_uploads/uploads_header.html
```diff
@@ -0,0 +1,3 @@
+<a href="{% url 'file_uploads:councils_list' %}">Home</a> •
+<a href="{% url 'file_uploads:file_upload_index' %}">Upload</a> •
+<a href="{% url 'logout' %}">Logout</a>
```
chris48s: TODO: This is an absolutely BARE minimum viable menu.

Style-wise, lets improve it pending https://github.com/DemocracyClub/dc_base_theme/issues/50 it'll do the job

Content-wise, it would be nice to have proper breadcrumbs here

polling_stations/apps/file_uploads/templates/file_uploads/file_detail.html
```diff
@@ -0,0 +1,46 @@
...
+      content: ":";
+  }
+  dd {
+      margin-left: 25%;
+  }
+  dd::after {
+      content: " ";
+      display: block;
+      clear: left;
+  }
```
chris48s: another one for the "stuff that should be in the base theme" bucket
https://github.com/DemocracyClub/dc_base_theme/issues/49#issuecomment-574300635

polling_stations/apps/file_uploads/templates/file_uploads/index.html
chris48s: I've done the second one of these in b754ee7

Now that I've done the other UI bits in this PR, I suspect that's actually going to end up being the main way we use this, so there might be a bit more fiddling to do here.

chris48s: Select2 now irrelevant as I've totally removed the generic upload page (with a drop-down) in 
a192baf

chris48s:
Having done a bunch of testing and tidied up the history a bit, I'm going to merge this and raise the remaining tasks as issues for future.

PR #2161, opened 2019-11-06
Blacken data_quality_report

GeoWill:
Black seemingly doesn't like trailing commas in methods or some such. This will shut it up when running `black .`.


GeoWill:
Old version of Black hiding on PATH. Sorry!

PR #2164, opened 2019-11-06
Import script for St Helens (closes #2124)

### LINKED ISSUE #2124: Import E08000013-St Helens Metropolitan Borough Council
Xpress DC

GeoWill:
https://trello.com/c/5HLxjs3v

### DIFF
--- a/polling_stations/apps/data_collection/management/commands/import_st_helens.py
+++ b/polling_stations/apps/data_collection/management/commands/import_st_helens.py
@@ -3,44 +3,60 @@ from data_collection.management.commands import BaseXpressDemocracyClubCsvImport
 
 class Command(BaseXpressDemocracyClubCsvImporter):
     council_id = "E08000013"
-    addresses_name = "local.2019-05-02/Version 1/Democracy_Club__02May2019helens.tsv"
-    stations_name = "local.2019-05-02/Version 1/Democracy_Club__02May2019helens.tsv"
-    elections = ["local.2019-05-02"]
+    addresses_name = "parl.2019-12-12/Version 1/Democracy_Club__12December2019sthel.tsv"
+    stations_name = "parl.2019-12-12/Version 1/Democracy_Club__12December2019sthel.tsv"
+    elections = ["parl.2019-12-12"]
     csv_delimiter = "\t"
+    allow_station_point_from_postcode = False
 
     def station_record_to_dict(self, record):
-        if record.polling_place_id == "2694":
-            record = record._replace(polling_place_postcode="WA12 0NH")
+        if record.polling_place_id == "3607":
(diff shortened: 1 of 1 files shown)


chris48s:
Had a quick look over this one and there's quite a few obviously misplaced addresses in pickers (which are then missing from the postcodes they're supposed to be assigned to)

examples (but not necessarily limited to):

79 Newlove Avenue
WA119TB

32 Cunningham Court
WA101JT

6 Alexander Court
WA95GG

22 Dorothy Street
WA92RL

107A Springfield Road
WA93RR

Suspect some of these could also be errors in addressbase.
Any chance you can have a look into these? If the simplest thing is to bin data rather than fix it, do.
If there addressbase errors we need to feed back to OS, can you bung them in the spreadsheet. Ta

polling_stations/apps/data_collection/management/commands/import_st_helens.py
```diff
@@ -3,44 +3,45 @@
...
-            rec["accept_suggestion"] = True
+        if uprn == "39086246":
+            return None
+
+        # UPRNs 39089002 through 39089022 look fishy. Not sure if it's council or addressbase.
+        if "Alexander Court" in record.addressline1:
+            rec["uprn"] = ""
+
+        if uprn == "39091880":  # 79 Newlove Avenue
+            rec["postcode"] = "WA104DS"
```
chris48s: This doesn't actually fix this one.
You might need to set `rec["accept_suggestion"] = False` here too

polling_stations/apps/data_collection/management/commands/import_st_helens.py
```diff
@@ -3,44 +3,45 @@
...
+        if uprn == "39091918":  # 32 Cunningham Court
+            rec["postcode"] = "WA104GA"
+        if uprn == "39099008":  #
+            rec["postcode"] = "WA95RL"
+        if uprn == "39097964":  # 107A Springfield Road
+            rec["postcode"] = "WA103RR"
```
chris48s: Also, there's definitely a few more of these to spot e.g:

WA102TL
14A Hartford Road

WA102TJ
53 The Boulevard

polling_stations/apps/data_collection/management/commands/import_st_helens.py
GeoWill: I only checked the fixed postcode :roll_eyes:

GeoWill:
Rebased and squashed, hope there's not more - if there are let me know how you're finding them!

chris48s:
don't think your rebase went to plan :grimacing: 
if its easier to copy the script to a new branch and create a new PR for it, go for it..

PR #2181, opened 2019-11-07
Update Runnymede for parl.2019-12-12

alexsdutton:


chris48s:
Cheers. I've pushed a quick fixup to this one. Although we've got gridrefs for most stations, we've only got partial coverage:

Screenshot at 2019-11-08 12-03-36

I've set `allow_station_point_from_postcode` so we won't fall back to postcode for the ones with no gridref.

PR #2269, opened 2019-11-13
Council fixes

GeoWill:


polling_stations/apps/data_collection/management/commands/import_welwyn_hatfield.py
```diff
@@ -30,6 +30,14 @@ def get_station_point(self, record):
         return None
 
     def station_record_to_dict(self, record):
+        if record.code == "BFB":
```
chris48s: we're doing this same one in a slightly different way a few lines down. Just for tidyness, can we just do it once.

chris48s:
Aah I mean the _other_ place we were doing the same thing a few lines down ;)
I've pushed a fixup

PR #2275, opened 2019-11-14
Import South Kesteven (closes #2195)

### LINKED ISSUE #2195: Import E07000141-South Kesteven District Council
Xpress DC

alexsdutton:


polling_stations/apps/data_collection/management/commands/import_south_kesteven.py
```diff
@@ -43,16 +48,24 @@ def address_record_to_dict(self, record):
...
+            #     "10007263151",  # PE91JP -> PE92AP : Flat 3 Caymen House, Foundry Road, Stamford, Lincs
+            #     "10007263152",  # PE91JP -> PE92AP : Flat 4 Caymen House, Foundry Road, Stamford, Lincs
         ]:
             rec["accept_suggestion"] = True
 
         if uprn in [
             "10007248507",  # NG321LH -> NG318BY : Park Cottage, Park Lane, Denton, Grantham, Lincs
             "200001873982",  # PE91QP -> PE91QL : Hazel Grove, Emlyns Street, Stamford, Lincs
+            "10007257592",  # PE100TT -> PE100TG : Gatehouse Cottage, Fen Road, Dowsby, Bourne, Lincs
+            # Don't know
```
chris48s: Remember if you don't know, its better to do nothing here as opposed to explicitly going with either the addressbase ( `rec["accept_suggestion"] = True` ) or council data ( `rec["accept_suggestion"] = False` ). If we just leave it as a warning, we'll discard the data and say "we don't know", which is the safe option if we can't make a good decision.

alexsdutton: Noted :-). Now removed

PR #2369, opened 2019-11-20
Import Gedling (closes #2225)

### LINKED ISSUE #2225: Import E07000173-Gedling Borough Council
Xpress DC

alexsdutton:
I'm not too sure what to do with the stray properties for which this currently returns `None`. A second pair of eyes here would be very useful!


GeoWill:
If you mean the two postcodes  ("NG5 8PQ" and "NG4 4PS") that you discard, I think what you've done looks like the safe option. You either have to keep them whole or throw them away - as you say. I also looked at the four uprns you `return None` for and would've done the same :slightly_smiling_face:

PR #2378, opened 2019-11-21
Import scripts 20191121

GeoWill:


polling_stations/apps/data_collection/management/commands/import_bedford.py
```diff
@@ -4,33 +4,42 @@
...
+            record = record._replace(yordinate="247440")
 
         return super().station_record_to_dict(record)
 
     def address_record_to_dict(self, record):
-        rec = super().address_record_to_dict(record)
         uprn = record.uprn.strip().lstrip("0")
 
+        # Lot's of junk UPRNs
+        record = record._replace(uprn="")
```
chris48s: Gah - I hadn't spotted these - I'd have bounced it back to them for a fresh export if I had.
If you're removing all the UPRNs here, how are you still matching on UPRN a file lines down with `if uprn in ["10024229957"]:`? Did you mean to just bin a subset of them?

polling_stations/apps/data_collection/management/commands/import_birmingham.py
```diff
@@ -1,43 +1,77 @@
...
+        rec = super().station_record_to_dict(record)
+
+        if record.pollingstationname == "St Mary and Ambrose Church Hall":
+            rec["postcode"] = "B5 7RA"
+            rec["location"] = Point(-1.904365, 52.458623, srid=4326)
+
+        if record.pollingstationnumber in ["145", "146"]:
+            rec[
+                "internal_council_id"
+            ] = f"{record.pollingstationnumber}-pentecostal-city-mission-church"
```
chris48s: This isn't doing what you want it to - that's why you've got >1500 addresses with invalid station refs when you run it.

polling_stations/apps/data_collection/management/commands/import_bedford.py
GeoWill: I thought I could get away with it by grabbing it in line 24 before I delete it. Do you want me to pull it out and send it back anyway?

chris48s: I get that you can set the "accept_suggestion" flag on the right record because you've grabbed the UPRN to a local var before binning it, but what I mean is if `rec` doesn't have a UPRN on it, there's no suggestion to accept. By the time you've finished with it, the dict is

```py
{
    'address': 'SKYLARKS, KIMBOLTON ROAD, BOLNHURST, BEDFORD',
    'postcode': 'MK44 2EY',
    'polling_station_id': 'NAV_1',
    'uprn': '',
    'accept_suggestion': True
}
```

..so you've told it to accept a suggestion, but it has no UPRN to look up in addressbase and make a suggestion with so we need to do the correction manually. Setting the flag probably does silence the warning even if we can't do anything with it though - lol.

If you reckon this one is pretty tidy apart from the UPRNs, I don't mind rolling with it. If you reckon its masking a bunch of problems or making it hard to check, we can go back and ask for a new export on Monday - I'll defer to your judgement.

polling_stations/apps/data_collection/management/commands/import_birmingham.py
GeoWill: I was only looking at half the problem :man_facepalming: .

polling_stations/apps/data_collection/management/commands/import_bedford.py
GeoWill: OK that makes sense. I'll have a look.

GeoWill:
OK I think those two fixups should sort the issues. With Bedford I think it's OK to go.

chris48s:
Nice one - cheers.

On the Birmingham one, we've got 2 places where we're essentially looking for the same situation with 2 different statements:

`if record.pollingstationnumber in ["145", "146"]:`
`if record.pollingstationname == "Naseby Youth & Community Centre":`

(I think both of them should be essentially matching the same records?)
Assuming that's true, just for the sake of future code archaeology, can we use the same check in both cases and stick a comment in explaining it so its a bit clearer why we've done this. Does that make sense?

If you do that and squash the fixups I'll merge :+1:

GeoWill:
Squashed :slightly_smiling_face:

PR #2353, opened 2019-11-15
Import Derby (closes #2205)

alexsdutton:

PR #2381, opened 2019-11-22
Import scripts 20191122

GeoWill:


chris48s:
Haven't had a proper look at this one yet, but it looks like 779672f is a stray commit that doesn't belong..

https://github.com/DemocracyClub/UK-Polling-Stations/pull/2353 got merged earlier today..

PR #2436, opened 2019-11-26
Import scripts 2019 11 26

GeoWill:


polling_stations/apps/data_collection/management/commands/import_dartford.py
```diff
@@ -3,44 +3,33 @@
...
-        if rec["internal_council_id"] in [
-            "54-hodsoll-street-ridley-village-hall",
-            "57-fawkham-and-hartley-church-centre",
+        # These are single records listed as 'Other Electors Address'
+        # no addresses are assigned. So probably don't cause problems, but
+        # dropping them to squash warnings.
+        if (record.pollingstationnumber, record.pollingstationname) in [
+            ("63", "Hodsoll Street & Ridley Village Hall"),
+            ("64", "Longfield & Hartley Scout HQ"),
+            ("66", "Fawkham and Hartley Church Centre"),
```
chris48s: What you've done here is right - no change needed.

Just to add a bit of explanation, I think the reason we sometimes see this is because this council runs all of the polling stations for a whole constituency even though some of it is outside their area.
In this case, those 3 stations are in the sort of "panhandle" of the [Dartford constituency](https://mapit.mysociety.org/area/66011.html) off the South East corner of [Dartford Council](https://mapit.mysociety.org/area/2358.html). So Dartford council book the venues for that, but all the addresses assigned to those 3 stations are all in Sevenoaks. Usefully all these venues also appear in the Sevenoaks data :)

I've not checked Watford in detail, but it will be the same deal.

PR #2448, opened 2019-11-28
Import scripts 20191128

GeoWill:
Plus a council update.

### DIFF
--- a/polling_stations/apps/data_collection/management/commands/import_cambridge.py
+++ b/polling_stations/apps/data_collection/management/commands/import_cambridge.py
@@ -4,15 +4,20 @@ from data_collection.management.commands import BaseHalaroseCsvImporter
 
 class Command(BaseHalaroseCsvImporter):
     council_id = "E07000008"
-    addresses_name = "local.cambridge.newnham.by.2019-08-08/Version 1/polling_station_export-2019-07-19.csv"
-    stations_name = "local.cambridge.newnham.by.2019-08-08/Version 1/polling_station_export-2019-07-19.csv"
-    elections = ["local.cambridge.newnham.by.2019-08-08"]
-    csv_encoding = "windows-1252"
+    addresses_name = (
+        "parl.2019-12-12/Version 1/polling_station_export-2019-11-15cam.csv"
+    )
+    stations_name = "parl.2019-12-12/Version 1/polling_station_export-2019-11-15cam.csv"
+    elections = ["parl.2019-12-12"]
+    allow_station_point_from_postcode = False
 
     def station_record_to_dict(self, record):
         rec = super().station_record_to_dict(record)
 
-        if rec["internal_council_id"] == "42-the-c3-centre":
+        if rec["internal_council_id"] == "17-st-philip-howard-church-centre":
+            rec["location"] = Point(0.16067, 52.18589, srid=4326)
+
+        if rec["internal_council_id"] == "38-the-c3-centre":
             rec["location"] = Point(0.1572, 52.2003, srid=4326)
 
         # user issue report #105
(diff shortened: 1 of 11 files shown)


polling_stations/apps/data_collection/management/commands/import_wealden.py
```diff
@@ -3,17 +3,28 @@
...
     csv_delimiter = "\t"
+    allow_station_point_from_postcode = False
 
     def address_record_to_dict(self, record):
         rec = super().address_record_to_dict(record)
         uprn = record.property_urn.strip().lstrip("0")
 
+        if record.addressline6 in [
+            "BN27 1DQ",
+            "TN210BA",
```
chris48s: this should probably have a space in it to match on `addressline6` in an Xpress export

polling_stations/apps/data_collection/management/commands/import_telford.py
```diff
@@ -4,18 +4,22 @@
...
-    elections = ["local.2019-05-02"]
+    elections = ["parl.2019-12-12"]
     csv_delimiter = "\t"
+    allow_station_point_from_postcode = False
 
     def address_record_to_dict(self, record):
         rec = super().address_record_to_dict(record)
         uprn = record.property_urn.strip().lstrip("0")
 
+        if record.addressline6 == "TF108AB":
```
chris48s: same here..

polling_stations/apps/data_collection/management/commands/import_kensington_and_chelsea.py
```diff
@@ -24,6 +24,24 @@ def station_record_to_dict(self, record):
             rec["location"] = Point(-0.200082, 51.491363, srid=4326)
             return rec
 
+            # Updates From Council
+            # Chelsea Old Town Hall, 165-181 King's Road, LONDON
+            record = record._replace(polling_place_uprnproperty_urn="217046329")
```
chris48s: This shouldn't be inside the if block for "St Cuthbert`s Centre"

polling_stations/apps/data_collection/management/commands/import_kensington_and_chelsea.py
```diff
@@ -24,6 +24,24 @@ def station_record_to_dict(self, record):
...
+            record = record._replace(polling_place_northing="178093")
+
+        # Kensington Central Library, 12 Phillimore Walk, LONDON
+        if record.polling_place_id == "":
+            record = record._replace(polling_place_uprnproperty_urn="217068403")
+            record = record._replace(polling_place_easting="525421")
+            record = record._replace(polling_place_northing="179598")
+
+        # Christ Church, Victoria Road, LONDON, W8 5RQ
+        if record.polling_place_id == "":
```
chris48s: neither of these conditions are going to match anything

polling_stations/apps/data_collection/management/commands/import_kensington_and_chelsea.py
```diff
@@ -24,6 +24,24 @@ def station_record_to_dict(self, record):
...
+            record = record._replace(polling_place_uprnproperty_urn="217089088")
+            record = record._replace(polling_place_easting="525968")
+            record = record._replace(polling_place_northing="179187")
```
chris48s: this isn't going to cause a problem as such, but if we've got UPRN and E/N we're only going to use one of them. I'm guessing in these cases, it doesn't make much difference (I've not checked), but we might as well just set the one we want to use.

polling_stations/apps/data_collection/management/commands/import_dorset.py
```diff
@@ -1,142 +1,122 @@
...
+    stations_name = "parl.2019-12-12/Version 1/merged.tsv"
+    elections = ["parl.2019-12-12"]
     csv_delimiter = "\t"
-    station_postcode_search_fields = [
-        "polling_place_postcode",
-        "polling_place_address_4",
-        "polling_place_address_3",
-        "polling_place_address_2",
-    ]
+    allow_station_point_from_postcode = False
```
chris48s: With this one, we've got so many corrections for cases where the map based on postcode wasn't quite right, I'm tempted to enable the maps on the basis we've probably already solved most of the problems that would get reported.. what do you reckon?

polling_stations/apps/data_collection/management/commands/import_cambridge.py
```diff
@@ -33,12 +38,17 @@ def address_record_to_dict(self, record):
         rec = super().address_record_to_dict(record)
         uprn = record.uprn.strip().lstrip("0")
 
+        if record.housepostcode in [
+            "CB5 8JJ",
+            "CB58HL",
```
chris48s: will this one match, or does it need a space?

polling_stations/apps/data_collection/management/commands/import_dorset.py
GeoWill: Yep, I think it's worth a try. There are a lot of polling stations, so still room for more errors, but it will be interesting to see if there's any difference in reaction.

PR #2477, opened 2019-12-02
Import Flintshire (closes #2329)

### LINKED ISSUE #2329: Import W06000005-Flintshire County Council
Xpress DC, use Version 2

alexsdutton:


polling_stations/apps/data_collection/management/commands/import_flintshire.py
```diff
@@ -0,0 +1,45 @@
...
+    allow_station_point_from_postcode = False
+
+    def station_record_to_dict(self, record):
+        if record.polling_place_id == "5162":
+            # Parish Hall/Neuadd Y Plwyf Halkyn/Helygain
+            record = record._replace(polling_place_postcode="CH8 8BU")
+        if record.polling_place_id == "5237":
+            # Ysgol Y LLan, Whitfield
+            record = record._replace(polling_place_uprn="10013703522")
+        if record.polling_place_id == "10090462071":
```
chris48s: This is not an Xpress polling station ID

polling_stations/apps/data_collection/management/commands/import_flintshire.py
```diff
@@ -0,0 +1,45 @@
...
+            # War Memorial Institute, Rhydymwyn
+            record = record._replace(polling_place_uprn="10090462071")
+
+        return super().station_record_to_dict(record)
+
+    def address_record_to_dict(self, record):
+        if record.addressline6 in [
+            "CH4 9BS" "CH8 7AX",
```
chris48s: If you put 2 strings next to each other like this in python, they get concatenated
so this is actually saying 

`if record.addressline6 in ["CH4 9BSCH8 7AX", "CH6 5NF"]:`

which is probably not what you wanted to happen.

alexsdutton: Yep. I missed a comma, and then black will have put them on the same line. Will fix :-)

polling_stations/apps/data_collection/management/commands/import_flintshire.py
alexsdutton: Thanks for the catch. Now fixed :-)

polling_stations/apps/data_collection/management/commands/import_flintshire.py
alexsdutton: Fixed now \o/

PR #2483, opened 2019-12-03
Import Selby (closes #2414)

### LINKED ISSUE #2414: Import E07000169-Selby District Council
Xpress DC

alexsdutton:


polling_stations/apps/data_collection/management/commands/import_selby.py
```diff
@@ -3,7 +3,22 @@
...
+        if record.polling_place_id == "4923":
+            # Lady Popplewell Centre, Beechwood Close, Sherburn in Elmet, Leeds
+            record = record._replace(polling_place_postcode="LS25 6HU")
+        return super().station_record_to_dict(record)
+
+    def address_record_to_dict(self, record):
+        if record.addressline6 in [
+            "LS24 9PA",
+        ]:
+            return None
```
GeoWill: Needs a: `return super().station_record_to_dict(record)`

alexsdutton: :sob: 

Now done

alexsdutton:
> Looks like there's something up with YO8 5DB. I'd have a quick re-check of the`Largest distances to polling places` postcodes as there's quite a lot of singletons which is often a red flag that a single address has been incorrectly geocoded.

I've now matched that suss property up to AddressBase, and it looks legit — just over a boundary. I've also had a look through the other long-distance singletons without UPRNs, and I reckon they look okay.

PR #2486, opened 2019-12-03
Import Redditch (closes #2421)

### LINKED ISSUE #2421: Import E07000236-Redditch Borough Council
Xpress DC

alexsdutton:


alexsdutton:
Postcode fixes done. The other one was correct, so I've explicitly dropped it to suppress the warning.

GeoWill:
> Postcode fixes done. The other one was correct, so I've explicitly dropped it to suppress the warning.

Thanks. I think explicitly dropping them is the way to go, because it means they've definitely been looked at.

PR #2489, opened 2019-12-03
Import scripts 20191203

GeoWill:
NB in polling_stations/apps/data_collection/management/commands/import_croydon.py I deleted all the postcodes for the stations, as these appear in the address and we're not using the postcodes to geocode them anyway. This saved me looking up all the codes in the Euros file and then the GE file. If they do need updating for some reason let me know :+1:


chris48s:
yeah it looks like in a previous election Croydon gave us a file with loads of missing postcodes and then we got updates by email: https://trello.com/c/htLwYd27 If they're all in Xpress now, happy days.

Merging with a fixup to make sure Bromsgrove closes the right issue when I merge it..

PR #2568, opened 2020-02-06
Add use postcode centroids flag

GeoWill:
We generally don't want this behavior in production because it introduced more
problems than it solved - "The church is at the other end of the street!" etc.
However when checking that polling districts and their stations look plausible
it is really useful. So add the flag:
	$ manage.py import_place -p
and the postcode centroids will be used to create station points.

I haven't added tests because as far as I can tell the other flags aren't. Can do if it would be valuable.

### DIFF
--- a/polling_stations/apps/data_collection/base_importers.py
+++ b/polling_stations/apps/data_collection/base_importers.py
@@ -73,6 +73,7 @@ class BaseImporter(BaseCommand, PostProcessingMixin, metaclass=abc.ABCMeta):
     logger = None
     batch_size = None
     imports_districts = False
+    use_postcode_centroids = False
 
     def write_info(self, message):
         if self.verbosity > 0:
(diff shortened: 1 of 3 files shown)


GeoWill:
That was useful, because I hadn't committed swapping the dashes to underscores...

GeoWill:
Reminder: s/georeference/geocode/

chris48s:
Now that we have this, it makes setting the `allow_station_point_from_postcode` on an individual import script completely meaningless. We can set `allow_station_point_from_postcode = True` or `allow_station_point_from_postcode = False` and either way, it will get overridden by whether we do or don't pass the `-p` flag.

That's fine (its what we're aiming for), but we've got some-hundred import scripts that are explicitly setting a value for that. Its not causing a problem as such, but its also a bit confusing. Imagine you're new to the codebase:

> what does the `allow_station_point_from_postcode` value on all these import scripts do?

> oh nothing :smile: 

Do you reckon its worth doing a massive find + replace now to clean them all up, or shall we just gradually sort them out over the next few months?

GeoWill:
:see_no_evil: 
I _think_ if it's just you, Alex and me then slight preference for doing it gradually? Reason being I think there might be one or two where it's set to `True` and I think that will be useful (FSVO) when updating the scripts.
If there are other folk coming to this fresh then I'd probably just get rid as it's one less thing to explain/overwhelm.

Issue #2553, opened 2020-01-30
Provide a suggestion/auto-correct invalid postcodes


chris48s:
https://github.com/DemocracyClub/UK-Polling-Stations/blob/6f9cf568d69857ba1f5a841c399d0b09153c5c9e/polling_stations/apps/data_collection/data_types.py#L148-L158

Look up the UPRN, if the match quality is good enough, auto-correct it, otherwise provide a suggestion. It might be that the easiest way to handle this is to move where we perform this check.

GeoWill:
closed via  #2573

PR #2573, opened 2020-02-10
Attempt to use addressbase data to correct invalid postcodes

GeoWill:
Just parking this here for now. It was a stab at beginning to tackle #2553 By getting a suggestion from addressbase where we could. However when I started thinking about testing it I realised that there's some duplication/overlap with the `handle_invalid_uprns method`. 
Any drive by comments welcome, but currently planning to rethink the approach.

### DIFF
--- a/polling_stations/apps/data_collection/data_types.py
+++ b/polling_stations/apps/data_collection/data_types.py
@@ -44,9 +44,6 @@ District = namedtuple(
 )
 
 
-postcode_validator = GBPostcodeField()
-
-
 class CustomSet(metaclass=abc.ABCMeta):
     def __init__(self):
         self.elements = set()
@@ -145,17 +142,8 @@ class AddressList:
             return
 
         if address["slug"] not in self.seen:
-            try:
-                postcode_validator.clean(address["postcode"])
-                self.elements.append(address)
-                self.seen.add(address["slug"])
-            except ValidationError:
-                self.logger.log_message(
-                    logging.WARNING,
-                    "Discarding record with invalid postcode:\n%s\n",
-                    variable=address,
-                    pretty=True,
-                )
+            self.elements.append(address)
+            self.seen.add(address["slug"])
         else:
             self.logger.log_message(
                 logging.DEBUG,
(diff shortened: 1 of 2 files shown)


polling_stations/apps/data_collection/data_types.py
```diff
@@ -164,6 +164,68 @@ def append(self, address):
...
+            if record["uprn"] not in addressbase_data:
+                invalid_postcode_warning(
+                    record, extras=" and record UPRN not in addressbase"
+                )
+
+            addressbase_record = addressbase_data[record["uprn"]]
+            if "postcode" in addressbase_record:
+                # At the moment this just suggests a correction from ABP if possible.
+                # It would be better to catch the obvious errors and just fix them.
+                # eg. '0'=>'O', 'I|L'=>'1', etc.
```
chris48s: How about using the same approach here that we use in `handle_invalid_uprns()`? Something like:

```py
if fuzzy_match:
    match_quality = fuzz.partial_ratio(
        record["address"].lower().replace(",", ""),
        addressbase_record["address"].lower().replace(",", "")
    )
    if match_quality >= match_threshold:
        # accept the correction from addressbase
    else:
        # bounce it back as a suggestion
```

That's probably going to do the job for us in most cases.

GeoWill:
This adds another rather large method to data_types.py :grimacing: . However I think a lot of the length, and complexity, is in providing useful logging messages which should save thinking when doing the import scripts, and so is hopefully justified. 
I did have a stab at factoring out a method to generate the warning messages, but felt like it was a layer of abstraction too many.

polling_stations/apps/data_collection/data_types.py
```diff
@@ -164,6 +152,91 @@ def append(self, address):
...
+                        logging.INFO,
+                        "Replacing %s with %s for record:\n%s\n",
+                        variable=(
+                            record["postcode"],
+                            addressbase_record["postcode"],
+                            record,
+                        ),
+                    )
+                    return True
+                except ValidationError:
```
chris48s: I reckon we can drop the second call to `clean()`/`except ValidationError` here and assume that if the postcode came from addressbase, it is valid. We make the same assumption everywhere else in the codebase.

GeoWill: Done :)

PR #2582, opened 2020-02-14
Uprn council spatial join

GeoWill:
At the moment this mocks the ONSUD csvs, but with only the `lad` and `uprn` columns populated.
This is a proof of concept before creating a model that is just the look-up table and pointing [uk-geo-utils at it](https://github.com/DemocracyClub/UK-Polling-Stations/blob/434738b764e2ab0d0c04e3793cff33215885b808/polling_stations/settings/base.py#L258).

I've run `create_blacklist` and a few import scripts based on the database being set up like this and nothing has fallen over, and I can't see anywhere else `addressbase_onsud` is used.

TODO
* ~strip out the mocking of ONSUD from the create lookup command.~


GeoWill:
pushed these up while I investigate test failures to show general approach I'm taking.

GeoWill:
First stab at deployment updates here: https://github.com/DemocracyClub/polling_deploy/pull/201

polling_stations/apps/pollingstations/models.py
```diff
@@ -141,20 +141,26 @@ def get_absolute_url(self):
...
-            will not.
-
-            We might need to take a more sophisticated approach as we add more custom finders
-            that accept postcodes (e.g: a postcode format flag in the database).
-            At the moment I only have this one to work with.
-            """
-            finder.encoded_postcode = urllib.parse.quote(Postcode(postcode).with_space)
+            code = geocoder.get_code("lad")
+            if code.startswith("N"):
+                finder = self.get(pk="N07000001")
```
GeoWill: Thought I'd present pros and cons of a couple of alternative approaches here to stop slack swallowing them. The current option I just did as 'the simplest way to make it work™', an am not committed to. I think option 2 might be best, I think I'd want to write some tests so that if the councils were updated and no new CustomFinder object was created we'd know about it. 

#### Option 1
The approach in this commit is obviously hacky, but is easy to change from and leaves all the existing code to handle custom finders in place. I've checked it with `BT15 3JX` on the front end and the API, and as long as Local Authorities codes in NI keep being prefixed with an `N` is should keep working.

#### Option 2 Make a CustomFinder object for the NI Local Auths
Do [this](https://github.com/DemocracyClub/UK-Polling-Stations/blob/697437a02b87959116e45a3e41accd34a56184a4/polling_stations/apps/pollingstations/migrations/0013_customfinders.py) but `for gss in ["N09000001", "N09000002"....]` for the 11(?) Local Authorities. Pros: we can still introduce new CustomFinders easily. The only con I can think of is the need to keep on top of any changes to the codes and make the appropriate migrations. Given this is a rare occurrence it won't be much work, but it would be easy to forget to do. 

#### Option 3 Bin CustomFinder Model 
We could get rid of the CustomFinder completely and just insert code to check for a NI local authority and hard code the `context['custom']` [here](https://github.com/DemocracyClub/UK-Polling-Stations/blob/dcbdfa247d8a98ed217528a62de0ac77a1661bcc/polling_stations/apps/data_finder/views.py#L175). Which I think would work.
The pros of this are that we'd be reducing code since the only custom finder we handle is EONI. The cons are we might want to support others in the future, and I'm not certain it's quite as straightforward as I've described.

polling_stations/apps/addressbase/management/commands/create_uprn_council_lookup.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+
+    def handle(self, *args, **kwargs):
+
+        self.cursor = connection.cursor()  # Does this need to specify DEFAULT?
+        # Set where we'll write the join query to.
+        if kwargs["destination"]:
+            self.dst = Path(kwargs["destination"]).open("w+b")
+        else:
+            self.dst = Path("./uprn-to-councils.csv").open("w+b")
```
chris48s: Any particular reason we need to use binary mode here?

polling_stations/apps/addressbase/management/commands/import_uprn_council_lookup.py
```diff
@@ -0,0 +1,31 @@
...
+        if not self.path.exists():
+            raise FileNotFoundError(f"No csv found at {kwargs['path']}")
+
+        cursor = connection.cursor()
+        self.stdout.write("clearing existing data..")
+        cursor.execute("TRUNCATE TABLE %s;" % (self.table_name))
+
+        self.stdout.write("importing from CSV..")
+        with self.path.open("r") as f:
+            cursor.copy_from(f, self.table_name, sep=",")
```
chris48s: If/when you come to write a unit test for this, I suspect you might find you need to write this differently. I'll stick this here just in case you need to refer to it: https://github.com/DemocracyClub/UK-Polling-Stations/commit/35371c2df37cd953cf0eaef2a9fb77e4bb512eaf

polling_stations/apps/addressbase/models.py
```diff
@@ -19,9 +23,14 @@ class Address(AbstractAddress):
     objects = AddressManager()
 
 
-class Onsud(AbstractOnsud):
+class UprnToCouncil(models.Model):
     class Meta:
-        index_together = ("lad",)
+        indexes = [models.Index(fields=["lad",], name="lookup_lad_idx")]
```
chris48s: Nice :+1:

polling_stations/apps/addressbase/models.py
```diff
@@ -41,7 +50,7 @@ class Meta:
 
 def get_uprn_hash_table(council_ids):
     # get all the UPRNs in target local auth
-    # which exist in both Addressbase and ONSUD
+    # NB we miss ~25 over the country because lighthouses etc.
```
chris48s: ..and piers
Stupid piers

polling_stations/apps/addressbase/management/commands/create_uprn_council_lookup.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "-d", "--destination", help="Path to write csv to", default=None
+        )
+
+    def handle(self, *args, **kwargs):
+
+        self.cursor = connection.cursor()  # Does this need to specify DEFAULT?
```
chris48s: Nope - it uses by default by default (turns out the clue's in the name :smile:)
We only need to specify the connection if we want the logger DB

polling_stations/apps/councils/models.py
```diff
@@ -9,7 +9,7 @@ class Council(models.Model):
     website = models.URLField(blank=True)
     postcode = models.CharField(blank=True, null=True, max_length=100)
     address = models.TextField(blank=True, null=True)
-    area = models.MultiPolygonField(null=True, blank=True, geography=True, srid=4326)
+    area = models.MultiPolygonField(null=True, blank=True, srid=4326)
```
chris48s: I've got this weird nagging feeling there is code somewhere that expects this, but I can't find it for the life of me, so I reckon just crack on with it..

polling_stations/apps/pollingstations/models.py
chris48s: I reckon the way you've done this is fine. You can probably bin the CustomFinder model, but its not _harming_ anything to keep it there.

chris48s:
Oh, there are also a Couple of additional follow-up jobs relating to this work:

1. Have a look at the wiki: https://github.com/DemocracyClub/UK-Polling-Stations/wiki There's probably a mention or two of ONSUD in there that needs updating.
2. The code that makes @polling-bot-4000 raise notifications about updating the ONSUD like https://github.com/DemocracyClub/polling_deploy/issues/195 lives in https://github.com/DemocracyClub/ONS-Geography-Releases-Scraper/blob/master/scraper.py Not top priority, but we don't really need those any more so we can get rid of ONSUD from that (but we want to keep the others). Happy to run answer questions on that if necessary too.

polling_stations/apps/addressbase/management/commands/create_uprn_council_lookup.py
GeoWill: I feel like it stopped something breaking, but I can't remember what and it seems to work fine with just `"w"`.

polling_stations/apps/councils/models.py
GeoWill: I've had quite a good scour through the code base and can't see anything. It also doesn't make loads of sense that we should need the geography type for an application that applies to the UK. But I'll keep it in mind if anything weird crops up.

GeoWill:
>     1. Have a look at the wiki: https://github.com/DemocracyClub/UK-Polling-Stations/wiki There's probably a mention or two of ONSUD in there that needs updating.

I think I've got rid of these, I haven't written any new docs explaining how to use the new commands - I'll do this as part of the 'setting up a dev copy' section in the readme since it's automated in the build/deploy steps.

>     2.  The code that makes @polling-bot-4000 raise notifications about updating the ONSUD like [DemocracyClub/polling_deploy#195](https://github.com/DemocracyClub/polling_deploy/issues/195) lives in https://github.com/DemocracyClub/ONS-Geography-Releases-Scraper/blob/master/scraper.py Not top priority, but we don't really need those any more so we can get rid of ONSUD from that (but we want to keep the others). Happy to run answer questions on that if necessary too.

PR here: https://github.com/DemocracyClub/ONS-Geography-Releases-Scraper/pull/2

polling_stations/apps/pollingstations/models.py
GeoWill: I'm nervous of getting into the weeds with that, and it should probably be a separate PR anyway, so will leave it as is.

PR #2616, opened 2020-02-25
Dashboard marker clusters

GeoWill:
Need to think about whether this is useful. At the moment it indiscriminately clusters everything. I ideally it would only cluster markers of the same type (i.e. by pollingstation group),  and I think the threshold needs to be reduced. 
Anyway parking this here for now.

### DIFF
--- a/polling_stations/settings/static_files.py
+++ b/polling_stations/settings/static_files.py
@@ -6,22 +6,43 @@ PIPELINE = get_pipeline_settings(
     extra_js=[],
 )
 
-PIPELINE["STYLESHEETS"]["map"] = {
-    "source_filenames": [
-        "leaflet/dist/leaflet.css",
-        "custom_css/map.css",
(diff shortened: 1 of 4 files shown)


polling_stations/settings/static_files.py
```diff
@@ -6,22 +6,43 @@
...
+                "leaflet-extra-markers/dist/css/leaflet.extra-markers.min.css",
+            ],
+            "output_filename": "css/map.css",
+        },
+        "dashboard": {
+            "source_filenames": [
+                "leaflet.markercluster/dist/MarkerCluster.css",
+                "leaflet.markercluster/dist/MarkerCluster.Default.css",
+            ],
+            "output_filename": "css/map.css",
```
chris48s: This bundle should be called `dashboard.css`

GeoWill: Ah yes. Thanks!

chris48s:
will leave it with you to tidy up + merge

Issue #1630, opened 2019-03-08
Keep track of new local auth workaround


chris48s:
We've put a number of workarounds in place to deal with the creation of new councils this April:

https://github.com/DemocracyClub/UK-Polling-Stations/pull/1468
https://github.com/DemocracyClub/UK-Polling-Stations/pull/1455
https://github.com/DemocracyClub/UK-Polling-Stations/pull/1593
https://github.com/DemocracyClub/UK-Polling-Stations/pull/1610
https://github.com/DemocracyClub/UK-Polling-Stations/pull/1628

Although _theoretically_ we can roll all of this back some time post-May I suspect it probably makes sense to live with these for some time because we probably need to go through the same process again next year with Northamptonshire and Buckinghamshire (and to a lesser extent The Glasgow and North Lanarkshire Boundaries Amendment Order 2018 ). AFAIK, once these particular changes are reflected in data, we should be able to set `NEW_COUNCILS = []` and `OLD_TO_NEW_MAP = {}` but leave all this code and everything should Just Work™.

I'm opening this issue so that we've got a place to keep track of the workarounds we've put in place so that when (if) we want to get rid of them (or possibly refactor them into something more permanent :grin: ) , we know what they all were.

PR #2619, opened 2020-02-25
remove local auth workarounds

chris48s:
Now that we're using the latest ONSPD and we've got rid of ONSUD completely, everything we rely on is using the new codes/boundaries for all these areas.

Refs #1630
Refs https://trello.com/c/4JIFou7P
Refs https://trello.com/c/piIbUuL2
Refs https://trello.com/c/OCFJAUBn

I'm leaving any code in place that uses `OLD_TO_NEW_MAP` and `NEW_COUNCILS` which refs the councils table or ONSPD as we may need this feature again for Buckinghamshire in May 2020 and Northamptonshire in 2021.

I'm removing any code which assumes that the ONSUD exists or could not be up-to-date on the basis that we'll apply modifications to the councils table then generate a uprntocouncil table which reflects those changes, so it will always be as up-to-date as the councils table.

### DIFF
--- a/polling_stations/apps/addressbase/helpers.py
+++ b/polling_stations/apps/addressbase/helpers.py
@@ -1,6 +1,5 @@
 import logging
 from collections import namedtuple
-from django.conf import settings
 from django.contrib.gis.geos import GEOSGeometry
 from django.db import connection
 from councils.models import Council
@@ -72,8 +71,6 @@ class EdgeCaseFixer:
             council_id = c.council_id
         else:
             council_id = address.council_id
-            if council_id in settings.OLD_TO_NEW_MAP:
-                council_id = settings.OLD_TO_NEW_MAP[council_id]
 
         if council_id != self.target_council_id:
             # treat addresses in other council areas as district not found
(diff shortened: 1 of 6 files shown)


chris48s:
Yep, so just to work an example though in more detail: If we don't have new local auth boundaries yet and we were to set say for example:

```py
OLD_TO_NEW_MAP = {
    # Buckinghamshire
    "E07000004": "E06000060",
    "E07000005": "E06000060",
    "E07000006": "E06000060",
    "E07000007": "E06000060",
}

NEW_COUNCILS = ["E06000060"]
```

then when we run the councils importer, that would use `OLD_TO_NEW_MAP` to generate a composite boundary here: https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/councils/management/commands/import_councils.py#L120-L129

Then when you run `create_uprn_council_lookup`, `import_uprn_council_lookup`, `create_blacklist` etc, all of that is now going to be based on the councils table instead of using the ONSUD. That allows us to remove all the stuff in this PR which is there to account for the ONSUD not using the new codes because now the "ONSUD" isn;t really the ONSUD - its a custom table we built based on AddressBase and the councils table (which does use the new boundary).

The stuff I've left in place either relates to building the councils table, or dealing with the ONSPD not being up-to-date.

PR #2621, opened 2020-02-27
Add ONSPD postcode centroid to dashboard.

GeoWill:
This might be useful when there's a split postcode. I haven't totally decide about this as we don't want the bar for throwing away a single property from a postcode (as opposed to the whole postcode) to be high (i.e. it's obviously really wrong). So I'm not sure this makes it an easier call, but it is an extra bit of information.


GeoWill:
Some examples of where I've found this helpful:

- Dacorum (E07000096)
  - [HP42RS](http://127.0.0.1:8000/dashboard/postcode/HP42RS/) - I didn't throw this out because the onspd centroid implies 100081110218 and 200001052307 are part of the postcode.

GeoWill:
moved to #2747

PR #2690, opened 2020-03-22
use rapidfuzz instead of fuzzywuzzy

maxbachmann:
FuzzyWuzzy is GPLv2 licensed which would force you to licence the whole project under GPLv2. I had the same problem on one of my projects and so I wrote [rapidfuzz](https://github.com/rhasspy/rapidfuzz) which is implementing the same algorithm but is based on a version of fuzzywuzzy that was MIT Licensed and is therefor MIT Licensed aswell, so it can be used in here without forcing a License change. As a nice bonus it is fully implemented in C++ and comes with a few Algorithmic improvements making it between 5 and 100 times faster than FuzzyWuzzy.

### DIFF
--- a/polling_stations/apps/data_collection/data_types.py
+++ b/polling_stations/apps/data_collection/data_types.py
@@ -8,7 +8,7 @@ from collections import namedtuple
 
 from django.conf import settings
 from django.forms import ValidationError
-from fuzzywuzzy import fuzz
+from rapidfuzz import fuzz
 from localflavor.gb.forms import GBPostcodeField
 
 from addressbase.models import Blacklist, get_uprn_hash_table
@@ -383,7 +383,7 @@ class AddressList:
                     self.logger.log_message(
                         logging.INFO,
                         "Correcting postcode based on UPRN and fuzzy match.\nInput Record:\n%s\nAddressbase record:\n%s\nMatch quality: %s\n",
-                        variable=(record, addressbase_record, match_quality),
+                        variable=(record, addressbase_record, round(match_quality)),
                     )
                     record["postcode"] = addressbase_record["postcode"]
                 elif match_quality <= 50:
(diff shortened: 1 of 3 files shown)


chris48s:
Hello.
I'd be interested in having a look at this, but we seem to be hitting an error trying to install the lib in CI.
https://travis-ci.org/github/DemocracyClub/UK-Polling-Stations/builds/665639726
Are you able to advise on this or reproduce?
If not, I'll try to reduce it to a minimal repro case.

maxbachmann:
Hm yes and no. I know there is some reason that it does not successfully build the wheel in travis (it works e.g. in github actions). However in general there should be a prebuild python wheel https://pypi.org/project/rapidfuzz/#files.

So in case you find out what makes it fail in travis I would be happy to fix it

chris48s:
Thanks for this.

For the benefit of other maintainers, I ran an import script and compared all the match quality values from fuzzywuzzy and rapidfuzz across an import script and they were all the same within a rounding error. I'm happy to consider this a drop-in replacement for our use. I'll also take a 

@maxbachmann
- In our case, the build failure was caused by an issue with tox using an out of date version of wheel or setuptools or something by default which was causing us to try and compile from source instead of installing a bdist.
- In reviewing this, I found a minor difference between the behaviour of fuzzywuzzy and rapidfuzz was that fuzzywuzzy returns a rounded value from `partial_ratio()` whereas rapidfuzz returns it unrounded, so I wanted to round the value for display purposes. This is not a big deal, but possibly worth considering as you continue your mission to submit a PR to every non-GPL licenced public repo using fuzzywuzzy ;)

Merging this with a couple of fixups.

maxbachmann:
@chris48s yes I was not quite sure whether rounding this value is a good idea. It gives you a more clean result but at the same time e.g. when comparing long strings with small differences they would all return the same results.

PR #2691, opened 2020-03-23
Text Edits

chris48s:
* Coronavirus-specific text
* Revert all May elections text edits

### DIFF
--- a/polling_stations/templates/base.html
+++ b/polling_stations/templates/base.html
@@ -12,6 +12,19 @@
     <p>This is a staging site.</p>
   </div>
 </div>
+{% else %}
+<div class="container">
+  <div class="card">
+    <h2>Coronavirus</h2>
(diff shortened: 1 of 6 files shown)


polling_stations/templates/base.html
```diff
@@ -12,6 +12,19 @@
     <p>This is a staging site.</p>
   </div>
 </div>
+{% else %}
+<div class="container">
+  <div class="card">
+    <h2>Coronavirus</h2>
+    <p><strong>May 2020 scheduled elections</strong>: The Government have announced
+      they plan to postpone the election in May until 2021. We will update
+      ths information on this site once the postponement has become law.
```
GeoWill: Typo: ths -> this. I'd probably drop 'on this site' too.

chris48s: cheers. I think it was actually supposed to be "the information on this site", but I've gone with "this information"

Need to make the same fix on WCIVF :D

PR #2703, opened 2020-04-03
Match polling stations to addressbase

GeoWill:

PR #2738, opened 2020-05-14
Improve uprn to council model

GeoWill:
Stub that might allow some tidying up once bigger things have happened in #2703

### DIFF
--- a/polling_stations/apps/addressbase/models.py
+++ b/polling_stations/apps/addressbase/models.py
@@ -23,7 +23,7 @@ class Address(AbstractAddress):
     objects = AddressManager()
 
 
-class UprnToCouncil(models.Model):
+class UprnDirectory(models.Model):
     class Meta:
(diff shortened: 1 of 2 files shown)


GeoWill:
Depends on https://github.com/DemocracyClub/uk-geo-utils/issues/15

symroe:
To update this:

1. DemocracyClub/uk-geo-utils#15 is now fixed, so that shouldn't be a blocker

2. It would be nice to consider https://github.com/DemocracyClub/UK-Polling-Stations/issues/2823 when doing this

chris48s:
> DemocracyClub/uk-geo-utils#15 is now fixed, so that shouldn't be a blocker

You will need to publish a release of uk-geo-utils to PyPI first - maintenance docs are at https://github.com/DemocracyClub/uk-geo-utils/wiki

PR #2743, opened 2020-05-19
Initial importer and model changes for EC API

symroe:
The API isn't live yet, but this is the rough shape of what we need to do, I think.

Still do to:

* [x] Update the frontend to support `tel` being a list
* [x] Support registration and electoral services contact details, where different. We might want to only show registration details at some points on the electoral timeline, TBC
* [x] Update the API. This has a knock on for the devs.DC API, and might need a version bump from `v1` to `v2` (or `v.1.1`?)

I've left this work until the API is actually live - this has only been tested using a sample file sent to me.


polling_stations/settings/constants/councils.py
```diff
@@ -2,6 +2,9 @@
 
 YVM_LA_URL = "https://www.yourvotematters.co.uk/_design/nested-content/results-page2/search-voting-locations-by-districtcode?queries_distcode_query="  # noqa
```
chris48s, line 3: I think you can probably just ditch `YVM_LA_URL` here. In fact I probably I should have done that ages ago.

polling_stations/apps/councils/management/commands/import_councils.py
```diff
@@ -18,6 +15,21 @@ def union_areas(a1, a2):
...
+    "AND",
+    "ANN",
+    "BFS",
+    "CCG",
+    "DRS",
+    "FMO",
+    "LBC",
+    "MEA",
+    "MUL",
+    "NMD",
```
chris48s, line 29: If the councils will have register codes attached to them in the DB, could you infer this based on the register codes on councils where `council_id__startswith="N09"` instead of hard-coding it?

polling_stations/apps/councils/management/commands/import_councils.py
```diff
@@ -83,97 +87,82 @@ def get_json(self, url):
...
-            council.postcode = contact_details.postcode
+        This is normally only an issue with councils covered by EONI, so if
+        we see one of them without a name, we assign a hardcoded name.
+
+        """
+        name = None
+        if council_data["official_name"]:
+            name = council_data["official_name"]
+        else:
+            if council_data["code"] in NIR_IDS:
```
chris48s, line 136: I'm inferring from this line (and other stuff in this PR) that Numiko/the Commission are using gov.uk register codes as they key for councils/areas. This is probably the most concerning thing here, especially given the potential upcoming changes to councils. GSS codes have their problems - its annoying that they change in response to minor boundary changes but at least when re-organisations happen we can get hold of the new GSS codes ahead of time. It would be useful if they were actually published ahead of time, but in lieu of that at least ONS Geography support are really helpful (and consistently have been for many years). Gov.uk register codes are conceptually nice but they're a black box, their release schedule is non-existent and their support are somewhere on the spectrum between useless and actively obstructive (did I forward you my last email exchange with them?). Even though they're a conceptually nicer identifier if we care about an organisation rather than a polygon, these are big problems and this sets up yourself/Numiko/the EC for an annoying interoperability/temp codes problem.

Just as a kind of status update on this
https://elections.democracyclub.org.uk/organisations/combined-authority/north-of-tyne/2018-11-02/
This CA has existed since 2018. All other CAs have a register code. This one doesn't. I've never been able to get an answer on whether this is because it isn't supposed to have one (why?) or if its an oversight. It still has a temp slug in EE.

https://elections.democracyclub.org.uk/organisations/local-authority/buckinghamshire/2020-04-01/
This UA was created in April 2020. We gave it a temp slug in EE because it didn't have a register code yet. I was unable to really get an answer on when it would be assigned one. This one actually does have a code now, so I'll raise an issue on EE to update it.

https://elections.democracyclub.org.uk/organisations/local-authority/north-northamptonshire/2020-05-07/
https://elections.democracyclub.org.uk/organisations/local-authority/west-northamptonshire/2020-05-07/
These UAs will come into existence in April 2021 (the start dates in EE are 2020 because they were supposed to have "shadow" elections in 2020). The register codes have not been published yet. We don't know when they will be and we don't know what they are yet (but we do already know the GSS codes).

Obviously its you that has to live with the consequences of this, not me, so I'll make my point and leave you to think about it. Its worth at least having a think about what to do about this stuff/how to do temp IDs anyway..

polling_stations/apps/councils/management/commands/import_councils.py
symroe: This would be a good idea for the longer term, but this is a temporary work around: At the moment the name field is NULLABLE, and in the live API all the NI names are NULL. I think this is because the display name != council name, but I'd rather the EC have control over the display name in all cases, so we're working on making the name field required. 

When / if it is, or when they have populated the name field in all cases, this code won't be needed.

Because of that, I'll just leave it for the moment!

symroe:
> Have you or Numiko/the EC given any thought to how (or indeed if we should) solve "the Banff and Buchan problem"?. This will rear its ugly head again for the Scottish Parl elections. I think we mostly don't care for our own apps but the EC will.

No! I'd forgotten about that one, so thanks for the reminder! The EC aren't doing any sort of resolving of location to contact details (they just have a flat list, keyed against council ID), so I expect some sort of "fix" like we had last time will be needed again.

PR #2961, opened 2020-12-18
Make CI check for missing migrations

GeoWill:
As suggested here: https://github.com/DemocracyClub/EveryElection/pull/930#issuecomment-747552394
Thanks @chris48s

### DIFF
--- a/polling_stations/settings/base.py
+++ b/polling_stations/settings/base.py
@@ -1,12 +1,11 @@
 import os, sys
-from os.path import join, abspath, dirname
 
 # PATH vars
 
-here = lambda *x: join(abspath(dirname(__file__)), *x)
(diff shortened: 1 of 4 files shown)


polling_stations/settings/base.py
```diff
@@ -301,3 +301,11 @@
 # importing test settings file if necessary (TODO chould be done better)
 if len(sys.argv) > 1 and sys.argv[1] in ["test", "harvest"]:
     from .testing import *  # noqa
+
+
+def is_ci():
+    return os.environ.get("CIRCLECI")
+
+
+if is_ci:
```
michaeljcollinsuk, line 310: Missing the method call `()` so I think this will always resolve to `True`.

GeoWill: Good catch, thanks :). Refactor makes sense too.

PR #2962, opened 2020-12-18
Add nation to council

### LINKED ISSUE #2823: Return UK 'territory' in postcode_location key
There are many cased where downstream users (via the devs.DC API) will want to know the territory* of a postcode / address.

For example, lots of election law is changing in Wales and Scotland with the new devolved powers they have.

If the `postcode_location` field on the API response contained a territory code, it would be easier for consumers to show content specific to the users country / territory.

I suggest a format like this:

```json
"postcode_location": {
        "type": "Feature",
        "properties": {
            "nation": "ENG"
        },
        "geometry": {
            "type": "Point",
            "coordinates": [
                -2.20260988,
                51.74391142000001
            ]
        }
    },
```
This would be returned in all responses where a location is known, regardless of if a polling station is known.

## Implementation

Implementing this in `uk-geo-utils` is fairly easy - ONSUD and ONDPD have a `rgn` column that returns a pseudo GSS code for the "region". This starts with one of `E`, `S`, `W`, `N` like a normal GSS code. I've spoked an implementation of this here: https://github.com/DemocracyClub/uk-geo-utils/pull/17

The tricky part comes in this project, where we use `UprnToCouncil` as the `ONSUD_MODEL`. This is tricky for 2 reasons:

1. The table doesn't contain the `rgn` column, so we'd have to update the addressbase cleaning script to add it
2. It might have been possible to use the `lad` column for this when the value was a GSS code, but since #2818 the codes are now GDS register codes.

This might all change with #2738 of course, so we should consider this feature in that PR.

* is there a better term? One of England, Wales, Scotland or Northern Ireland - this is the term used by EE FWIW

GeoWill:
Closes #2823

Might be a bit ott, but if  `identifiers` can have anything in it then probably worth being fussy.


polling_stations/apps/councils/models.py
```diff
@@ -34,3 +36,45 @@ def __str__(self):
...
+            """
+            ^        # Start of string
+            (        # Entity Codes:
+                E06   # Unitary Authorities (England)
+              | E07   # Non-metropolitan Districts (England)
+              | E08   # Metropolitan Districts (England)
+              | E09   # London Boroughs (England)
+              | N09   # Local Government Districts (Northern Ireland)
+              | S12   # Council Areas (Scotland)
+              | W06   # Unitary Authorities (Wales)
```
GeoWill, line 64: Do we know if there are any other types of entities that run elections?
@pmk01 @symroe

pmk01: Nope, that's it!

symroe (approved):

symroe:
I think this is a decent start - if we find it's not right to have it as a property we can always move this logic to the council importer and set a field on the model. At the moment this is unlikely to be doing a lot of work for any council, so I think it's fine in terms of performance. :+1:

PR #2968, opened 2021-01-11
WiP: Factories

GeoWill:
This is an initial pass at adding some factories and writing a couple of tests with them. 

My next step is to see if I can come up with some pattern for building out all the different scenarios we want to cover in the data_finder module which is one of the two apps where fixture management is complicated.

### DIFF
--- a/polling_stations/apps/addressbase/tests/factories.py
+++ b/polling_stations/apps/addressbase/tests/factories.py
@@ -1,0 +1,18 @@
+import factory
+
+from addressbase.models import Address, UprnToCouncil
+
+
+class AddressFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = Address
+
+    uprn = factory.Sequence(lambda n: f"{n}".zfill(9))
+    addressbase_postal = "D"
+
+
+class UprnToCouncilFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = UprnToCouncil
+
+    uprn = factory.SubFactory(AddressFactory)
--- a/polling_stations/apps/addressbase/tests/test_factories.py
+++ b/polling_stations/apps/addressbase/tests/test_factories.py
@@ -1,0 +1,17 @@
+from django.test import TestCase
+
+from addressbase.models import Address
+from addressbase.tests.factories import AddressFactory, UprnToCouncilFactory
+
+
+class TestAddressFactory(TestCase):
+    def test_address_factory(self):
+        address = AddressFactory()
+        self.assertEqual(len(address.uprn), 9)
+        self.assertEqual(address.addressbase_postal, "D")
+
+
+class TestUprnToCouncilFactory(TestCase):
+    def test_uprn_to_council_factory(self):
+        uprn_to_council = UprnToCouncilFactory()
+        self.assertIsInstance(uprn_to_council.uprn, Address)
--- a/polling_stations/apps/councils/tests/factories.py
+++ b/polling_stations/apps/councils/tests/factories.py
@@ -1,0 +1,13 @@
+import itertools
+import string
+
+import factory
+
+from councils.models import Council
+
+
+class CouncilFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = Council
+
+    council_id = factory.Iterator(itertools.product(string.ascii_uppercase, repeat=3))
--- a/polling_stations/apps/councils/tests/test_factories.py
+++ b/polling_stations/apps/councils/tests/test_factories.py
@@ -1,0 +1,9 @@
+from django.test import TestCase
+
+from councils.tests.factories import CouncilFactory
+
+
+class TestCouncilFactory(TestCase):
+    def test_council_factory(self):
+        council = CouncilFactory()
+        self.assertEqual(len(council.council_id), 3)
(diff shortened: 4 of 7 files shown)


michaeljcollinsuk:
Might already be aware of this, but in case not, worth checking out Faker to help to build out your factories with fake data. They have a lot of in built providers that can cover generating a lot of fake data for models https://faker.readthedocs.io/en/master/providers.html

GeoWill:
At the moment we handle our fixtures by defining them in json files. This is slowly getting more manageable by splitting the different scenarios into separate files, and with a bit of a push we could create standard mapping of postcode -> scenario we want to test, and use these across the project. 
What I dislike about this is that altering the json files is fiddly, and it can be hard to build a mental model of what is being tested without spending a lot of time studying the json - and it can diverge from what the test implies. 

This spike goes to the other extreme and abstracts the creation of test features into a `PostcodeBuilder` class. My concern with this is that if you want to make any changes you need to parse and edit this new class - which might be harder than just fiddling with the json. I'd really welcome some other opinions. 

A halfway house would be just to use the factory classes to build the different scenarios where we need them - this could be dry-ed out by doing something like this: https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/elections/tests/base_tests.py#L26

polling_stations/apps/data_finder/tests/test_postcode_builder.py
```diff
@@ -0,0 +1,65 @@
...
+class TestPostcodeTestCase(TestCase):
+    def test_construction(self):
+        postcode = PostcodeBuilder()
+        self.assertIsInstance(postcode.postcode, str)
+        self.assertEqual(len(postcode.councils), 1)
+        self.assertIsInstance(postcode.councils[0], Council)
+        self.assertEqual(postcode.councils[0].council_id[0], "E")
+
+    def test_construction_with_multiple_councils(self):
+        postcode = PostcodeBuilder(multiple_councils=True, nations="W")
```
symroe, line 18: Should this be a list or set? I think it's currently working because a `str` is an iterable, but it looks odd to have a string here and a tuple later

GeoWill: I'd settled on passing a tuple as the argument - just missed it in the test. Realise I also miss the trailing comma here: https://github.com/DemocracyClub/UK-Polling-Stations/blob/eea82d38edfcf849e447418b70a300f282681d03/polling_stations/apps/data_finder/tests/utils.py#L44
Perhaps there's a reason not to use a tuple though?
A TODO if we go with this is to add some validation - eg `(W,S)` shouldn't be allowed.

symroe: Would this also work in the builder pattern, with a `in_wales()` (etc) method?

GeoWill: Hmm, yeah that would work, I guess what I'm not sure about is calling a `in_country` method every time, but actually maybe being explicit is quite good. :thinking:

polling_stations/apps/addressbase/tests/test_address.py
```diff
@@ -0,0 +1,47 @@
...
+            pk=self.uprn_without_station.uprn.uprn
+        )
+        self.address_with_station = Address.objects.get(
+            pk=self.uprn_with_station.uprn.uprn
+        )
+        self.addresses = [self.address_no_station, self.address_with_station]
+
+    def test_council(self):
+        for address in self.addresses:
+            self.assertEqual(address.council, self.council)
```
michaeljcollinsuk, line 32: You could use `subTest` here (and test below) as then if one of the addresses failed it would be clearer which one.

michaeljcollinsuk:
I do like the builder idea, but it does add a further level of abstraction that you need to understand in order to properly understand the tests. Although once you do have that knowledge of the builder, it does mean that your tests are very concise. Although personally I don't mind the tests to be more explicit at the expense of DRY-ness, so that I can understand what is happening inline with the actual tests.

Another thought is that you may be able to achieve some of what the builder is doing via factory boy [post generation hooks](https://factoryboy.readthedocs.io/en/stable/reference.html#post-generation-hooks)? For example, you could have a different `CouncilFactory` class for different scenarios e.g. `CouncilWithAssignedAddressFactory` which utilises post-generation methods to create the PollingStations, similar to what is happening in your `with_assigned_addresses` method on the builder.

michaeljcollinsuk:
Pseudo code example of sort of thing that I meant in previous comment:
```
# new class inheriting from base CouncilFactory
class CouncilWithAssignedAddress(CouncilFactory):

    @factory.post_generation
    def create_address(self, create, extracted, **kwargs):
        polling_station = PollingStationFactory(council=self)
        # it may be that calculating the 'overlap' can be on the AddressFactory rather than here
        overlap = self.geometry.intersection(council.area)
        address = AddressFactory(
            location=get_unique_point_in_extent(overlap.extent), 
            postcode=kwargs['postcode'],  # you can pass the postcode in the kwargs when using the factory
        )
        UprnToCouncilFactory(
            uprn=address, 
            lad=self.council_id, 
            polling_station_id=polling_station_id,
        )

# then factory could be used in the test (or setup)...

def test_redirect_with_assigned_addresses(self):
    # set or could use faker to create a random one
    postcode = 'TE1 1ST'

    # create the council and pass in postcode for use in post get method
    CouncilWithAssignedAddress(create_address__postcode=postcode)

    response = self.client.post(
            reverse("home") + "?utm_source=foo&something=other",
            {"postcode": postcode},
            follow=False,
    )
    self.assertEqual(302, response.status_code)
    self.assertRegex(response["location"], "/address/[0-9]{9}/")
```

PR #2973, opened 2021-01-20
Create dev env without addressbase

GeoWill:


chris48s:
Does this give you 2 management commands in the application both called `import_cleaned_addresses`  (one in the geo_utils app and one in the addressbase app)? I'll leave it with you to make a decision about which way to go with it, but it is worth understanding the rules that define how django resolves the conflict if your combination of installed apps means you have >1 commands with the same name: https://docs.djangoproject.com/en/2.2/ref/settings/#installed-apps (essentially the order of `INSTALLED_APPS` is significant)

polling_stations/apps/addressbase/management/commands/import_cleaned_addresses.py
```diff
@@ -0,0 +1,31 @@
+from uk_geo_utils.helpers import get_onsud_model
+from uk_geo_utils.management.commands import import_cleaned_addresses
+import os
+from django.db import connection
+
+
+class Command(import_cleaned_addresses.Command):
+    def import_addressbase(self):
```
symroe, line 8: I understand the parent command doesn't do this, but given that this deletes a load of data, do we want to add a warning / confirmation to this command? Or does it not really matter?

GeoWill:
Thanks Chris. AIUI there's only one command available through `manage.py` now, as I don't think there's any namespacing - but which one you get is defined by the installed apps order as you point out. I imagined sending the patch upstream, (as I don't think it will break anything in uk-geo-utils), and then dropping the UK-Polling-Stations version. But wanted something that would just work for now.

polling_stations/apps/addressbase/management/commands/import_cleaned_addresses.py
GeoWill: Maybe - it should be in uk-geo-utils too then. I don't think it really matters though, because it won't run unless you have a file to point it at, and if you have a file to point it at then I guess that's what you want. Adding some help text to the command that makes it clear that it's replacing, rather than adding, data would be probably be good.

PR #2988, opened 2021-01-27
Remove requests for donation

### LINKED ISSUE #2986: Remove donate button in footer
We're not asking for money at the moment (at least not small donations), can the link in the footer be removed for the moment?

GeoWill:
closes #2986


polling_stations/assets/custom_css/style.scss
```diff
@@ -45,20 +45,3 @@ select[size], select[multiple] {
   background-color: $dc-green;
   color: #FFF;
 }
-
-
```
VirginiaDooley, line 49: I don't actually see these classed being used in the file where you've deleted the donation link, which makes me wonder whether they are in use somewhere else  (and should be kept) if at all (and safe to delete)?

VirginiaDooley (commented):
It looks like the deleted classes are still in this [file](https://github.com/DemocracyClub/UK-Polling-Stations/commit/1beb17f4d7e57708e678b352b137552df68bacb1#diff-cb058d08096fa55859b7bc06774ece60e4ac85922e62828b07d26c9b578ac9e1). Did you want to delete the card as well? If not, the classes should stay.

GeoWill:
> It looks like the deleted classes are still in this file. Did you want to delete the card as well? If not, the classes should stay.

Do you mean `polling_stations/templates/fragments/donate.html`? if so it should be deleted as part of the commit (https://github.com/DemocracyClub/UK-Polling-Stations/pull/2988/files#diff-cb058d08096fa55859b7bc06774ece60e4ac85922e62828b07d26c9b578ac9e1). I think GitHub doesn't unfurl the diff when a whole file goes, so might not be obvious when you're reviewing.

polling_stations/assets/custom_css/style.scss
GeoWill: `grep -r 'donate_card'` in the root of the repository doesn't show anything, so should be safe to go. I think line 1 of  polling_stations/templates/fragments/donate.html was the only place they were used.

VirginiaDooley (approved):
LGTM 🚀

PR #2991, opened 2021-02-02
Add council geography model

GeoWill:


polling_stations/apps/data_importers/base_importers.py
```diff
@@ -238,17 +238,11 @@ def check_station_point(self, station_record):
...
-                        "Polling station %s is in %s (%s) but target council is %s (%s) - manual check recommended\n",
-                        variable=(
-                            station_record["internal_council_id"],
-                            council.name,
-                            council.council_id,
-                            self.council.name,
-                            self.council.council_id,
-                        ),
+                        f"Polling station {station_record['internal_council_id']} is in {council.name} ({council.council_id}) "
+                        f"but target council is {self.council.name} ({self.council.council_id}) - manual check recommended\n",
```
GeoWill, line 245: Is there any reason for these to not be f-strings?

polling_stations/apps/api/councils.py
```diff
@@ -52,38 +72,19 @@ class Meta:
...
-            "email",
-            "phone",
-            "website",
-            "postcode",
-            "address",
-            "electoral_services_contacts",
-            "registration_contacts",
-            "identifiers",
-        )
+        fields = COUNCIL_FIELDS
```
GeoWill, line 87: I think this is right... I copied the pattern from EE as best I could.

polling_stations/apps/data_importers/data_quality_report.py
```diff
@@ -247,14 +249,15 @@ def get_districts_containing_more_stations(self):
 class AddressReport:
     def __init__(self, council_id):
         self.council_id = council_id
+        self.gss_code = Council.objects.get(pk=council_id).geography.gss
```
GeoWill, line 252: I guess `gss_code` could be a property on the council model, but part of the point of this pr is to make it a seperate concept - so I'm inclined not to.

polling_stations/apps/data_importers/base_importers.py
chris48s: Drive-by: The main reason why a lot of the code in this codebase uses `%` or `.format()` instead of f-strings is because a lot of it was written before python 3.6 came out (or before we were deploying on it), not because f-strings are bad. AFAIK you're on python 3.6 in prod (I think) so no reason not to start using them.

GeoWill:
A non obvious thing this PR cements in is the need to use a register code in an import script, and therefore in the naming of our data folders on s3. The file uploader does this, so as long as we do that it should be fine. If for some reason one wanted to import old data it's necessary to change the bucket name to whatever the reg code is. I think this makes sense, though it could make your historical imports a bit funky as the list of addresses a council sends us is (hopefully) a function of gss code (i.e. the councils boundary when they produce the list) rather than reg code.

PR #3005, opened 2021-02-10
Pre election copy changes

GeoWill:
This takes care of the tasks mentioned [here](https://github.com/DemocracyClub/UK-Polling-Stations/wiki/Maintenance-tasks#pre-election), and also adds a card about offering working for the council.
The wording of the latter probably needs some work.

I've deployed this to staging, in case anyone wants a look.

### DIFF
--- a/polling_stations/apps/data_finder/helpers/every_election.py
+++ b/polling_stations/apps/data_finder/helpers/every_election.py
@@ -204,3 +204,11 @@ class EveryElectionWrapper:
 
     def get_id_pilot_info(self):
         return self.get_metadata_by_key("2019-05-02-id-pilot")
+
+    @property
+    def multiple_elections(self):
+        if self.has_election and self.request_success:
+            uncancelled_ballots = [b for b in self.ballots if not b["cancelled"]]
+            return len(uncancelled_ballots) > 1
+        else:
+            return False
--- a/polling_stations/templates/base_full.html
+++ b/polling_stations/templates/base_full.html
@@ -10,7 +10,7 @@
     {% endblock base_title %}
   </title>
 
-  <meta name="description" content="{% block description %}Find out where to vote on 6 May 2021{% endblock %}">
+  <meta name="description" content="{% block description %}Find out where to vote{% if NEXT_CHARISMATIC_ELECTION_DATE %} on {{ NEXT_CHARISMATIC_ELECTION_DATE|date:'F j Y' }}{% endif %}{% endblock %}">
   <meta name="keywords" content="{% block keywords %}Polling stations,Election,General Election,Local Election{% endblock %}">
   <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
@@ -23,7 +23,7 @@
   <meta name="twitter:site" content="@democlub">
   <meta property="og:image" content="{{ CANONICAL_URL }}{% block og_image %}{% static "dc_theme/images/logo.png" %}{% endblock og_image %}">
   <meta property="og:title" content="{% block og_title %}{{ site_title }}{% endblock og_title %}">
-  <meta property="og:description" content="{% block og_description %}Find out where to vote on 6 May 2021{% endblock og_description %}">
+  <meta property="og:description" content="{% block og_description %}Find out where to vote{% if NEXT_CHARISMATIC_ELECTION_DATE %} on {{ NEXT_CHARISMATIC_ELECTION_DATE|date:'F j Y' }}{% endif %}{% endblock og_description %}">
   {% if noindex %}
   <meta name="robots" content="noindex">
   {% endif %}
--- a/polling_stations/templates/base_full.html
+++ b/polling_stations/templates/base_full.html
@@ -10,7 +10,7 @@
     {% endblock base_title %}
   </title>
 
-  <meta name="description" content="{% block description %}Find out where to vote{% endblock %}">
+  <meta name="description" content="{% block description %}Find out where to vote on 6 May 2021{% endblock %}">
   <meta name="keywords" content="{% block keywords %}Polling stations,Election,General Election,Local Election{% endblock %}">
   <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
@@ -23,7 +23,7 @@
   <meta name="twitter:site" content="@democlub">
   <meta property="og:image" content="{{ CANONICAL_URL }}{% block og_image %}{% static "dc_theme/images/logo.png" %}{% endblock og_image %}">
   <meta property="og:title" content="{% block og_title %}{{ site_title }}{% endblock og_title %}">
-  <meta property="og:description" content="{% block og_description %}Find out where to vote{% endblock og_description %}">
+  <meta property="og:description" content="{% block og_description %}Find out where to vote on 6 May 2021{% endblock og_description %}">
   {% if noindex %}
   <meta name="robots" content="noindex">
   {% endif %}
--- a/polling_stations/templates/fragments/help_your_council.html
+++ b/polling_stations/templates/fragments/help_your_council.html
@@ -1,0 +1,16 @@
+{% load i18n %}
+<div class="card">
+    {% blocktrans with council.electoral_services_phone_numbers.0 as council_phone and council.name as council_name and council.electoral_services_email as council_electoral_services_email %}
+        <h3>Your election (might) need you!</h3>
+        <p>
+          A lot of people are needed to make an election happen.
+          Your council may be hiring polling station and counting staff for the upcoming elections.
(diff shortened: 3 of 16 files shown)


polling_stations/templates/fragments/help_your_council.html
```diff
@@ -0,0 +1,16 @@
...
+        <h3>Your election (might) need you!</h3>
+        <p>
+            A lot of people are needed to make an election happen.
+	    You can contact your council electoral services team to see if they need help.
+        </p>
+        <p>
+            You can <strong>
+		    <a href="{{ council_electoral_services_email }}">email</a>
+	    </strong> {{ council_name }} electoral services on or call them on
+	    <strong><a href="tel:{{ council_phone }}">{{ council_phone }}</a></strong>.
```
GeoWill, line 13: @pmk01 do you have any thoughts on making this more explicit? I didn't know exactly what we're asking for. I also worry about pushing folk to ringing ES teams who may have enough on their plates already.

polling_stations/templates/home.html
```diff
@@ -32,7 +32,7 @@ <h1>{% trans "Find your polling station" %}</h1>
 
 {% if show_polls_open %}
 <div class="card centered-card">
-  {# <h2>Put a Title in Here</h2> #}
+  { <h2>6 May 2021 Elections</h2> }
```
GeoWill, line 35: Is this the best collective title for these elections?

GeoWill:
This doesn't fit inline, but is connected to this PR. Once we start getting data we'll begin showing the [explanations fragment](https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/templates/election_explainers.html). However somewhere like GL10 3DE ([who link](https://whocanivotefor.co.uk/elections/GL10%203DE/)) will be having three elections (two of which are local) but will only show an explanation about the PCCs - so there'll be nothing indicating how many elections they have. I get the feeling making this work in Who hasn't been totally straight forward, so would appreciate any input. 
My initial thought is that showing election information is useful, but only if it's complete. An omission sort of implies something isn't happening, so if getting explanations into EE is too big a job then I'd drop them completely.

polling_stations/templates/fragments/help_your_council.html
```diff
@@ -0,0 +1,16 @@
...
+    {% blocktrans with council.electoral_services_phone_numbers.0 as council_phone and council.name as council_name and council.electoral_services_email as council_electoral_services_email %}
+        <h3>Your election (might) need you!</h3>
+        <p>
+            A lot of people are needed to make an election happen.
+	    You can contact your council electoral services team to see if they need help.
+        </p>
+        <p>
+            You can <strong>
+		    <a href="{{ council_electoral_services_email }}">email</a>
+	    </strong> {{ council_name }} electoral services on or call them on
```
michaeljcollinsuk, line 12: Think there is a rogue `on` here (the first one) that should be removed here? So that it reads "You can email {{ council_name }} electoral services or call them on..."

GeoWill: Yes! Thanks :)

polling_stations/templates/fragments/help_your_council.html
pmk01: I think we could make it more explicit, yes - 70+% of electoral admins say they don't have enough staff or resources for these elections, and every council site I have looked at is advertising for poll workers. To quote Martyn H when I put the suggestion to him: "Yes I think that would be a great idea- perhaps mention polling and counting staff, with multi day counts we’re all after bodies."

So I would suggest something like:

"A lot of people are needed to make an election happen. Your council may be hiring polling station and counting staff for the upcoming elections. Contact your council electoral services team to see if they need help."

I think any phone calls we create this way will be a drop in the ocean compared to their usual numbers, and offers of help will probably be welcome!

pmk01: Or a link to the webpage, as you suggested on the call!

GeoWill: The websites we have are just council websites - so not actually a useful link. As you pointed out volume isn't going to be overwhelming so will leave the contact details in.

polling_stations/apps/data_finder/helpers/every_election.py
```diff
@@ -204,3 +204,12 @@ def get_metadata_by_key(self, key):
 
     def get_id_pilot_info(self):
         return self.get_metadata_by_key("2019-05-02-id-pilot")
+
+    @property
+    def multiple_elections(self):
+        if self.has_election and self.request_success:
+            uncancelled_ballots = [b for b in self.ballots if not b["cancelled"]]
+            if len(uncancelled_ballots) > 1:
```
michaeljcollinsuk, line 212: Really small thing, but I would return the result of the comparison statement here:
```
return len(uncancelled_ballots) > 1
```
Means you lose the nested if statement and also ensures that the method always explicitly returns either `True` or `False` (currently you implicitly also return `None`)

GeoWill: Sounds good - I had originally written it without `if self.has_election and self.request_success:` so there was no implicit `None`. I'd missed it with the extra `if` - that's a sensible way of getting rid of it.

polling_stations/templates/base_full.html
```diff
@@ -23,7 +23,7 @@
   <meta name="twitter:site" content="@democlub">
   <meta property="og:image" content="{{ CANONICAL_URL }}{% block og_image %}{% static "dc_theme/images/logo.png" %}{% endblock og_image %}">
   <meta property="og:title" content="{% block og_title %}{{ site_title }}{% endblock og_title %}">
-  <meta property="og:description" content="{% block og_description %}Find out where to vote{% endblock og_description %}">
+  <meta property="og:description" content="{% block og_description %}Find out where to vote on 6 May 2021{% endblock og_description %}">
```
symroe, line 26: Can this be based off `NEXT_CHARISMATIC_ELECTION_DATE`?

e.g:

1. add to [`global_settings`](https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/pollingstations/context_processors.py#L8)
2. `{% if NEXT_CHARISMATIC_ELECTION_DATE %} on {{ NEXT_CHARISMATIC_ELECTION_DATE|date }}{% endif %}`

polling_stations/templates/fragments/polling_station_known.html
```diff
@@ -34,7 +34,7 @@ <h2>{% trans "Your polling station" %}</h2>
   </div>
 {% endif %}
 
-<p><strong>Polling stations are open from 7am to 10pm.</strong></p>
+<p><strong>Polling stations are open from 7am to 10pm on Thursday 6 May.</strong></p>
```
symroe, line 37: as per `NEXT_CHARISMATIC_ELECTION_DATE` above

polling_stations/templates/postcode_view.html
```diff
@@ -60,12 +60,15 @@
...
+    {% if multiple_elections %}
+       {% include "fragments/multiple_elections.html" %}
+    {% endif %}
 
     {% include "fragments/register_to_vote.html" %}
     {% include "fragments/you_dont_need_poll_card.html" %}
 
-    {% if we_know_where_you_should_vote and election_explainers and has_election and not error %}
-      {% include "election_explainers.html" %}
+    {% if council.electoral_services_phone_numbers and council.electoral_services_email %}
```
symroe, line 70: How does this work in Scotland and Northern Ireland?

symroe: Also, what's the system for taking this info off the site? Do we have a reminder? Or should we try to write code to turn it off?

GeoWill: Scotland and Northern Ireland are good points. I'll include `has_election` in the if. Also we shouldn't be showing it in NI ever, as we don't store council specific info for NI. I think it's OK for Scotland because we do have council specific electoral services info.

GeoWill: It's mentioned in [the wiki](https://github.com/DemocracyClub/UK-Polling-Stations/wiki/Maintenance-tasks#post-election) and in the `election playbook` google sheet. I think that's ok for now.

symroe: And for Scotland, are we sure that it's the local council who hire for polling stations? I assume so, but would be good to double check!

GeoWill: @pmk01 do you know the answer to this :point_up:?

pmk01: Pretty sure it's the council, or at any rate the info will be found on the council website eg.: https://www.highland.gov.uk/info/799/elections_and_voting/603/training_and_employment_at_elections

polling_stations/templates/base_full.html
GeoWill: Had a bash at an implementation in 005765f. As discussed on slack `NEXT_CHARISMATIC_ELECTION_DATE` is a string, so had to be cast to a date in the context processor.

PR #3013, opened 2021-02-15
Tamworth

smsmith97:
import script for Tamworth


GeoWill (changes requested):
The actual import script is great, we just need to squash the two commits together. Happy to pair on this to show you how to that :+1:

GeoWill (approved):

PR #3021, opened 2021-02-17
Warn on stations with same location, different postcode

GeoWill:
It's not uncommon to find two polling stations with the same lat/lon in
the council file, but different addresses - i.e. one has the wrong
lat/lon. However it's also common to have multiple stations at the same
location, e.g. different rooms in a school. In order to catch the
stations that have been given the wrong location in error, this check
compares the distance between the station to be added to the stationset
and all the stations added so far.

This check won't happen when the --nochecks flag is passed.

This commit also refactors the check_station_point method now there are
two checks, both of which require a location.

The qgis filter expression line allows the user to quickly identify the
stations in question by applying a filter to the polling stations layer.

### DIFF
--- a/polling_stations/apps/data_importers/base_importers.py
+++ b/polling_stations/apps/data_importers/base_importers.py
@@ -9,6 +9,7 @@ import os
 import tempfile
 import urllib.request
 
+import rtree
 from django.apps import apps
 from django.contrib.gis import geos
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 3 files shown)


polling_stations/apps/data_importers/tests/test_base_stations_importer.py
```diff
@@ -0,0 +1,78 @@
...
+
+from django.contrib.gis.geos import Point
+from django.test import TestCase
+
+from councils.tests.factories import CouncilFactory
+from data_importers.base_importers import BaseStationsImporter
+from data_importers.data_types import StationSet
+
+
+class MockLogger:
```
michaeljcollinsuk, line 12: I think using this to then assert against in your tests is a really nice solution. 

My only other suggestion would be to use a unitest `MagicMock` object instead, which would save you from needing your own class with these methods. You could then use the [autospeccing](https://docs.python.org/3/library/unittest.mock.html#autospeccing) with your actual logger class, and then you could use some of the convenience methods like [`assert_called_once_with`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with) to check the logger is called correctly.

But like I say, think the way you are doing works fine and is really clear when reading the tests. So this is more just me suggesting a possible other way rather than saying its better!

polling_stations/apps/data_importers/tests/test_base_stations_importer.py
```diff
@@ -0,0 +1,78 @@
...
+        self.base_stations_importer.add_polling_station(ps1)
+        self.base_stations_importer.check_duplicate_location(ps2)
+        self.assertListEqual(
+            self.base_stations_importer.logger.logs,
+            [
+                "Polling stations 'bar' and 'foo' are at approximately the same location, "
+                "but have different postcodes:\nqgis filter exp: \"internal_council_id\" IN ('02','01')"
+            ],
+        )
+        if os.environ.get("CIRCLECI"):
```
michaeljcollinsuk, line 69: Its not clear to me why the third assert only gets called in CI, but I assume there is a reason for it

polling_stations/apps/data_importers/base_importers.py
```diff
@@ -241,22 +242,79 @@ def get_station_hash(self, station):
...
+                    record_geom.y + threshold,
+                ),
+                1,
+            )
+        )
+        if nearest_ids:
+            for i in nearest_ids:
+                station = stations_with_location_and_different_postcode[i.id]
+
+                def get_name(address):
```
michaeljcollinsuk, line 290: One issue with defining this function here is that if you wanted to write a test for the logic of extracting the name it would be quite tricky

polling_stations/apps/data_importers/base_importers.py
```diff
@@ -241,22 +242,79 @@ def get_station_hash(self, station):
...
+                (
+                    record_geom.x - threshold,
+                    record_geom.y - threshold,
+                    record_geom.x + threshold,
+                    record_geom.y + threshold,
+                ),
+                1,
+            )
+        )
+        if nearest_ids:
```
michaeljcollinsuk, line 286: I prefer to flip the conditional and return early in this circumstance, to avoid nesting:
```
if not nearest_ids:
    return

for I in nearest_ids:
...
```

polling_stations/apps/data_importers/tests/test_base_stations_importer.py
GeoWill: It's because my local environment will fail it, because of how my spatial libraries are set up! This does mean I'm not totally confident I'm going to get all the warnings I want, but there are other ways to check for the same thing ,so it's not the end of the world.

polling_stations/apps/data_importers/tests/test_base_stations_importer.py
GeoWill: OK, glad it makes sense - that was my concern. I'll have a look at the `MagicMock` for other times.

polling_stations/apps/data_importers/base_importers.py
GeoWill: Yep, I did it like this because you're not allowed a backslash in the `{}` bit of an `f-string`. Guess I could've used some other formatting method, but thought this was acceptable.

polling_stations/apps/data_importers/base_importers.py
GeoWill: :D I was thinking of your trick with this when I wrote line 255, but had forgotten by the time I got here.

PR #3036, opened 2021-02-22
import script for worthing (closes #3011)

### LINKED ISSUE #3011: Import WOT-Worthing Borough Council
EMS: Xpress DC
Files:
- `WOT/2021-02-15T11:21:49.443491/WBCDemocracy_Club__06May2021.tsv`

smsmith97:


polling_stations/apps/data_importers/management/commands/import_worthing.py
```diff
@@ -1,29 +1,31 @@
...
+            record = record._replace(polling_place_easting="514729")
+            record = record._replace(polling_place_northing="102975")
+
+        # Improved accuracy of station locations
+        if (
+            record.polling_place_id == "2143"
+        ):  # St. Richard's Church, Collingwood Road, Worthing
+            record = record._replace(polling_place_easting="511512")
+            record = record._replace(polling_place_northing="103575")
+
```
GeoWill, line 24: I can see why you made this change, and it definitely shows good attention to detail! However I'd be reluctant to include it without checking with the council - in this case if you look on a satellite image there are two buildings and the point provided by the council is on the one that doesn't appear in osm. We shouldn't be the ones figuring it out, so should ask or leave it. On a related note from a user perspective, by the time they've got that close there should be sufficient signage for them to figure it out.

polling_stations/apps/data_importers/management/commands/import_worthing.py
```diff
@@ -1,29 +1,31 @@
...
+        if (
+            record.polling_place_id == "2155"
+        ):  # Palatine Park Football Centre, Palatine Road, Worthing
+            record = record._replace(polling_place_easting="511285")
+            record = record._replace(polling_place_northing="103912")
+
+        return super().station_record_to_dict(record)
```
GeoWill, line 31: Here I think you're more likely to be right about the correction, but I'd still leave it out for reasons out lined above. The reason for leaving the correction to assembly rooms in is a. it's from a previous script which references a user report; and b. it's a much bigger change, so will impact on users/is more 'obviously wrong.

GeoWill (changes requested):
Few minor changes but looking good :)

GeoWill (approved):

PR #3073, opened 2021-03-02
Handle addresses that aren't in a council

GeoWill:
These are mostly piers, and there's only 20 odd of them.

These changes mean that we don't consider them when looking if a
postcode is split, and if all the other addresses in that postcode have
the same council we assume the pier (or lighthouse) is also in that
council.

### DIFF
--- a/polling_stations/apps/addressbase/management/commands/import_uprn_council_lookup.py
+++ b/polling_stations/apps/addressbase/management/commands/import_uprn_council_lookup.py
@@ -2,6 +2,8 @@ from django.db import connection
 from django.core.management.base import BaseCommand
 from pathlib import Path
 
+from addressbase.models import Address, UprnToCouncil
+
 
 class Command(BaseCommand):
     """
@@ -29,3 +31,18 @@ class Command(BaseCommand):
             cursor.copy_from(f, self.table_name, sep=",")
 
         self.stdout.write("...done")
+        self.stdout.write(
+            "Looking for addresses outside council areas... (aka the pier check...)"
+        )
+        for address in Address.objects.filter(uprntocouncil__isnull=True):
+            council = address.get_council_from_others_in_postcode()
+            if council:
+                self.stdout.write(
+                    f"Creating UprnToCouncil record for {address.uprn} with gss {council.geography.gss}"
+                )
+                UprnToCouncil.objects.create(
+                    uprn=address, lad=council.geography.gss, polling_station_id=""
+                )
+            else:
+                self.stdout.write(f"Council ambiguous for {address.uprn}, deleting")
+                address.delete()
--- a/polling_stations/apps/addressbase/models.py
+++ b/polling_stations/apps/addressbase/models.py
@@ -9,6 +9,17 @@ from pollingstations.models import PollingStation
 
 
 class Address(AbstractAddress):
+    def get_council_from_others_in_postcode(self):
+        others = (
+            Address.objects.filter(postcode=self.postcode)
+            .exclude(uprntocouncil__isnull=True)
+            .distinct("uprntocouncil__lad")
+        )
+        if len(others) == 1:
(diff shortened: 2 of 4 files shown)


chris48s:
aah we meet again. Its these ones, isn't it..

* CROMER PAVILION THEATRE, THE PIER, PROMENADE, CROMER
* C & A THOMASON LTD, CULLIVOE PIER, CULLIVOE, YELL, SHETLAND
* CLARKS BOATWORKS LTD, CROWN WHARF 1-5, CASTLETOWN, PORTLAND
* 3 ALBERT QUAY, FOWEY
* WORTHING PIER EVENTS LTD, WORTHING PIER, MARINE PARADE, WORTHING
* THE BOAT THAT ROCKS, 7 HAMM BEACH ROAD, PORTLAND
* THE PHAROS TRUST, LIGHT VESSEL 18, THE QUAY, HARWICH
* JOLLY ROGER, THE PIER, CLACTON-ON-SEA
* CROMER LIFEBOAT STATION, PROMENADE, CROMER
* THE CROFT HOUSE, GARTHS OF HAM, BRESSAY, SHETLAND
* THE CLACTON PIER CO LTD, THE PIER, CLACTON-ON-SEA
* DEAN & REDDYHOFF LTD, 6 HAMM BEACH ROAD, PORTLAND
* HALFPENNY PIER VISITORS CENTRE, THE QUAY, HARWICH
* PIER POINT, ARDMOR ROAD, AROS, ISLE OF MULL
* RANGE SAFETY CRAFT, 6 HAMM BEACH ROAD, PORTLAND
* THE PIER HEAD, CRAIGNURE, ISLE OF MULL
* 1 MULBERRY QUAY, MARKET STRAND, FALMOUTH
* STENA LINE, FISHGUARD HARBOUR, GOODWICK

just to throw it out there as an option, instead of doing this at query-time in the address model, you could make this assumption at import time when you build the uprntocouncil table and make sure you've got a uprntocouncil record for every uprn. You could also chuck out the list at the end of the management command to sanity check the list/assumptions.

GeoWill:
It is great that you have that list to hand Chris :joy: ! 

I did it this way (instead of dealing with them at import time) because I couldn't find a performant way of identifying them.

```sql
select * from addressbase_address aa
where aa.uprn not in (select au.uprn from addressbase_uprntocouncil au)
```
Takes an age. 

Doing it when the point in polygon lookup happens also really slows down the query - i.e. if I make this line https://github.com/DemocracyClub/UK-Polling-Stations/blob/f214b45607d34bbb42a2d2fdd28de29815006348/polling_stations/apps/addressbase/management/commands/create_uprn_council_lookup.py#L54 a `LEFT JOIN` (meaning we'd get the points that aren't covered by any council).

The other angle I've considered is a slow command that gets the list, and running that on every new addressbase release, put the list in a `BAD_UPRNS` setting, and then another command to delete them - but that felt fiddlier/more hands on than this. 

Perhaps there's another way?

chris48s:
> It is great that you have that list to hand Chris

What list? Typed them in from memory :rofl: 
(that list was in the wiki)

> Perhaps there's another way?

I tried running this locally, which identifies any records that do exist in `addressbase_address` but don't have a corresponding record in `addressbase_uprntocouncil`:

```sql
SELECT aa.uprn, aa.address, aa.postcode, au.uprn
FROM addressbase_address AS aa
LEFT JOIN addressbase_uprntocouncil AS au
ON aa.uprn=au.uprn
WHERE au.uprn IS NULL;
```

and it took about a minute for me. Doing it with a `LEFT JOIN` on UPRN rather than `IN` allows the query to use the index. That's not _quick_ , but I reckon adding another minute to the end of the addressbase build/import would be an OK tradeoff here.

GeoWill:
> That's not quick

But plenty quick enough.
Nice one :pray:

PR #3137, opened 2021-03-10
Import East Staffordshire

alexsdutton:


alexsdutton:
Gonna look at this again before re-requesting a review, and will also squash that wip commit.

polling_stations/apps/data_importers/management/commands/import_east_staffs.py
```diff
@@ -2,55 +2,32 @@
...
 class Command(BaseDemocracyCountsCsvImporter):
-    council_id = "E07000193"
-    addresses_name = "parl.2019-12-12/Version 1/Democracy Club Polling Districts.csv"
-    stations_name = "parl.2019-12-12/Version 1/Democracy Club Polling Stations.csv"
-    elections = ["parl.2019-12-12"]
+    council_id = "EST"
+    addresses_name = "2021-02-23T13:27:13.045364/Democracy Club - Polling Districts.csv"
+    stations_name = "2021-02-23T13:27:13.045364/Democracy Club - Polling Stations.csv"
+    elections = ["2021-05-06"]
     allow_station_point_from_postcode = False
```
GeoWill, line 9: Another thing that's changed! This is just the default behaviour now. So there's the `-p` flag for when we run a command locally, but we never do that in prod because it made too many dodgy maps.

GeoWill (changes requested):
1 small change, otherwise looking good :+1:

PR #3204, opened 2021-03-16
Welsh council names

alexsdutton:


alexsdutton:
CI failure is because there's now a second form (language selection) on the pages tested. Will need to fix the tests to pick out the right form.

alexsdutton:
Merging so we can get transifex set up :-)

PR #3319, opened 2021-03-23
Deal with duplicate elements in addresslist

GeoWill:
The check against addressbase was mutating the list it was iterating
over, meaning duplicate elements weren't all getting deleted, and then
uprns were getting assigned pollingstations, where we weren't confident
of the match.

### DIFF
--- a/polling_stations/apps/data_importers/data_types.py
+++ b/polling_stations/apps/data_importers/data_types.py
@@ -314,6 +314,7 @@ class AddressList(AssignPollingStationsMixin):
         self.elements = [e for e in self.elements if e["uprn"] in addressbase_data]
 
     def remove_records_that_dont_match_addressbase(self, addressbase_data):
+        to_remove = []
         for input_record in self.elements:
             addressbase_record = addressbase_data[input_record["uprn"].lstrip("0")]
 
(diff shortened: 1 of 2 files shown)


polling_stations/apps/data_importers/data_types.py
```diff
@@ -323,7 +324,8 @@ def remove_records_that_dont_match_addressbase(self, addressbase_data):
             ):
                 continue
             else:
-                self.elements.remove(input_record)
+                to_remove.append(input_record)
+        self.elements = [e for e in self.elements if e not in to_remove]
```
symroe, line 328: Not sure if this matters, but you could use sets for `to_remove` and `self.elements` and then do 

```python
self.elements = self.elements - to_remove
```

If there's a reason for `self.elements` to be a list, then never mind :)

polling_stations/apps/data_importers/tests/test_address_list.py
```diff
@@ -372,6 +372,50 @@ def test_remove_records_that_dont_match_addressbase(self):
         address_list.remove_records_that_dont_match_addressbase(addressbase_data)
         self.assertEqual(expected, address_list.elements)
 
+    def test_remove_records_that_dont_match_addressbase_with_duplicates(self):
```
symroe, line 375: By the way, it's really nice that we can unit test like this :)

polling_stations/apps/data_importers/data_types.py
GeoWill: I think there's an `AddressList` used to be an `AddressSet` -> https://github.com/DemocracyClub/UK-Polling-Stations/commit/4949ddaaa69b1d416da82ba930398c6279404d1e
TBH I just stuck with the framework that was there :). 
I'll make a note to revisit whether there'd be any advantage to going back to a set when there's less imports to write!

PR #3320, opened 2021-03-23
import script for darlington (closes #3258)

### LINKED ISSUE #3258: Import DAL-Darlington Borough Council
EMS: Xpress DC
Files:
- `DAL/2021-03-19T13:08:53.439791/darlington.gov.uk-1616156390000-.tsv`

smsmith97:
Changed from a coord to a postcode fix for consistency.

### DIFF
--- a/polling_stations/apps/data_importers/management/commands/import_darlington.py
+++ b/polling_stations/apps/data_importers/management/commands/import_darlington.py
@@ -3,20 +3,16 @@ from data_importers.management.commands import BaseXpressDemocracyClubCsvImporte
 
 
 class Command(BaseXpressDemocracyClubCsvImporter):
-    council_id = "E06000005"
-    addresses_name = (
-        "parl.2019-12-12/Version 1/Democracy_Club__12December2019darling.tsv"
-    )
-    stations_name = (
-        "parl.2019-12-12/Version 1/Democracy_Club__12December2019darling.tsv"
-    )
-    elections = ["parl.2019-12-12"]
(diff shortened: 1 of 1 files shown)


polling_stations/apps/data_importers/management/commands/import_darlington.py
```diff
@@ -1,49 +1,33 @@
...
 
     def station_record_to_dict(self, record):
-
-        if record.polling_place_id == "5634":  #  The Reading Room
-            rec = super().station_record_to_dict(record)
-            rec["location"] = Point(-1.49308179, 54.48766450, srid=4326)
-            return rec
+        # The Reading Room Neasham Darlington DL2 1PH
+        if record.polling_place_id == "6423":
+            record = record._replace(polling_place_postcode="DL2 1QX")
```
GeoWill, line 14: Makes sense to fix the postcode if it's wrong, but no need to throw away the coordinates - we won't actually show users a map if all we have is the postcode. 

You can combine the two approaches (manipulation record and rec) like this: 

```python
    def station_record_to_dict(self, record):
        # The Reading Room Neasham Darlington DL2 1PH
        if record.polling_place_id == "6423":
            record = record._replace(polling_place_postcode="DL2 1QX")
            rec = super().station_record_to_dict(record)
            rec["location"] = Point(-1.49308179, 54.48766450, srid=4326)
            return rec
        return super().station_record_to_dict(record)
```

GeoWill (approved):

Issue #2051, opened 2019-05-23
Time-specific content for polling day


chris48s:
https://github.com/DemocracyClub/UK-Polling-Stations/pull/2037 introduces the idea of content that is only shown while polls are open for one election. There's probably scope to generalise/extend that concept slightly. For example:

* "election is today polls open at 7am" before 7am (just in case, for example, a major data partner links to us at exactly midnight :wink: )
* "polls have now closed" after 10pm

etc

Think about this a bit more. See also: `NEXT_CHARISMATIC_ELECTION_DATE` setting

PR #3488, opened 2021-04-06
Tidying show_polls_open

alexsdutton:
Shows the card until 10pm on polling day. Shows "today" instead of the date on polling day itself. All marked for translation, and uses localised date formats and day of week name.

Screenshots with a Welsh `Accept-Language`, but without Welsh translations available:

Polls open tomorrow:
Screenshot from 2021-04-08 14-18-41

Polls open today:
Screenshot from 2021-04-08 14-19-18

### DIFF
--- a/locale/cy/LC_MESSAGES/django.po
+++ b/locale/cy/LC_MESSAGES/django.po
@@ -15,7 +15,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: UK-Polling-Stations\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-04-06 20:32+0100\n"
+"POT-Creation-Date: 2021-04-08 14:13+0100\n"
 "PO-Revision-Date: 2016-03-16 21:36+0000\n"
 "Last-Translator: Gareth Morlais\n"
 "Language-Team: Welsh (United Kingdom) (http://www.transifex.com/democracy-"
@@ -535,7 +535,7 @@ msgstr ""
 
 #: polling_stations/templates/fragments/you_dont_need_poll_card.html:12
 #: polling_stations/templates/fragments/you_dont_need_poll_card.html:30
-#: polling_stations/templates/home.html:38
+#: polling_stations/templates/home.html:47
 msgid "You must vote at your assigned polling station."
 msgstr ""
 
@@ -544,7 +544,7 @@ msgid "You don't need your poll card to vote"
 msgstr ""
 
 #: polling_stations/templates/fragments/you_dont_need_poll_card.html:21
-#: polling_stations/templates/home.html:41
+#: polling_stations/templates/home.html:50
 msgid ""
 "If you are registered to vote, but you don't have your poll card, you can go "
 "to the polling station and give them your name and address."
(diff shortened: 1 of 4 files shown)


GeoWill:
I missed that `poll_date` got dropped in https://github.com/DemocracyClub/UK-Polling-Stations/commit/7a76ac1748b9b02db834ae67d2fd38e542fec88a#diff-d7c09c4eedb17666b003d288e905d31514d1a355cfde36f8ca30a8f0e2b05265L36.
I guess that's a change in behaviour, because before we started showing information about when polling day was whenever we changed these [dates](https://github.com/DemocracyClub/UK-Polling-Stations/blob/7903967ae3c04ba1191ede330f489c83b58e6b81/polling_stations/apps/data_finder/views.py#L90-L93) - eg 7f240b0072ae7ee0ef5bed335073c1de28caee9a. 
If I've correctly understood how 7a76ac1748b9b02db834ae and this PR interact then I don't think we'll show users the date that the polls open untill midnight on the 5th/6th May. Traffic starts ramping up a couple of days before that, so showing the polling date a few days early would be preferable.

alexsdutton:
:-(. That'll be because I misinterpreted the variable name when I came to do the translation stuff. I can reinstate it. Do we just want to say "in the X days before", with a fixed X?

#2051 suggests having the election date in a setting, so I could get round to implementing that?

GeoWill:
So we already have `NEXT_CHARISMATIC_ELECTION_DATE` which gets set [here](https://github.com/DemocracyClub/polling_deploy/blob/4875b5b4a01b7c115113cd8861d1d25e9a463876/webapp_settings/production.py#L67). But that isn't in the nice human readable 'Thursday 6th May` form. 

`NEXT_CHARISMATIC_ELECTION_DATE`  exists to solve other problems - eg by elections happening just before the 'main' election. I don't think that's happening this year, but I don't want to get rid of the concept either. I don't have a neater model in my head for how to stitch all this together than using a conditional to set the `poll_date` value in `context` - so would probably just pull that back in - it worked for 2019 GE. Although a lot of this is quite manual I feel like that's OK becuase it's also occaisional. 

> Do we just want to say "in the X days before", with a fixed X?

I think not, I don't have this totally clear in my head but I feel like it could bite us in the event that there was a by election in the days running up to the main election WDIV was covering.

> #2051 suggests having the election date in a setting, so I could get round to implementing that?

I want to get it clear to myself why it would be different to `NEXT_CHARISMATIC_ELECTION_DATE` - which I'm not currently before adding another setting.

chris48s:
If you want to put `NEXT_CHARISMATIC_ELECTION_DATE` into another format, you can do something like 

```py
from datetime import datetime
datetime.strptime(settings.NEXT_CHARISMATIC_ELECTION_DATE, "%Y-%m-%d").strftime("%a, %d %b %Y")
```

although you might want to take the decision that you present dates using `"%d/%m/Y"` as your human-readable format for example so that you don't have to work out how to translate the output of strptime. It might be nicest to do this as a context processor https://github.com/DemocracyClub/UK-Polling-Stations/blob/eaaa1a23eadb0cb1844222fbacbc099b7bf869a6/polling_stations/apps/pollingstations/context_processors.py

I don't think there is any need to have a separate setting for this, but you do need to tolerate the case where `NEXT_CHARISMATIC_ELECTION_DATE` is None or in the past.

GeoWill (approved):
This looks good. 
We now can't set `NEXT_CHARISMATIC_ELECTION_DATE` without showing the show_polls_open card, but I don't think that's a problem becasue those are the polling stations we've imported. I've checked `NEXT_CHARISMATIC_ELECTION_DATE` as past, today and future and it behaves as expected :+1:

PR #3573, opened 2021-04-13
Change of a Torfaen station

smsmith97:


GeoWill (changes requested):
Looks good, just one small change. Once that's done could you also rebase it onto master. That will get rid of the merge conflicts github's complaining about.

PR #3661, opened 2021-04-20
Welsh council name imports

GeoWill:


polling_stations/apps/councils/management/commands/import_councils.py
```diff
@@ -174,6 +175,9 @@ def import_councils_from_ec(self):
                 council.registration_phone_numbers = registration["tel"]
                 council.registration_website = registration["website"].replace("\\", "")
 
+            if council.council_id in WELSH_COUNCIL_NAMES:
+                council.name_translated["cy"] = WELSH_COUNCIL_NAMES[council.council_id]
```
alexdutton, line 179: I would ensure that the translation is removed if the council disappears from `WELSH_COUNCIL_NAMES`, as it could be suprising for someone to remove one and then it not disappear from the site, e.g.:

```python
if council.council_id in WELSH_COUNCIL_NAMES:
    council.name_translated["cy"] = WELSH_COUNCIL_NAMES[council.council_id]
elif council.name_translated.get("cy"):
    del council.name_translated["cy"]
```

GeoWill: Good shout.

PR #3725, opened 2021-04-27
Uncomment the language chooser form in the base template

alexdutton:
We now have better Welsh translations (#3724), so let's let people use them.

Merge this after #3724.


alexdutton:
Struggling with getting the Aloe feature tests to pass, so will come back to this tomorrow.

GeoWill (approved):
:wales: :rocket:

PR #3742, opened 2021-04-28
Add Mapbox directions client

GeoWill:


polling_stations/settings/constants/directions.py
```diff
@@ -4,5 +4,10 @@
 GOOGLE_API_KEYS = []
 BASE_GOOGLE_URL = "https://maps.googleapis.com/maps/api/directions/json?units=imperial"
 
+# Mapbox settings used by directions helper
+MAPBOX_API_KEY = ""
```
chris48s, line 8: Couple of quick comments here. The first is blindingly obvious - remember to do a corresponding PR to the polling_deploy repo which adds this setting to production.py and populates it from the vault :) Less obviously, one of the things we did in WDIV is we added some [django system checks](https://docs.djangoproject.com/en/3.2/topics/checks/) to try and help guide setup and/or prevent ourselves from misconfiging the app. There are a few of these knocking about but the system checks for the data finder app are in https://github.com/DemocracyClub/UK-Polling-Stations/blob/c5c9c1cd4290bbb0bc67d3552cafbd98c0aae60e/polling_stations/apps/data_finder/checks.py Tbh, there is some cruft here. I don't think we have used MapQuest in looooong time and I deleted all the code for it, but maybe this is a good opportunity to update these with the current constraints/assumptions.

GeoWill: Blindingly obvious, often isn't! In this case I'd had a go - https://github.com/DemocracyClub/polling_deploy/pull/266.

Good pointer about the checks - that's a module (and concept - django checks) I had no awareness of even existing :smile:

polling_stations/apps/data_finder/helpers/directions.py
```diff
@@ -97,12 +110,69 @@ def get_route(self, start, end):
...
+
 class DirectionsHelper:
     def get_directions(self, **kwargs):
         if kwargs["start_location"] and kwargs["end_location"]:
-            clients = (GoogleDirectionsClient(),)
+
+            clients = (
+                MapboxDirectionsClient(),
+                GoogleDirectionsClient(),
+            )
```
michaeljcollinsuk, line 172: Workaround idea I was thinking about on standup - if you had a conditional here
```
if settings.SKIP_MAPBOX:
    clients = (GoogleDirectionsClient(),)
```
But would still like to understand how to properly mock within the integration tests

polling_stations/settings/constants/directions.py
GeoWill: Going to punt this to here https://github.com/DemocracyClub/UK-Polling-Stations/issues/3751

polling_stations/apps/data_finder/helpers/directions.py
```diff
@@ -97,12 +98,67 @@ def get_route(self, start, end):
...
+        if directions["code"] != "Ok":
+            raise DirectionsException(
+                "Mapbox Directions API error: {}".format(directions["code"])
+            )
+
+        route = directions["routes"][0]["geometry"]
+
+        time = f'{math.ceil(directions["routes"][0]["duration"]/60)} {_("minute")}'
+
+        dist = f'{math.ceil(directions["routes"][0]["distance"]*0.0006213712)} {_("miles")}'
```
alexdutton, line 139: 0.0006213712 looks like a magic number; is it metres to miles?

GeoWill: Yep - is there a better way of doing this?

chris48s: If you give it a name (a const or a function) then it will be clearer what `0.0006213712` is/means when someone comes to read this in future. For example:

```
def metres_to_miles(metres):
    return metres*0.0006213712
```

`dist = math.ceil(metres_to_miles(directions["routes"][0]["distance"]))`

documents this

polling_stations/apps/data_finder/helpers/directions.py
chris48s: In general I think moving to factories etc for test data is a good move for unit tests :+1:  VCR is useful in certain limited situations - there's a bit more discussion about where VCR works well/badly (IMHO) in https://github.com/DemocracyClub/aggregator-api/pull/56#issuecomment-507747159

That said, with the end-to-end tests, they are a blackbox test: Because you are literally spinning up an application server and a (headless) web browser and making HTTP calls to the app server, it is much harder to "mock" individual components inside the server so easily like you can in a unit test so VCR (or something similar) is the easy option for mocking external interactions. Its a blackbox test and a blackbox approach to mocking. I'd say your options are basically:

- Create VCR responses for the mapbox calls (is there a serious downside to this you are trying to avoid?)

- If you just want to call gmaps, use the test settings to configure what services you are going to call. A `SKIP_MAPBOX` setting is fairly crude, but settings are just python, so you could literally set `DIRECTIONS_CLIENTS = (MapboxDirectionsClient(), GoogleDirectionsClient(),)` in `base.py` and `DIRECTIONS_CLIENTS = (GoogleDirectionsClient(),)` in `testing.py` then set `clients = settings.DIRECTIONS_CLIENTS` here. Another approach would be to determine which clients you're going to call based on which services you have keys for e.g:

```py
clients = []

if settings.MAPBOX_API_KEY:
    clients.append(MapboxDirectionsClient())

# you can actually bin the ability to rotate multiple google keys now - we only need one
if settings.GOOGLE_API_KEYS:
    clients.append(GoogleDirectionsClient())
```

..and then in `testing.py` you can set a dummy key for whichever service you want to use/mock under test, That said, for dev purposes it might be useful to retain the ability to do low scale anonymous usage on some service without having a key for either.

In fact, you might be able to just skip this interaction completely when running the end-to-end tests (I can't remember if we actually make any assertions about map/directions in the end-to-end tests?). If not you could give neither a key or set `DIRECTIONS_CLIENTS = []` under test and bin the mocks for a directions service.

Anyway, there's a couple of other ideas for controlling it with test settings..

polling_stations/apps/data_finder/helpers/directions.py
alexdutton: I've picked this up in https://github.com/DemocracyClub/UK-Polling-Stations/blob/translation-fixes-2021-04-29/polling_stations/apps/data_finder/helpers/directions.py#L22-L24.

Both Google and Mapbox return a numerical distance in metres, so we can put that in our Direction object and fetch the more usual unit from a property in the template.

PR #3744, opened 2021-04-28
Welsh translation data

alexdutton:


alexdutton:
No longer needed. Refreshed from Transifex in #3745, and the JS change is in #3746.

PR #3755, opened 2021-04-29
Add covid keep safe message

GeoWill:
This seems like it might be too many words, interested to hear other's thoughts. As it's below the scroll bar on most screens I don't really mind including it though. I'd be nervous of putting it on the results page as it wouldn't get read at the bottom, and any higher and it would push other stuff way down.

Overview shot of home page: 
Screenshot from 2021-04-29 15-36-31

This is what the card looks like:
Screenshot from 2021-04-29 15-35-13

### DIFF
--- a/polling_stations/templates/fragments/keep_safe.html
+++ b/polling_stations/templates/fragments/keep_safe.html
@@ -1,15 +1,15 @@
 {% load i18n_with_welsh %}
-  <div class="card centered-card">
+  <div class="card">
     <h2>{% trans "Keeping you and fellow voters safe!" %}</h2>
       <p>{% trans "Local election teams will be doing everything they can to ensure voting in person on polling day is safe for everyone. This means there will be safety measures in place at polling stations to prevent the spread of COVID-19, many of which you’ll be very familiar with! " %}</p>
 
       <ul>
-        <li>{% trans "There will be a limit to the number of people allowed in a polling station at any given time to allow for social distancing." %}</li>
-        <li>{% trans "Please remember your face covering- you will need this to enter the polling station and will be expected to wear it throughout." %}</li>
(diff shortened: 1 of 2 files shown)


polling_stations/templates/fragments/keep_safe.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n_with_welsh %}
+  <div class="card centered-card">
```
symroe, line 2: Can we get rid of `centered-card` please?

polling_stations/templates/fragments/keep_safe.html
```diff
@@ -0,0 +1,15 @@
...
+        <li>{% trans "Please remember your face covering- you will need this to enter the polling station and will be expected to wear it throughout." %}</li>
+        <li>{% trans "Make sure to sanitise your hands before and after voting. In many places, sanitiser will be available but it may be a good idea to carry your own!" %}</li>
+        <li>{% trans "Some polling stations will supply clean pencils for voters but it may be best to bring your own with you." %}</li>
+        <li>{% trans "Polling Stations will be cleaned regularly by polling staff. You may need to wait while a booth is sanitised before you enter." %}</li>
+        <li>{% trans "Polling Station staff may be working behind safety screens- this doesn’t mean you can’t ask for assistance if you require it!" %}</li>
+      </ul>
+      <p>{% trans "With these new safety measures in place, it may take a little bit longer to vote than usual. Remember that if you are in the queue to vote before 10pm, you will still be able to vote even when polls ‘officially close’." %}</p>
```
symroe, line 14: Why the scare quotes?

GeoWill: Good question! Not needed.

GeoWill:
Fixups: Screenshot from 2021-04-29 16-16-40

PR #3756, opened 2021-04-29
Translation fixes 2021 04 29

alexdutton:
Three things that improve translations:

* more strings marked for translation
* the "It's about _x_ miles away or a _y_ minute walk/drive from _z_." string creation has been tidied up so it can be translated whole
* likewise with the "get Google / CycleStreets directions" paragraph, as before the "Get … or …" part was untranslatable.

As a side effect, directions handling has been made a bit more verbose, but is hopefully clearer now.

We'll run makemessages on a separate PR, to catch the covid-safe messaging text introduced in #3755.


GeoWill (approved):
This looks great!

Nice to have consistent handling of time/distance between the routing clients and general clean up. :+1:

PR #3759, opened 2021-04-30
Fix spacing in Welsh string

zuzak:
The website header at the moment is "Etholiadau6 Mai 2021 ".
It should probably be "Etholiadau 6 Mai 2021".


alexdutton:
Hi @zuzak! Thank you for that; it's much appreciated! :-)

PR #3780, opened 2021-05-04
Start logging referer when we log postcode views

GeoWill:
This should give us an idea of what proportion of users are seeing an
addresspicker.


chris48s:
Hopefully I am telling you about something you already have made a plan for, but juuust in case...

Remember this table lives on the shared RDS not the local instance and migrations are not automatically applied to the shared RDS. This means if you just build an image with this and deploy it naively, the new image will never swap in because every healthcheck ping will throw a 500 trying to insert referer into a column that doesn't exist and the instance will never be marked healthy (the healthcheck pings a real postcode search, including creating a log entry).

Assuming you're going to try and deploy this before polling day, that means you need to do one of three things to deploy it:

1. Approach one - the clean way:

    - Write the logging code so it can cope with the `referer` column either existing or not existing
    - Deploy this
    - Migrate logger
    - Rewrite the code to simplify it back to a stage where it just expects `referer` to exist
    - Build and deploy again

2. Approach two - the dirty way: Leave the code as-is and find a way to apply this migration to the logger DB before you try and deploy an image with this change applied. For adding a column, this should be fine as Django's ORM always explicitly lists all the columns on `INSERT` to protect you here. I think in the past I've either sneakily SSHed into one of the new instances while it is pre-warming (but you need to be quick), or done something like (memory is fuzzy):

    - Bring up a staging instance with this code
    - Migrate staging logger (either do this fast or disable the healthchecks on staging so the image will swap in)
    - Manually fiddle `production.py` to point at the prod DB instead of staging
    - Restart gunicorn
    - Migrate logger
    - Shut down the "dirty" staging instance
    - Bring up a "clean" staging instance
    - Deploy the new image to prod

GeoWill:
Thanks Chris! How to do migrations on the RDS db was something I wasn't 100% about - so very helpful 'just in case' :grinning:

Issue #1400, opened 2018-06-28
Fix drafter install instructions for Mac OSX


chris48s:
See https://github.com/DemocracyClub/UK-Polling-Stations/pull/1391#issuecomment-398789958

VirginiaDooley:
This issue could not be resolved because Drafter v5 cannot be used with current versions of Xcode. 

For now, we have to rely on non mac users to implement design system classes to the API documentation. In the mean time, we will explore alternative object builders that can render objects in cases when they exist and when they don't exist - as is the case with elections.

GeoWill:
For reference on why we use api blueprint system: https://github.com/DemocracyClub/ec-api-proxy/issues/3#issuecomment-772834670

TLDR: we want multiple documented examples for a a given endpoint - e.g. what a `/postcode/...` response looks like when there is/isn't an address picker.

PR #3837, opened 2021-06-17
Update install instructions

VirginiaDooley:
After going through the install in README.md, I found a few edits and added some additional instructions for beginners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ For other linux distributions, see [here](https://docs.djangoproject.com/en/2.2/
 
 The API docs rely on [drafter](https://github.com/apiaryio/drafter/) for parsing API Blueprint. This does not need to be installed manually on linux. On OSX this can be installed using
 ```
-brew install --HEAD https://raw.github.com/apiaryio/drafter/master/tools/homebrew/drafter.rb
+brew install drafter
 ```
 
 ### Install python dependencies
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ For other linux distributions, see [here](https://docs.djangoproject.com/en/2.2/
 
 The API docs rely on [drafter](https://github.com/apiaryio/drafter/) for parsing API Blueprint. This does not need to be installed manually on linux. On OSX this can be installed using
 ```
-brew install --HEAD https://raw.github.com/apiaryio/drafter/master/tools/homebrew/drafter.rb
+brew install drafter
```
chris48s, line 41: The existing install instruction here is probably wrong, but the "corrrection" probably is too.
django-apiblueprint-view needs specifically drafter v3.2.7
If you're not able to use the bundled shared object then you need to install that version. That command will install the latest version (v5.0.0) which will throw an error trying to render the API docs.
I've never used homebrew (or a mac) so I have no idea what the relevant brew incantation would be but it is probably.. not that.

GeoWill: Thanks for flagging that Chris. Since this has an issue (#1400) I'm happy for it to be merged - I don't think it's _more_ broken... 
Since this PR is the precursor to Virginia doing some work on WDIV design I guess we'll see if building the docs locally is needed, and cross that bridge then.

GeoWill (approved):

README.md
chris48s: > Since this has an issue (#1400) 

Oh well done Chris-from-the-past ;)

> I guess we'll see if building the docs locally is needed

In order to drop dc_base_theme and replace it, it will be necessary to render the API docs so you probably do need to solve this one way or another.

Issue #3792, opened 2021-05-06
Logo is squashed in Safari desktop


michaeljcollinsuk:
### Issue and Steps to Reproduce
The democracy club logo appears squashed (egg-like) when viewing in Safari desktop browser

### Screenshots
<img width="1552" alt="Screenshot 2021-05-06 at 10 37 51" src="https://user-images.githubusercontent.com/15347726/117277537-e3131b00-ae57-11eb-9f23-9f8bd7f82c5a.png">

#### Expected
Less-egg like logo looking

### Additional Details
Hopefully using the design system will resolve the issue

PR #3841, opened 2021-06-21
Design system 

VirginiaDooley:
co-author @michaeljcollinsuk 

This work includes implementing `design_system`, removing `dc_base_theme`, replacing functionality lost with that removal (such as jQuery) and introducing `dc_utils`. 

Closes https://github.com/DemocracyClub/UK-Polling-Stations/issues/3792

## Will's ToDo Before Merging:

- [ ] build and image
- [ ] deploy to staging


symroe:
I've updated the API blueprints to be in line with the design system. I know that the content isn't 100% perfect, but as this PR is about the design system and design I think I'll leave it as out of scope. Here's what it looks like:

Screenshot 2021-07-01 at 18-02-40 Where Do I Vote

GeoWill (commented):
Thanks for all the work on this, it's quite the 'first pull request in a new project' :grinning:! 
I'm learning (/getting confused by) a lot reviewing this - so for now I'm going focus on a list of problems I've spotted, rather than propose solutions. This is because I'm not immeadiatly sure how to solve them, and some of it might be obvious to you - if there's bits where it's not obvious we can pair on it, or ask for help! 
 
## Formatting, linting, nitpicks etc

### Formatting scss, js, html, templates
It would be really good to sort this out as it makes the review process much trickier. Ideally we can sort this before moving onto changing anythign else. Basically what I'd like to do is know what formatter you're using for (s)css/js etc. and run it on the codebase, commit, then rebase your pull request on that commit. I realise this is going to be a bit painful, but I think it will make it much clearer what is getting changed. I've tried this with `djhtml` and it wasn't too bad, but I don't think that is what you're using. Alternatively we can just strip out the formatting from this pr - I'm happy to help with the git gymnastics that might be involved to do that. 

### Other stuff
- Make sure everything is formatted with black and the githooks (eg whitespace at end of files).
- There are some pyflakes failings you can see thise in the `pyflakes-check` section if you run `pytest --flakes`

## Warnings and tests
- Warning: `?: (urls.W005) URL namespace 'dc_signup_form' isn't unique. You may not be able to reverse all URLs in this namespace`
Is it something you've encountered elsewhere?
- There are quite a lot of failing tests - if these are things you're getting to then no worries, if you're stuck with what they're meant to be testing/how to fix them then I'm happy to work on that together. I haven't really dug into them yet.

## Big list of things that doesn't look quite right 
- When swapping between the language picker I get glitch where the DC logo flashes up filling my browser window. It's not 100% of the time, but I wondered if it had anything to do with how we're loading/resizing the logo? I also get this if I go to /dashboard.
- There's no break between the ical card and the feedback card.
- Text box in feedback card doesn't stay within card/blue fill when viewed on a smaller device.
ical/feedback-textbox
- There's no option to 'report a problem' in the feedback form
- There's no map at /example
- text selection doesn't look right on the dashboard links - this may also be true elsewhere.
text_select
- data quality report is turned into one long string in dashboard view
dataqualityreport


- Map not appearing in dashboard postcode view

I think getting the sample data to import will be pretty useful for debugging this so perhaps that's something we can work on next week.

GeoWill:
> I've updated the API blueprints to be in line with the design system. [...]

These look great! I'm not seeing the change to `root.apibp` (retirieve entry point...) when I run this locally. Did it maybe get dropped during rebase faff?

VirginiaDooley:
Thanks for the review @GeoWill super helpful. Before I get started on addressing each comment, which browser are you using?

GeoWill:
Firefox 89.0 on PopOS (which is basically Ubuntu) 20.04

polling_stations/templates/api_docs/blueprints/councils.apibp
```diff
@@ -7,49 +7,41 @@ Polling stations and districts are maintained by local councils so all of our da
 
 ## Councils Collection: JSON [/councils.json]
 
+
+### Response [GET]
```
chris48s, line 11: Naming an [action](https://github.com/apiaryio/api-blueprint/blob/master/Tutorial.md#actions) "Response" doesn't really make sense in the semantics of API Blueprint. It also has the side-effect when it comes to docs rendering that you're going to have loads of headers with the same anchor. i.e: at the moment I can deep link to https://wheredoivote.co.uk/api/#retrieve-a-council-geojson-4 but this will give you loads of different headings that all have the anchor `#response-4`.

If you wanted to just get rid of these headers completely, you can omit the action name and just use an HTTP verb for the action (i.e: make the header `### GET` for all of yours) and the action header will be omitted in the output. Given all your resources (endpoints) only have one action (there are no endpoints that accept a `PUT`, `PATCH`, `DELETE`, etc), that is another option and probably makes things shorter. The requests and responses are already individually labelled.

polling_stations/apps/bug_reports/tests/test_bug_report.py
```diff
@@ -67,7 +67,7 @@ def test_post_bug_report_view_invalid(self):
         self.assertEqual(200, response.status_code)
         reports = BugReport.objects.all()
         self.assertEqual(0, len(reports))
-        self.assertIn('<div class="form-group has-error">', str(response.content))
+        # self.assertIn('<div class="form-group has-error">', str(response.content))
```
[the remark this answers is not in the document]
VirginiaDooley: This was an effort at debugging a failing test - but I didn't meant to commit it.

polling_stations/settings/local.example.py
```diff
@@ -2,7 +2,7 @@
     "default": {
         "ENGINE": "django.contrib.gis.db.backends.postgis",
         "NAME": "polling_stations",
-        "USER": "postgres",
+        "USER": "virginiadooley",
```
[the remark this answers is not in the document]
VirginiaDooley: No it is not; I have removed it.

VirginiaDooley:
**Re: Formatting scss, js, html, templates**
No, I didn't use djhtml bc it wasn't working as hoped; but I've just realised I had the setting configured incorrectly. I've just tested it in the pre-commit config so it should be ok going forward and I can scrap the VS code linter I was using (Beautify). Perhaps I could start by dropping [this commit](https://github.com/DemocracyClub/UK-Polling-Stations/pull/3841/commits/56fff2d4a8a60ce844ccc957516f6e4445d254f7). There is a fair bit of reindentation happening early on in [this commit](https://github.com/DemocracyClub/UK-Polling-Stations/pull/3841/commits/9f7c9635cabe9b37b4ce7157c353261fa0a875fa) - which we could pick out - or simply review together on a call so I can explain changes (which are really just replacing classes and reformatting). 

**Re: Warnings and tests**
There is something going on with `dc_utils` and forms that is causing random errors, as Michael referred to on Friday. Let's look at this as a team since it affects several products. 

**Re: Big list of things that doesn't look quite right**
Thanks for catching those. Let's review together with data import as suggested. 
I managed to import councils but not ONSPD data. 
I think the logo issue is about how it's being loaded. However, when I add a logo file to assets, I get a pipeline error. I think the solution may be to add the icon to `dc_utils` since it's now used across products in combination with a text based logo.

polling_stations/settings/base.py
```diff
@@ -1,8 +1,8 @@
 import os, sys
 
 from django.utils.translation import gettext_lazy as _
-
-
+from dc_utils.settings.pipeline import *
+from dc_utils.settings.pipeline import get_pipeline_settings
```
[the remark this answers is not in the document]
VirginiaDooley: I've resolved this by removing line 5 and adding: 

`flakes-ignore = env/*` to `pytest.ini`

polling_stations/templates/api_docs/blueprints/councils.apibp
symroe: @GeoWill are you able to look at this when you get in to updating the content of the docs? Doesn't have to be this PR, as the main point here is to add the CSS.

(Chris, Virginia can't install blueprint on her computer — or put another way, that version can't be installed on Macs at all any more — so me and Will are having to do all work that involves the docs)

GeoWill: Yes for sure.

polling_stations/apps/mailing_list/urls.py
```diff
@@ -0,0 +1,23 @@
...
+        csrf_exempt(
+            SignupFormView.as_view(
+                template_name="base.html",
+                form_class=MailingListSignupForm,
+                backend="local_db",
+            )
+        ),
+        name="mailing_list_signup_view",
+    ),
+    url(r"^api_signup/v1/", include("dc_signup_form.signup_server.urls")),
```
VirginiaDooley, line 22: @michaeljcollinsuk Can you check that this is correct? I have a duplicate url error.

GeoWill:
So just a blitz of things that jump out based on running this locally and having a click around, rather than digging through the code. As I said if you want a hand with the failing tests let me know. 

http://127.0.0.1:8000/address/100040234437/
- Maps are back, but icon illustrations on the pins are gone - contrast with https://wheredoivote.co.uk/example/. This is also the case on the dashboard/postcode/<postcode> pages. I'm pretty sure these are courtesy of the `leaflet-extra-markers` library
- No gap between ical feed card and feedback card - it makes it look like the feedback is especially to do with election reminders.
- Is it possible for 'yes', 'no' and 'report a problem' to all be on the same row? I think there's a weird visual hierarchy with them stacked.
- This might be outwith this pr but the amount of information you get before having to scroll is quite a bit less compared to https://wheredoivote.co.uk/example/. I think this is mainly down to the language bar and the white space before the first card - what do you thinkg? 

http://127.0.0.1:8000/dashboard/council/EXE/
- data quality report is much better, but would be nicer to just see the whole thing (at the moment I have to scroll up and down in a 'code block' style box) - may seem minor but not when you have to look at lots of them. 

http://127.0.0.1:8000/dashboard/
Summary table doesn't show full width - ditto may seem minor but not when you have to look at it many times.  
Link highlighting different but still not right - should this be bumped to an issue on the design system repo? 
text-highlight

VirginiaDooley (commented):
@GeoWill Since I've moved everything that the `dc_base_theme` was adding to the html structure into `base.html`, this test was failing. I've updated it but it would also be good to confirm whether `whitelabel` is being used in anywhere else. I think Sym said we plan to remove this functionality from the code base.

VirginiaDooley (commented):
@GeoWill I've added the script tag here for the time being because I can't recall how to include jquery so it's compiled properly. I'll have to check back with @symroe on this next week, but I didn't want it to be a blocker.

chris48s:
> it would also be good to confirm whether whitelabel is being used in anywhere else

The whitelabel app is what makes https://wheredoivote.co.uk/embed/ work

`/embed` in an `<iframe>` tag is the no javascript fallback for the widget. Note the recommended embed code on https://democracyclub.org.uk/projects/polling-stations/embed/ and https://github.com/DemocracyClub/WhereDoIVote-Widget/#including-the-widget-on-your-page is 

```
<noscript>
  <iframe src="https://wheredoivote.co.uk/embed/"
    style="width:100%; height:1100px" frameborder="0" scrolling="no">
  </iframe>
</noscript>
<div id="dc_wdiv" aria-live="polite" role="region"></div>
<script type="text/javascript"
  src="https://widget.wheredoivote.co.uk/wdiv.js">
</script>
```

with `<noscript><iframe src="https://wheredoivote.co.uk/embed/" ...`

Also there are some councils still using `/embed` in an `<iframe>` in preference to the widget. For example
https://www.croydon.gov.uk/council-and-elections/voting-and-elections/voting/how-vote/polling-station-finder uses it.
In some cases councils aren't able to embed the widget because `widget.wheredoivote.co.uk` is not listed in the allowed domains in their Content-Security-Policy headers and they can't add it because Electoral Services don't control that (I think Teignbridge was an example of this).

Although the legacy use-case for the unbranded or alternately branded versions of the site is gone, I think there are still definitely some uncommon-but-valid use-case still supported by `/embed`

polling_stations/templates/base.html
```diff
@@ -46,8 +46,8 @@
         {% endblock site_meta %} 
 
     </head>
-
-    <body {% block body_attrs %}{% endblock body_attrs %} class="ds-scope">
+    {% block body_base %}
+    <body {% block body_class_attrs %}{% endblock body_class_attrs %} class="ds-scope">
```
michaeljcollinsuk, line 50: This can remain named as `{% block body_attrs %}` because you might want to add other attrs rather than just the class. But more importantly, your default `class="ds-scope"` is defined outside of the block. So this means when you do use the block in the `base_embed.html` template, you are not overriding the existing attrs, you are addinging additional ones (So you would have `<body class="embed" class="ds-scope"`.

So change to:
```
<body {% block body_class_attrs %}class="ds-scope"{% endblock body_class_attrs %}>

File: polling_stations/templates/base.html [hunk 7]
```diff
@@ -46,8 +46,8 @@
         {% endblock site_meta %} 
 
     </head>
-
-    <body {% block body_attrs %}{% endblock body_attrs %} class="ds-scope">
+    {% block body_base %}
```
michaeljcollinsuk, line 49: I don't think we need to bring this back to just to use it in one place `base_embed.html` file. I think the aim should be work with the new `base.html` file. So I would try removing this again and then see comment about the `base_embed.html` structure

polling_stations/templates/base_embed.html
```diff
@@ -0,0 +1,23 @@
+{% extends "base.html" %}
+{% load i18n static %}
+
+
+{% block body_class_attr %}class="embed"{% endblock body_class_attr %}
```
michaeljcollinsuk, line 5: Re previous comment, keep this as `body_attrs`

polling_stations/templates/base_embed.html
```diff
@@ -0,0 +1,23 @@
...
+{% block body_base %}
+
+ {% block content %}{% endblock content %}
+
+    <footer>
```
michaeljcollinsuk, line 10: Try moving lines 10-17 to within the `site_footer` block

polling_stations/templates/base_embed.html
```diff
@@ -0,0 +1,23 @@
+{% extends "base.html" %}
+{% load i18n static %}
+
+
+{% block body_class_attr %}class="embed"{% endblock body_class_attr %}
+{% block body_base %}
```
michaeljcollinsuk, line 6: As per previous comment, we shouldn't need to keep this

polling_stations/templates/base_full.html
```diff
@@ -0,0 +1,42 @@
+{% load i18n %}
```
michaeljcollinsuk, line 1: I don't think we need this file, as we will create the embed file with blocks instead

polling_stations/templates/base.html
VirginiaDooley: Yes thanks, caught this one after I pushed.

GeoWill (commented):
Sorry I've been a bit slow to return to this. 
From a 'using the website point of view' it is looking great! 
I've left a couple of minor comments inline and there's one commit which still has 'WiP' in the commit message - I don't think that's true anymore, so can it be taken out. 
After that, and if you think it's ready then I'll build and image off it and do a deploy to staging, and then if that looks good we can merge.

VirginiaDooley:
https://github.com/DemocracyClub/UK-Polling-Stations/pull/3841#discussion_r686953001

This class went away with the removal of `dc_base_theme` but I'm not sure the way it's been reconstructed in `dc_utils` is in line with the design system, so I need to fix it there first before this test can be rewritten.

polling_stations/apps/bug_reports/tests/test_bug_report.py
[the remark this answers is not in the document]
VirginiaDooley: This class went away with the removal of dc_base_theme but I'm not sure the way it's been reconstructed in dc_utils is in line with the design system, so I need to fix it there first before this test can be rewritten.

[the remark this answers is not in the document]
VirginiaDooley: In the mean time, I've updated the class so the test passes.

GeoWill (approved):
Okay, I'm pretty sure this is working as intended from a user point of view, and while the commit history is quite confusing I don't think it's worth the time to unpick it all.  Also merging this should unblock #3858.

PR #3858, opened 2021-07-14
WIP/Validate html

VirginiaDooley:
Ref https://github.com/DemocracyClub/dc_django_utils/issues/6


polling_stations/apps/pollingstations/tests/test_urls.py
```diff
@@ -30,3 +34,28 @@ def test_trailing_slashes(self):
...
             assert url[-1] == "/", url + " does not end with /"
+
+
+class TestHtml:
+    @pytest.fixture
+    def urls(self):
+        return [
+            reverse("home"),
+            reverse("address_view", kwargs={"uprn": AddressFactory().uprn}),
+            # reverse("address_select_view", kwargs={"postcode": "CC11AA"}),
```
VirginiaDooley, line 45: @GeoWill I'm trying to use a Factory to test this view. I've tried `AddressFactory().postcode` (which is blank) as well as `AddressFactory(postcode="CC11AA")` without success. Do you have any suggestions?

GeoWill: I think it might need to be more like:
```python
    def urls(self):
        UprnToCouncilFactory.create_batch(3, lad="X01000000", uprn__postcode="CC11AA")
        return [...]
```

PR #3906, opened 2021-09-13
Feature/django upgrade

VirginiaDooley:
This work upgrades Website to Django 3.2 and updates site packages accordingly and addresses deprecation warnings.


polling_stations/apps/pollingstations/tests/test_urls.py
```diff
@@ -27,6 +27,7 @@ def test_trailing_slashes(self):
         )
         urls = [rec["url"] for rec in data]
         for url in urls:
+
             if self.is_exception(url):
                 continue
-            assert url[-1] == "/", url + " does not end with /"
+            assert url[-1] == "/" or ">", url + " does not end with /"
```
VirginiaDooley, line 33: I wasn't sure about this edit. This assert was failing at /admin/<url> which makes sense, but other than just checking for a trailing < along with / , is there another way to do this?

VirginiaDooley:
I'm not totally clear on what's being asked in the remaining deprecation warnings related to https://docs.djangoproject.com/en/3.2/releases/3.2/#automatic-appconfig-discovery

michaeljcollinsuk:
Just a couple of warnings when I run `manage.py check`:
```
WARNINGS:
councils.Council.name_translated: (fields.W904) django.contrib.postgres.fields.JSONField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.
	HINT: Use django.db.models.JSONField instead.
data_finder.LoggedPostcode.has_election: (fields.W903) NullBooleanField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.
	HINT: Use BooleanField(null=True) instead.
```

I think changing both of these should be fine, without any data loss.

polling_stations/apps/pollingstations/tests/test_urls.py
michaeljcollinsuk: This is the one thing I'm not sure about as well, as I dont know why we would want a URL to end with `>`

Do you know what change caused this test to begin to fail?

michaeljcollinsuk: So I think I figured this out, the URL that is breaking the test is for a view that was added in django 3.2, see the last bullet point here https://docs.djangoproject.com/en/3.2/releases/3.2/#minor-features

I added a `ipdb.set_trace()` in to the test to check the `urls` and that is indeed the only one that doesn't end in the trailing slash.

So I wonder if rather than change the assert to allow `>`, that could lead to us defining other bad urls elsewhere without us realising, we should remove this url from the list of urls to test? So that would mean adding:
```
urls.remove("/admin/<url>")
```
before the for loop.

michaeljcollinsuk (approved):

polling_stations/apps/pollingstations/tests/test_urls.py
symroe: I like Michael's suggestion here: we don't want to add a URL ending in `>` by mistake. 

Another option we might want to think about, especially if we start using `path` with placeholders more, is to allow `/<*>` at the end of a URL 9(e.g, match the opening brace and some value in the middle.

I don't think we need to do that at the moment, but I can imagine that being a valid use case in the future.

symroe (approved):

requirements/base.txt
```diff
@@ -4,7 +4,7 @@ django-basicauth==0.5.3
 django-extensions==3.1.3
 django-localflavor==3.0.1
 django-markdown-deux==1.0.5
-django==2.2.22  # pyup: >=2.2,<3.0
+Django<=3.3 # pyup: >=2.2,<3.2
```
chris48s, line 7: Drive by, but for an application (rather than a library) this should be a pin not a range. You don't want a surprise django upgrade (or downgrade - `<=3.3` really gives you some scope :laughing: ) when you deploy :grimacing:

Also pyup hasn't been running on this repo for over a year. Pyup comments can be removed. If you want to direct dependabot to ignore upgrades to the next LTS, the syntax is https://github.com/DemocracyClub/WhoCanIVoteFor/blob/efabcb7078d1f9be05baba746475ca11deb49e44/.github/dependabot.yml#L8-L10 although it would want to be `versions: [">=4.0", "<4.2"]`. In fact, that probably wants bumping on the WCVIF repo too. It looks like you've upgraded to 3.2 over there and overlooked it.

PR #3950, opened 2021-10-23
Council login

symroe:
I wanted to see what this would take. Started drawing it out and realised I could think about it better by doing it quickly and seeing what problems come up.

I mostly ripped off @michaeljcollinsuk's work from the EC API, with some additional logic for the M2M between council and user.

Here's what I've done:

## Added OTT login for gov.uk email addresses

A new form field that has custom validation for valid council email addresses. Looking at the existing email addresses in the current database, every email address ends with `gov.uk`, so this is going to validate all valid emails, however it will accept too many emails (there are a lot of valid `gov.uk` emails that aren't valid users of this service!). We can extend this to look at the domain part of the councils table, hard code them, or some other solution. The good news is that it's unlikely we're going to get malicious users from a valid `gov.uk` email address.

The system of logging in involves password-less accounts being created and log in happening via a one time token emailed to the user.

## Added M2M between `User` and `Council`

A user can now have `n` councils assigned to them. It's more than one because some people manage more than one council, or people at VJBs might want to upload data for each of the councils they cover. 

After log in, we look for existing councils that have the email domain and add them to the set of councils the user manages. 

At some point later we'll want to add this to the admin interface, so we can manually add and remove councils for a user.

## Council user facing views

Allow council users to view the uploader, and filter the councils they see by their assigned councils. Later, it would be nice to redirect to a single council view if the user only has a single council assigned (the majority case).

## Future work

There's a fair amount more to be done here, not least wrapping things up so it's easier for the users to use. We'll want to add some explanations at least. 

We also need to write some tests, text for the emails, etc etc. This is very much a (working) proof of concept rather than a finished implementation!


symroe:
@chris48s On emails:

Yep you're right about the domains, in theory at least.

As the [validation is a regex](https://github.com/DemocracyClub/UK-Polling-Stations/pull/3950/files#diff-8045943e259c2cd6f980673a0c98fdbf136070afde13329eddf3a8d0c64cfef5R15), we can `|` join together all suffixes or even whole domains as we see fit. I started with the intention of doing that, but then found that all [current contact details for electoral and registration contacts do end in a `gov.uk` domain](https://www.electoralcommission.org.uk/api/v1/data/local-authorities.json).

On the admin site: yes, I anticipate an admin interface for assigning users to councils. We might not use it much, as the churn in users might make it pointless. In that case, the person staffing the email inbox would just upload to the council using their own account, as per the current set up (as you point out). This is about reducing work for us rather than removing it.

polling_stations/apps/councils/models.py
```diff
@@ -97,3 +99,8 @@ class CouncilGeography(models.Model):
     )
     gss = models.CharField(blank=True, max_length=20)
     geography = models.MultiPolygonField(null=True)
+
+
+class UserCouncils(models.Model):
+    user = models.ForeignKey("auth.User", on_delete=models.CASCADE)
+    council = models.ForeignKey(Council, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
symroe: Yep, I'd considered that this would be a problem but not really thought about it.

One thing to think about though: the way we're doing a (form of) `get_or_create` on user accounts (using the OTT) means that it doesn't actually matter if a User is `CASCADE DELETE`d at some point: it'll just get re-created next time they log in.

Also, as council records change infrequently (their existence at least), and when they do it''s normally not around the time that we'll be getting data, it seems reasonable to run the `import_councils` command on the logger DB and just use that for the JOIN to users.

Of course, there's bound to be some problems some of the time, but see the other comment about being designed to reduce work load rather than eliminate it for us.

We're also looking at moving to an _entirely different_ deployment model that I'd be interested in talking to you about at some point (the same as the WCIVF one, if you've been following that at all)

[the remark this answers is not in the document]
chris48s: > We're also looking at moving to an entirely different deployment model that I'd be interested in talking to you about at some point (the same as the WCIVF one, if you've been following that at all)

To an extent.. I'd be interested in hearing about it. Feel free to give me a shout.

symroe:
https://github.com/DemocracyClub/polling_deploy/pull/295 adds syncing councils in the logger DB to cron and image building

symroe:
I've made a bunch of updates to this over the last few days:

## Supporting the logger DB
The top level is:
1. Run the council importer on the logger DB on a cron (see DemocracyClub/polling_deploy#295)
2. _Make an assumption_ that the logger DB's list of councils is _good enough_ for what we need. It _might not_ be the same as the list of councils used in the 'front end', but it doesn't actually matter. Think of them as two different models if you like (they are the same model in different database)
3. Update the code to use `using()` in the places where it matters. That is, in the admin interface

This needs to be tested in a prod-like environment, but local testing with two database suggests it will work.


## Enabled the Council Admin
Screenshot from 2022-01-12 15-00-43

The admin interface is now read only for `Council` objects, but does have an inline admin that allows us to assign users to councils. This should be populated automatically, but this system allows us to add people if we need to (for example, person with a `@foo.gov.uk` email address is also managing data for a `@bar.gov.uk` council)

## Better council list design

Screenshot from 2022-01-12 15-01-03

Added a base template, alpha banner and modified the council user facing table to show less stuff. The above shows a user who has those councils assigned to them manually in the admin interface

## Election dates for uploaded files

Screenshot from 2022-01-12 15-01-28

Query EE when uploading a new file to get the election date(s). If more than one future date, ask the user to assert the election the upload is for. If not more than one, store the info, but don't ask the user to pick.

This bit needs more work, especially around the path we use on S3. I've added a file version field, but not used it yet. WIP.

Screenshot from 2022-01-12 15-01-12


## Contact details on council page

So the council can check them at the point they update data

Screenshot from 2022-01-12 15-01-18

VirginiaDooley:
👏 Re: radio date picker, I realise that your layout follows the design system, however with a longer list of dates, I can imagine that it will become difficult to read. [More date padding and/or larger radio buttons](https://design-system.service.gov.uk/components/radios/) might be help in this case.

symroe:
> clap Re: radio date picker, I realise that your layout follows the design system, however with a longer list of dates, I can imagine that it will become difficult to read. [More date padding and/or larger radio buttons](https://design-system.service.gov.uk/components/radios/) might be help in this case.

Thanks!

Yes, I agree — I think we should consider making the buttons larger by default in the design system. I prefer the GDS style large buttons.

For this PR though, I think I'm happy to just follow the system. In reality, we're unlikely to have more than 2 items in the list: a council doesn't normally have more than one upcoming election _date_.

polling_stations/apps/file_uploads/views.py
```diff
@@ -71,10 +89,24 @@ def get(self, request, *args, **kwargs):
...
             .exclude(council_id__startswith="N09")
             .get(council_id=self.kwargs["gss"])
         )
+        upcoming_election_dates = EveryElectionWrapper(
+            council_id=self.kwargs["gss"]
+        ).get_future_election_dates()
+
+        # If the list returns no items, flag that there are no upcoming elections
+        # that we know about.
+        context["NO_UPCOMING_ELECTIONS"] = bool(upcoming_election_dates)
```
GeoWill, line 103: Would this make more sense as `context["HAS_UPCOMING_ELECTIONS"] = bool(upcoming_election_dates)`. At the moment if `upcoming_election_dates` is an empty list then `context["NO_UPCOMING_ELECTIONS"]` will be `False`, which I find unintuitive. But might just be me :)

symroe: Yep, I think you're right. Maybe this is why I made the mistake about showing the button when no upcoming?

polling_stations/apps/file_uploads/views.py
```diff
@@ -71,10 +89,24 @@ def get(self, request, *args, **kwargs):
...
         context["council"] = (
-            Council.objects.all()
+            Council.objects.using(LOGGER_DB_NAME)
+            .all()
             .exclude(council_id__startswith="N09")
             .get(council_id=self.kwargs["gss"])
         )
+        upcoming_election_dates = EveryElectionWrapper(
+            council_id=self.kwargs["gss"]
+        ).get_future_election_dates()
```
GeoWill, line 99: If a council has an upcoming election that it manages, but isn't for that council then I'm not sure it will get to the list. E.g. `https://elections.democracyclub.org.uk/elections/parl.southend-west.by.2022-02-03/` doesn't appear in `https://elections.democracyclub.org.uk/api/elections.json?organisation_identifier=SOS&future=1` 

Which is where the upcoming dates come from in: https://github.com/DemocracyClub/UK-Polling-Stations/blob/993e1084698d5ff0f8079a815df16bcd71b701c1/polling_stations/apps/data_finder/helpers/every_election.py#L75 populating `self.elections` with: https://github.com/DemocracyClub/UK-Polling-Stations/blob/993e1084698d5ff0f8079a815df16bcd71b701c1/polling_stations/apps/data_finder/helpers/every_election.py#L44

I'm not yet sure how to fix this as I'm not sure we store a map of local authority -> managed elections anywhere :thinking:

polling_stations/apps/file_uploads/templates/file_uploads/upload.html
```diff
@@ -121,12 +121,18 @@ <h2>Upload EMS export</h2>
                 xhr.open("POST", url, true);
                 xhr.setRequestHeader("Content-Type", "application/json");
                 xhr.setRequestHeader("X-CSRFToken", getCookie('csrftoken'));
+                var election_date;
+                if (document.querySelector('input[name="election_date"]:checked')) {
+                    election_date = document.querySelector('input[name="election_date"]:checked').value;
+                } else {
+                    election_date = document.querySelector('input[name="election_date"]').value
+                }
```
GeoWill, line 129: @symroe Would be good to get a double check on this from you. It's meant to address the issue that when there's only one upcoming election we don't show radio buttons, so `document.querySelector('input[name="election_date"]:checked')` is `null`.

polling_stations/apps/file_uploads/views.py
symroe: https://github.com/DemocracyClub/EveryElection/pull/1336 is an attempt to fix this. If that's a reasonable approach then the change to `get_future_election_dates()` shouldn't be too hard.

polling_stations/apps/file_uploads/templates/file_uploads/upload.html
symroe: I think this way is fine as is. The other options are:

1. It looks like just getting the value works, but I don't know about browser support. See https://jsfiddle.net/vjop5xtq/
2. We could just render a checked radio button in the HTML rather than a hidden input (but also hide the radio button in CSS). This would mean the JS don't have to work with both cases

PR #3995, opened 2021-11-24
hotfix/Repair broken og image tag

VirginiaDooley:
Image preview on twitter cards was not showing as seen here https://twitter.com/democlub/status/1462818360660566019

This PR reformats the tag to render the preview image. I've also defined the canonical url as it was being referenced but hadn't previously been defined. 

Screen Shot 2021-12-01 at 10 51 22 AM

WDIV compiles static images differently than other sites so here I did not use the canonical url to build the image source.


GeoWill:
This branch is live on https://stage.wheredoivote.co.uk/

VirginiaDooley:
Rather than s3 domain, configure as described here https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#cloudfront

symroe (approved):

VirginiaDooley:
From Sym re: S3 

> It actually looks like this is causing the issue: https://github.com/DemocracyClub/polling_deploy/blob/master/webapp_settings/production.py#L50
> 
> I think we can remove that line and replace it with the line you've just added, except we want the value to be it's own domain name. I reckon this is getting enough in to the deployment side of things that we should wait for Will before we deal with anything
> 
> But, it's an issue for all the assets, so we don't need to worry about it just yet

chris48s:
To fill in the background on this: The reason for this little dance:

https://github.com/DemocracyClub/polling_deploy/blob/89cf2281484f2c4c3fc896a43c85a0700aab289e/webapp_settings/production.py#L50
https://github.com/DemocracyClub/polling_deploy/blob/89cf2281484f2c4c3fc896a43c85a0700aab289e/deploy.yml#L15-L17
https://github.com/DemocracyClub/polling_deploy/blob/89cf2281484f2c4c3fc896a43c85a0700aab289e/deploy.yml#L52-L56

is because the S3 assets get synced when the image is _built_ but building is an entirely separate process from _deploying_. Making a new bucket for each build means building an image (which may not be deployed for some time or may never even get deployed if its broken in some way) can't mess up anything in the asset bucket currently in use by prod.

There's more detail on this decision in

https://github.com/DemocracyClub/polling_deploy/pull/78#issuecomment-327841228

The useful thing about doing it that way (rather than say doing the collectstatic at the point you swap in a new image, for example) is it makes rollbacks easy. Every image has its own bucket URL baked into the settings.

Issue #4053, opened 2022-01-27
Support more browsers when uploading files


symroe:
The uploader currently uses [Promise](https://caniuse.com/promises) to deal with async token requesting and uploading. 

This is a good solution, but isn't supported by all browsers. Specifically IE11. Although IE11 is hardly used at all on the internet as a whole, I have a strong suspicion that the last 1% of users are exactly _our_ users: people in UK local authorities.

We have a couple options as to how to deal with this:

1. Just detect the browser (and no JS) and show instructions as to how to email us the data. We should do this _early_ in the process, e.g before they log in.
2. Refactor the existing JS to support IE11. There are various options here to a re-write of the code, to [using a polyfill](https://www.npmjs.com/package/promise-polyfill) (untested / used)

The initial work here is to understand the problem a little more (e.g the pre-signed upload system we're using, and why) and then to work on a solution that gives IE11 users a better experience than they have at the moment (it breaks in some ways I've not tested...maybe just hanging after pressing upload?)

symroe:
See also:

* #2545
* #2544
* #2543
* #2542

chris48s:
> understand the problem a little more (e.g the pre-signed upload system we're using, and why)

Just to explain this bit a little more:

- The reason it uses a pre-signed POST to S3 is so you can upload from the user's browser straight to S3. If you were to use a more "traditional" form based approach, you'd have to allow nginx to accept a >100Mb POST body in production and ensure your prod instance has the memory/disk space to accept/store that temporarily then send it on to S3. This has obvious downsides.
- The main nice thing Promise buys you over callbacks (aside from syntactic sugar) is the `Promise.all` at https://github.com/DemocracyClub/UK-Polling-Stations/blob/3c8cec245548b0060bb087dd7b9e4bec446a33ab/polling_stations/apps/file_uploads/templates/file_uploads/upload.html#L155 which gives you an easy way to kick >1 files off in parallel and then do something once _all_ of them are finished (this is more difficult to do with callbacks, although not impossible). I think this is mainly useful for Democracy Counts where you get 2 files rather than one but I don't remember off the top of my head.

---

The path of least resistance to solve this one problem is probably to polyfill it and go about your day, but other options would be:

- Explicitly rewrite this code use callbacks instead of promises

[...]

PR #4087, opened 2022-02-07
Hotfix/Promises in IE11

VirginiaDooley:
Ref #4053

This work adds[ `promise-polyfill` ](https://www.npmjs.com/package/promise-polyfill)and makes `Promise` available in IE11. 

Confirmed with s3 upload `2022-02-07 18:54:41    8611211 ABD/2022-05-05/2022-02-07T18:54:39.429722/wdiv-upload-test.tsv`

### DIFF
--- a/polling_stations/templates/base.html
+++ b/polling_stations/templates/base.html
@@ -22,8 +22,6 @@
         {% endblock top_banner %}
 
         {% block site_meta %}
-            <script src="https://code.jquery.com/jquery-3.6.0.min.js"></script>
-            <script src="https://cdn.jsdelivr.net/npm/promise-polyfill@8/dist/polyfill.min.js"></script>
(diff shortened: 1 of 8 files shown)


polling_stations/templates/base.html
```diff
@@ -23,6 +23,7 @@
 
         {% block site_meta %}
             <script src="https://code.jquery.com/jquery-3.6.0.min.js"></script>
+            <script src="https://cdn.jsdelivr.net/npm/promise-polyfill@8/dist/polyfill.min.js"></script>
```
symroe, line 26: You don't need to use the CDN if you're installing it with npm.

Can you add a new javascript file to pipeline, copying the dashboard here:

https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/settings/static_files.py#L45-L50

And then implement it like:

https://github.com/DemocracyClub/UK-Polling-Stations/blob/faf0ad38d24f78b2d46e45a561fcbe08682db19a/polling_stations/apps/dashboard/templates/dashboard/postcode.html#L14

But only on the pages where we're actually going to use it. I'm keen on not forcing all our users to download this script for every page load, when it's mostly not going to be used.

symroe: Actually, same for jQuery. We're already installing jQuery locally so we shouldn't be using the CDN version. I know that's not added in this PR, but it would be nice to get it out of every page load if we can!

symroe (approved):

Issue #4000, opened 2021-11-30
DC logo doesn't link to company website


pmk01:
The Democracy Club logo in the footer of WDIV links to the front page of WDIV, rather than to democracyclub.org.uk (which is where it goes on WCIVF).

Issue #4082, opened 2022-02-07
Logo does not render correctly in ie11


VirginiaDooley:
### Issue and Steps to Reproduce
The DC logo is 
IE11 Needs the image nested inside `<svg>` tags with width set explicitly. 

#### Expected

![Screen Shot 2022-02-07 at 3 23 22 PM](https://user-images.githubusercontent.com/7017118/152818184-13d194ee-bf79-4662-a5a3-86466f366913.png)

#### Actual

![Screen Shot 2022-02-07 at 3 28 28 PM](https://user-images.githubusercontent.com/7017118/152818482-a87a1b6c-2aa0-46da-8fc2-65cf70c2c456.png)

### Additional Details
* Browser and version: IE11
* Were you logged in? 

![Screen Shot 2022-02-07 at 3 23 22 PM](https://user-images.githubusercontent.com/7017118/152818133-44238692-a399-42cc-8cba-890dc1b78b3e.png)

![Screen Shot 2022-02-07 at 3 14 13 PM](https://user-images.githubusercontent.com/7017118/152818121-62907c9f-1409-482b-aece-6f44a977b327.png)

VirginiaDooley:
![Screen Shot 2022-02-07 at 3 23 22 PM](https://user-images.githubusercontent.com/7017118/152818133-44238692-a399-42cc-8cba-890dc1b78b3e.png)



![Screen Shot 2022-02-07 at 3 14 13 PM](https://user-images.githubusercontent.com/7017118/152818121-62907c9f-1409-482b-aece-6f44a977b327.png)

PR #4089, opened 2022-02-07
Hotfix/logo tidy up

VirginiaDooley:
Closes https://github.com/DemocracyClub/UK-Polling-Stations/issues/4000 and https://github.com/DemocracyClub/UK-Polling-Stations/issues/4082

- Fixes logo link in footer
- Resizes logo for IE11


polling_stations/templates/base.html
```diff
@@ -38,7 +38,7 @@
             <meta name="twitter:image" content="{% block og_image_twitter_content %}{% static 'images/logo_icon.png' %}{% endblock og_image_twitter_content %}" />
             <meta name="twitter:description" content="{% block og_twitter_description %}{% if NEXT_CHARISMATIC_ELECTION_DATE %}{% blocktrans with election_date=NEXT_CHARISMATIC_ELECTION_DATE|date:'F j Y' %}Find out where to vote on {{ election_date }}{% endblocktrans %}{% else %}{% trans "Find out where to vote" %}{% endif %}{% endblock og_twitter_description %}">
             <meta name="twitter:image:alt" content="{{ SITE_TITLE }}">
-            <link rel="shortcut icon" href="{%  static 'images/logo_icon.svg' %}">
+            <link rel="shortcut icon" href="{% static 'images/logo_icon.png' %}">
```
GeoWill, line 41: I'm not sure what is being fixed here. Previously for me this introduces a white box around the DC logo circle
svg:
SVG version
png:
PNG Version

There maybe other benefits to the png over svg, but it would be good to understand what they are.

VirginiaDooley: Actually, the filetype wasn't the problem after all. Made a change that should provide wider browser support.

VirginiaDooley: `.png` has wider browser support over `.svg` and then there is `.ico` for IE.

polling_stations/templates/base.html
```diff
@@ -27,21 +27,27 @@
             <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
             <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
 
+            <link rel="shortcut icon" href="{%  static 'images/favicon.ico' %}">
```
GeoWill, line 30: I now don't have any favicon on Firefox on Ubuntu, I don't know what the issue is, but I think it's better to break it for IE11 users than for firefox users.

PR #4090, opened 2022-02-07
Cache assets synced to S3 for, more or less, ever

symroe:


symroe:
Untested, so will need an ad-hoc stage deploy and a `collectstatic` before we merge

symroe:
:D It's my understanding that `immutable` means forever, no matter the value of the `max-age`? (and max-age is there or older browsers?) I can make the max age a lot longer of course, if that's not the case

chris48s:
Bah. I came here to post jokes rather than provide actual useful code review, but since you asked my understanding of `immutable` is more that it prevents any revalidations (e.g: `If-Modified-Since`, `ETag` checks, etc) until` max-age` has passed, so

`Cache-Control: public, max-age=604800, immutable`

says "cache this for a week and don't bother checking the ETag until a week has elapsed". Probs worth doing your own reading though. Everything about Cache-Control headers is confusing and easy to get wrong :)

symroe:
This is done in polling deploy. We should do it here and implement this, but closing this PR for now

PR #4132, opened 2022-02-18
Logo fixes

GeoWill:
-Link footer logo to DC Website
-Override logo height


GeoWill:
I merged this, but when I deployed it seems to have broken the css somewhere. I haven't managed to reproduce locally yet. However this branch is deployed to staging so you can see what it looks like.

VirginiaDooley:
This work is now redundant as a result of the implementation of the default base template from `dc_utils`

PR #4202, opened 2022-03-03
Import improvements

GeoWill:


polling_stations/apps/data_importers/base_importers.py
```diff
@@ -619,6 +618,21 @@ def import_residential_addresses(self):
...
+            )
+        )
+        self.addresses.elements = [dict(s) for s in element_set]
+        self.csv_row_count = len(self.addresses.elements)
+        self.write_info(
+            "Addresses: Found {:,} distinct records in input file".format(
+                self.csv_row_count
+            )
+        )
+        self.write_info("----------------------------------")  #
```
GeoWill, line 634: This is a bit janky, but I think the alternative is trying to strip out the duplicates from the csv itself as part of the lambda trigger. I think it's better to preserve exactly what council sends us, so I think this (or something similar) is the next best thing.

PR #4212, opened 2022-03-05
Add a council csv api endpoint

GeoWill:
This lets us easily import this into a google sheet, in order to keep
track of which councils are live on the site.

### DIFF
--- a/polling_stations/apps/api/councils.py
+++ b/polling_stations/apps/api/councils.py
@@ -1,3 +1,4 @@
+from django.db.models import Count
 from django.http import HttpResponsePermanentRedirect, Http404
 from rest_framework.decorators import action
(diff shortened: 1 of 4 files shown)


symroe:
The output looks good to me, but I think it's worth modifying the ViewSet queryset to annotate the count of stations, rather than doing O(n) queries:

```python
Council.objects.all().order_by("council_id").annotate(stations=Count("pollingstation"))
```

```python
station_count = serializers.IntegerField(
    source='stations', 
    read_only=True
)
```

(the last bit is untested, but it'll be something like that)

For good measure, can you add an `assertNumQueries` to the test too — we should only make one query for this.

symroe (approved):

Issue #4181, opened 2022-03-01
Feature: Support "Advance Voting"


symroe:
# Background
 
The Welsh Government are piloting "Advance Voting": voting in advance of polling day at defined locations.

[Welsh government statement and update on Flexible Voting Pilots](https://gov.wales/written-statement-update-flexible-voting-pilots)

The areas are:

> **Blaenau Gwent** – the centrally-located Ebbw Vale Learning Zone will be used as an advanced polling station for all residents of the county, including students of the college. Advance voting will be available during the week leading to Election Day.
> 
> **Bridgend** - polling stations in certain low turnout wards will be open for advanced voting during the week leading up to Election Day   A new polling station will also be created in a school for registered students at that school only.
> 
> **Caerphilly** – the council offices at Ystrad Mynach will be used as an advanced polling station for all residents of the county on the weekend before Election Day.
> 
> **Torfaen** –the council offices at Pontypool will be used as an advanced polling station for all residents of the county on the weekend before Election Day.

# Tasks

## Storing the data
It's likely that this pilot is continued in future years, and maybe applied to the whole of Wales. Because of this, it makes sense to find a good way of storing and managing the data.

My suggestion is that we manage this through the import scripts, and we attach the advance voting stations via an FK to the _polling station_ model. We should talk more about this, but the idea is that we are likely to want fine grained control of this in future. Some councils will have a single place for everyone, others might have different places depending on the polling station.

I think we need a new model that we can use to store some basic information like point location, address, maybe other details like opening times, directions or accessibility text. TDB depending on some research.

## Presenting the data

We need to present this data directly on the results page, next to the "normal" polling station info. It should be clear that the address is in a pilot area.

We also need to present this in the API and Widget.

- [x] Add model to store Advance Voting venues
- [x] Decide how to manage the data (via existing import scripts?)
- [x] Update front end to show Advance Voting venues on polling station results pages
- [ ] Update translations to support Advance Voting
- [x] Add to WDIV API — decide generic implementation or specific?
- [ ] Update WDIV API Docs
- [x] #4436
- [ ] https://github.com/DemocracyClub/aggregator-api/pull/296
- [ ] [WCIVF]: Update polling station card to show Advance Voting
- [ ] [Widget]: Update Widget

symroe:
> Decide how to manage the data (via existing import scripts?)

We decided to add the data via the import scripts, in a new method. Data comes from an object in the import script

symroe:
[...]

PR #4219, opened 2022-03-07
Add advance voting stations model and API

symroe:
Refs #4181 

Split this PR in two. This one is designed to add the advance polling station model, interface for importing and API responses. 

Another PR will add the frontend interface.


polling_stations/templates/fragments/advance_voting_station.html
```diff
@@ -0,0 +1,12 @@
+<p>{% trans "Your council is trialing a system that allow you to vote before polling day as a specific location." %}</p>
```
GeoWill, line 1: drive by: `s/polling day as a specific location/polling day at a specific location/`

polling_stations/apps/addressbase/management/commands/create_uprn_council_lookup.py
```diff
@@ -48,7 +48,8 @@ def handle(self, *args, **kwargs):
             COPY (SELECT
                     a.uprn as uprn,
                     c.gss as lad,
-                    '' as polling_station_id
+                    '' as polling_station_id,
+                    null as advance_voting_station_id
```
chris48s, line 52: When you re-import this from the CSV, will it actually import it as a `NULL`? (I've not tested it)

polling_stations/apps/addressbase/models.py
```diff
@@ -70,6 +70,12 @@ class Meta:
         db_column="uprn",
     )
     polling_station_id = models.CharField(blank=True, max_length=255)
+    advance_voting_station = models.ForeignKey(
```
chris48s, line 73: Again, not tested - does this FK constraint have any impact on the time it takes to do an addressbase import, or not really?

chris48s (commented):
This is highly exciting :tada:  A few comments from me (that you didn't ask for :smile_cat: ) while I am having a nose about

1. Just looking at these two screenshots, I'm really unclear which of these two addresses corresponds to the pin on the map. I wonder if there should be 2 pins, or 2 maps? (I'm not sure what the answer is)

2. WDIV tends to get a lot of traffic the day before polling day (esp the evening) as people plan where they are going to go tomorrow. Do you want to think about making the switch earlier than polling day itself (when do the advance voting stations close?), or specific messaging for the day before polling day?

symroe:
> Just looking at these two screenshots, I'm really unclear which of these two addresses corresponds to the pin on the map. I wonder if there should be 2 pins, or 2 maps? (I'm not sure what the answer is)

This is good feedback, but just to explain the context a little: we're doing this for the Welsh government, but we've not really talked about presentation or wording at this stage. I expect we'll feed in to the process a lot, but we're going to make sure we align what we do with what the government, councils, campaigns, etc are doing. This is about the lowest effort attempt at putting something visual out there, and it will be changed up a lot, I expect. (not everyone gets excited about a new feature by looking at the data model as we do ;) )

I reckon 2 points on the map is worth investigating, but a couple of things to consider:

1. Some people I've seen use the service have got confused by the home and polling station pins. This might just be because they're looking at _someone else's_ address, and it wouldn't happen if they were looking at their home address, but if there is any confusion then adding another pin might make that worse
2. [3 of the 4 councils only have one advance voting station](https://gov.wales/written-statement-update-flexible-voting-pilots). This means the map would have to be quite zoomed out to include all the points. Of course, there are some long journey times in rural Wales, but I think it's normally going to be true that the advance voting station is further away than the assigned polling station (to be tested, I guess!). So in the "best" case we'd have more or less the same zoom level, but in the "worst" case we'd zoom the map out so much that the directions to the polling station aren't useful. Something we need to explore when we have the actual data, I think.


> WDIV tends to get a lot of traffic the day before polling day (esp the evening) as people plan where they are going to go tomorrow. Do you want to think about making the switch earlier than polling day itself (when do the advance voting stations close?), or specific messaging for the day before polling day?

Yes indeed. See above about aligning messaging with the government etc. Making this experiment less confusing for people has to be our main goal, so figuring out the details is going to be a lot of that work! :)

polling_stations/apps/addressbase/management/commands/create_uprn_council_lookup.py
symroe: Yep!

polling_stations/apps/addressbase/models.py
symroe: I've not tested on a full import (I'm just using the testing data), so we'll have to investigate. My assumption that it would make very little difference. I don't know but I reckon the clever folks at Postgres have optimized for inserting nothing into a nullable column. (not a sarcastic comment...I really think they'll have done a better job at this than we can do)

chris48s:
Yeah all good points. If you have directions, doing it all on one map you'd actually have one 'home' point, 2 'polling station' points and 2 route lines which would probably be quite confusing (esp if the map is zoomed out really far and one line is really short).

Maybe one answer would be not to show any maps to start with but link out to maps/directions (the widget approach) so the user makes the choice of station first then follows a link to a map.

Another way to do it would be to have an intermediate question e.g::

> Your council is doing an early voting trial. When do you want to cast your ballot?
> (radio buttons)
> * Between Date X and Date Y
> * On Date Z

..and then direct them to a result page based on that answer.

chris48s:
Another approach might be show/hide maps e.g:

## Your polling stations

preamble

Polling Station 1
Foo Street
BarTown

<details>
    <summary>show map</summary>
    map div does in here
</details>

---

preamble

Polling Station 2
Baz Street
BarTown

<details>
    <summary>show map</summary>
    map div does in here
</details>

---

(github issue prototyping skillz)

symroe:
Good question, I don't actually know what the thinking behind `polling_stations_known` was. @chris48s do you remember?

symroe:
Thanks Chris that's useful. Let's just go with this here then. If we need a boolean we can add it later, or even add it in the devs.DC API (as it's designed to make our internal APIs more usable)

PR #4536, opened 2022-04-03
Import EONI data

symroe:
## To do before merge

- [x] Docs and custom finder work as outlined in https://github.com/DemocracyClub/UK-Polling-Stations/pull/4536#issuecomment-1086932426
- [x] Test with full data export
- [x] Clean up commit history

### DIFF
--- a/polling_stations/apps/data_importers/management/commands/import_eoni.py
+++ b/polling_stations/apps/data_importers/management/commands/import_eoni.py
@@ -56,6 +56,7 @@ class Command(BaseStationsImporter, CsvMixin):
         ]
         self.stations_only = options.get("stations_only")
         self.pre_process_data()
+        self.clear_old_data()
         self.copy_data()
         self.assign_uprn_to_councils()
         super().handle(*args, **options)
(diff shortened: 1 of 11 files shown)


chris48s:
> Note that to test this in the frontend we need to manually remove the CustomFinder entry for NI. If we continue to get NI data, we might be able to remove the idea of custom finders altogether. But that's for another PR. For now we can just stop importing the data.

IIRC the `CustomFinder` model lives in the local DB, rather than the logger (unless you've moved it) and that record is inserted in a migration: https://github.com/DemocracyClub/UK-Polling-Stations/blob/c5c9c1cd4290bbb0bc67d3552cafbd98c0aae60e/polling_stations/apps/pollingstations/migrations/0013_customfinders.py

When you get to the stage of having real data to import you probably want to remove it again in another migration so it stays removed.

The other thing I'll flag around this is comms/examples around Northern Ireland in API docs, so:

* https://github.com/DemocracyClub/aggregator-api/blob/249c178706c4c57e469aa1f2b90b4bf3546364a2/aggregator/apps/api/v1/docs/notes.apibp#L3-L11
* https://github.com/DemocracyClub/UK-Polling-Stations/blob/faf0ad38d24f78b2d46e45a561fcbe08682db19a/polling_stations/templates/api_docs/blueprints/finder.apibp#L205-L226 - nice and up to date that one! :grimacing:
* https://github.com/DemocracyClub/aggregator-api/blob/270991dfce308e65624dc9ba0c122d602cf5a8dd/aggregator/apps/api/v1/docs/sandbox.apibp#L19
* etc

Should be reviewed in light of the fact that you expect to carry data for NI this election.

polling_stations/apps/data_importers/management/commands/import_eoni.py
```diff
@@ -39,18 +39,28 @@ def add_arguments(self, parser):
...
         self.eoni_export_path = Path(options["eoni_csv"])
         self.eoni_data_root = self.eoni_export_path.absolute().parent
         self.addresses_path = self.eoni_data_root / "eoni_address.csv"
         self.uprn_to_council_path = self.eoni_data_root / "eoni_uprn_to_council.csv"
         self.stations_path = self.eoni_data_root / "eoni_stations.csv"
+        self.paths = [
+            self.addresses_path,
+            self.uprn_to_council_path,
+            self.stations_path,
+        ]
```
symroe, line 56: This could be a dict?

```
self.paths = {
    "addresses": self.eoni_data_root / "eoni_address.csv"
    ...etc
}
```
Then 
```
[path.unlink() for path in self.paths.keys() if path.exists()]
```

GeoWill: :+1:  [7522603](https://github.com/DemocracyClub/UK-Polling-Stations/pull/4536/commits/752260366348511a3559f2297d1f7aa5700756e6)

PR #4556, opened 2022-04-05
Add research question

VirginiaDooley:


VirginiaDooley:
@GeoWill Same issues as yesterday, the `feedback_form` template is not found. I've commented out the remainder of the tests for the time being until I can figure out what changed that would cause this failure to return. Can you spot anything? Very confused.

polling_stations/apps/feedback/tests/test_feedback.py
```diff
@@ -0,0 +1,77 @@
...
+#             geography__geography=None,
+#         )
+
+#         call_command(  # Hack to avoid converting all fixtures to factories
+#             "loaddata",
+#             "polling_stations/apps/data_finder/fixtures/test_single_address_single_polling_station.json",
+#             verbosity=0,
+#         )
+
+#     def test_feedback_form(self):
```
GeoWill, line 28: I think if you add `@override_settings(EVERY_ELECTION={"CHECK": False, "HAS_ELECTION": True})` as a decorator on the method it should work

GeoWill:
I think https://github.com/DemocracyClub/UK-Polling-Stations/blob/c80424ca0dd8d748f7da167a588582ee94986da0/polling_stations/apps/feedback/views.py#L34 will also need to be updated so that the feedback objects are correctly created. I'm a bit hazy on how the form->model creation process works though, so might be wrong about this

polling_stations/apps/feedback/tests/test_feedback.py
VirginiaDooley: That worked, thanks!

GeoWill (approved):
This LGTM, and I can leave feedback locally which all works fine.

michaeljcollinsuk (approved):

PR #4653, opened 2022-04-14
Advance voting frontend

symroe:
There are four cases this PR deals with:

# 1. No AVS

As per normal, but we move the header out of a card
Screenshot from 2022-04-14 15-19-34

# 2. Advance station, open in the future

This is also the display on election day, if the AVS is open on election day

Screenshot from 2022-04-14 15-19-57

# 3. AVS closed, polls open

Screenshot from 2022-04-14 15-20-36

# 4 Polls closed
No change.


symroe:
I don't quite know what the deal is when the AVS is the same as the polling station. @GeoWill does this happen, and do we have any way of knowing they're the same?

I guess it'll just not show on polling day, as we'll not mark the stations as open on polling day, if they're the same?

polling_stations/templates/fragments/polling_station_known.html
```diff
@@ -8,54 +8,97 @@ <h2>
     {% endblocktrans %}
 </h2>
 
-{% if not show_polls_open_card and advance_voting_station %}
-    <h3>{% trans "Vote in advance" %}</h3>
+{% if not show_polls_open_card and advance_voting_station and advance_voting_station.open_in_future %}
+    <p>
+        {% if NEXT_CHARISMATIC_ELECTION_DATE %}
+            {% blocktrans trimmed %}
+                Your council is trailing a system that allows you to vote before polling day.
```
VirginiaDooley, line 15: `trialing`?

polling_stations/templates/fragments/polling_station_known.html
```diff
@@ -8,54 +8,97 @@ <h2>
...
 
-{% if show_polls_open_card and advance_voting_station %}
-    <h3>{% trans "Alternative polling station" %}</h3>
-    {% include "fragments/advance_voting_station.html" %}
+{% if show_polls_open_card and advance_voting_station and not advance_voting_station.open_in_future %}
+    <div class="ds-card">
+        <div class="ds-card-body">
+            {% blocktrans %}
+                <h3>Advance voting trial</h3>
+                <p>Your council conduced a trial that allowed you to vote before polling day.</p>
```
VirginiaDooley, line 98: `conducted`?

GeoWill:
> I don't quite know what the deal is when the AVS is the same as the polling station. @GeoWill does this happen, and do we have any way of knowing they're the same?

This is happening in Bridgend this year. The only way we have of knowing is looking at the script: https://github.com/DemocracyClub/UK-Polling-Stations/pull/4491/commits/abb00a3242c9153db261da77d7b6ff17c2292451
 
> I guess it'll just not show on polling day, as we'll not mark the stations as open on polling day, if they're the same?

Exactly - I think they stop being an AVS on polling day.

polling_stations/templates/fragments/polling_station_known.html
```diff
@@ -8,54 +8,97 @@ <h2>
...
+                <p>The advance voting station are closed now, and you must vote at your assigned polling station.</p>
```
VirginiaDooley, line 99: *`is now closed`

GeoWill:
> This is also the display on election day, if the AVS is open on election day

I think its OK to think about this case but fwiw I don't think it's happening this year.

polling_stations/apps/pollingstations/context_processors.py
```diff
@@ -10,7 +12,7 @@ def google_analytics(request):
     settings, "NEXT_CHARISMATIC_ELECTION_DATE", None
 )
 NEXT_CHARISMATIC_ELECTION_DATE = (
-    datetime.strptime(NEXT_CHARISMATIC_ELECTION_DATE, "%Y-%m-%d")
+    datetime.strptime(NEXT_CHARISMATIC_ELECTION_DATE, "%Y-%m-%d").date()
```
GeoWill, line 15: Will this affect usage in templates: https://github.com/DemocracyClub/UK-Polling-Stations/blob/eaaa1a23eadb0cb1844222fbacbc099b7bf869a6/polling_stations/templates/base_full.html#L13

symroe: If anything it'll make it better. For some reason I couldn't get Django to parse the date using `|date` when it was a `datetime` object. Tested locally and that line is working.

GeoWill:
I basically think this is great. I did wonder what @chris48s suggestion of using `<detail>` tags would look like though. I quite like it, as (especially on mobile) without the 'Vote on polling day' card is a long way down, and scrolling past a map can be annoying (you just pan the map). What do you think? 

Screenshot from 2022-04-15 12-14-56

symroe:
Screenshot from 2022-04-19 12-19-24

Updated header to link to different cards

polling_stations/templates/fragments/polling_station_known.html
```diff
@@ -8,54 +8,98 @@ <h2>
...
+        {% else %}
+            {% blocktrans trimmed %}
+                Your council is trialling a system that allows you to vote in person before polling day.
+                You can <a href="#vote-in-advance">vote in advance</a>
+                at this location, or <a href="#vote-on-polling-day">vote at your polling station</a>
+                as normal on&nbsp;polling day.
+            {% endblocktrans %}
+        {% endif %}
+    </p>
     {% include "fragments/advance_voting_station.html" %}
```
GeoWill, line 29: This fragment is only included in `polling_station_known.html`. However it's possible we will have assigned an advance station to an address, but not a polling station - e.g. when every address in a council can vote at a single advance station. However I think this could end up being really confusing to users, so I'm not sure what's best - but realised I hadn't previously considered it - maybe you had?

GeoWill (commented):
Nearly there I think. Can't spot anymore typos, but think the station address formatting needs some thought - or specific guidance on how it should be formatted in the import scripts. Also a query about when advance station is known, but polling station isn't.

polling_stations/templates/fragments/polling_station_known.html
symroe: My feeling is that we should keep it simple: if we don't know the polling station, we don't know about advance stations. Not ideal, I know, but I think it's the only way to keep it simple.

We also don't show the advance station in the API if the station isn't known, I think

polling_stations/templates/fragments/advance_voting_station.html
```diff
@@ -1,18 +1,28 @@
...
-</address>
+<div class="ds-card" id="vote-in-advance">
+    <div class="ds-card-body">
+        <h3>{% trans "Vote in advance" %}</h3>
+        <address>
+            {{ advance_voting_station }}
+            {{ advance_voting_station.name }}<br>
+            {{ advance_voting_station.address|linebreaksbr }}
+            {{ advance_voting_station.postcode }}
+        </address>
```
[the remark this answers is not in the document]
GeoWill: Yep. That sorts it out. :+1:

polling_stations/templates/fragments/polling_station_known.html
GeoWill: We don't show the advance station on a postcode view, if the postcode is split between polling station known and unknown (even if all addresses go to the same advance station). However if you go to the `address` endpoint for a uprn where we don't know the polling station, then you will get an advance station if we know it. This is also true when none of the addresses in the postcode have a known polling station - however this is less plausible, because we don't return a list of uprns.

GeoWill: However I agree that we shouldn't show it on the front end.

PR #4676, opened 2022-04-20
Log all postcode lookups

symroe:
I'm not logging postcodes that look wrong, and at the moment we only care about postcodes, not addresses. 

We also don't care about API postcodes, as we'll assume they're logged via devs.DC.

Might be of use when working on https://github.com/DemocracyClub/WhoCanIVoteFor/issues/1166

### DIFF
--- a/polling_stations/apps/data_finder/views.py
+++ b/polling_stations/apps/data_finder/views.py
@@ -222,6 +222,14 @@ class PostcodeView(BasePollingStationView):
         if "postcode" not in kwargs or kwargs["postcode"] == "":
             return HttpResponseRedirect(reverse("home"))
 
+        # Log to firehose
+        entry = settings.POSTCODE_LOGGER.entry_class(
+            postcode=kwargs["postcode"],
+            dc_product=settings.POSTCODE_LOGGER.dc_product.wdiv,
+            **self.request.session.get("utm_data"),
+        )
+        settings.POSTCODE_LOGGER.log(entry)
+
         rh = RoutingHelper(self.kwargs["postcode"])
 
         if rh.view != "postcode_view":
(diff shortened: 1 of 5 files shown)


polling_stations/apps/data_finder/views.py
```diff
@@ -222,6 +222,14 @@ def get(self, request, *args, **kwargs):
...
             return HttpResponseRedirect(reverse("home"))
 
+        # Log to firehose
+        entry = settings.POSTCODE_LOGGER.entry_class(
+            postcode=kwargs["postcode"],
+            dc_product=settings.POSTCODE_LOGGER.dc_product.wdiv,
+            **self.request.session.get("utm_data"),
+        )
+        settings.POSTCODE_LOGGER.log(entry)
+
```
GeoWill, line 232: I assume the reason this is going here rather than the `LogLookUpMixin` class is so that we only log postcodes. In which case it makes sense. Once we're happy with firehose I guess we can extract this out if we want.

polling_stations/settings/base.py
```diff
@@ -294,6 +295,14 @@
...
 SHOW_ADVANCE_VOTING_STATIONS = True
 
+# DC Logging Client
+FIREHOSE_ACCOUNT_ARN = os.environ.get("FIREHOSE_ACCOUNT_ARN", None)
+if FIREHOSE_ACCOUNT_ARN:
+    firehose_args = dict(assume_role_arn=FIREHOSE_ACCOUNT_ARN)
+else:
+    firehose_args = dict(fake=True)
+POSTCODE_LOGGER = DCWidePostcodeLoggingClient(**firehose_args)
+
```
GeoWill, line 305: Is this handling local development - i.e. when not deployed with polling_deploy? 
If this is the case should it go in `local.py.example`? Then if someone adds the ARN to their settings rather than using and ENV variable they're less likely to check it in?

PR #4804, opened 2022-07-26
Hotfix/base template

VirginiaDooley:
Closes https://github.com/DemocracyClub/dc_django_utils/issues/28, https://github.com/DemocracyClub/dc_django_utils/pull/29, https://github.com/DemocracyClub/UK-Polling-Stations/issues/4815

This work:

- integrates changes in the base template from `dc-utils`, bringing back the default structure that existed with `dc_base_theme` before we changed to `dc_design_system`.
- Replaces `markdown_deux_tags` which is no longer compatible with `Django>=4.0`
- includes an upgrade to Django 4.1 and Python 3.8.12


VirginiaDooley:
[Match og_tags to dc_utils](https://trello.com/c/oyrIDWEA/3048-match-ogtags-to-dcutils)

Bekabyx:
It looks like this set of changes has broken `test_get_bug_report_view`  - is this a known issue or just an oopsie?

VirginiaDooley:
> It looks like this set of changes has broken `test_get_bug_report_view` - is this a known issue or just an oopsie?

I think it should be fine now. That error was based on changes to `dc_utils` which has since been upgraded.

requirements/base.txt
```diff
@@ -1,32 +1,32 @@
...
-django-apiblueprint-view==2.3.2
+dj-database-url==1.0.0
+django-apiblueprint-view
 django-basicauth==0.5.3
-django-extensions==3.1.5
+django-extensions==3.2.1
 django-localflavor==3.1
-django-markdown-deux==1.0.5
-django==3.2.13
+markdown==3.3.7
```
VirginiaDooley, line 7: I made this change bc `django-markdown-deux==1.0.5` was no longer being maintained and was not compatible with `Django>4.0`. As I'm making the same change to WCIVF, I'm only now realising that I'm not sure this package is actually being used.

Bekabyx (approved):

requirements/base.txt
chris48s: AFAIK There are 2 places it is used in WDIV - for rendering the explainers and cancelled election notifications in
https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/templates/election_explainers.html
and
https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/templates/fragments/cancelled_election.html
but it looks like you've caught them both in this PR anyway.

polling_stations/apps/core/opening_times.py
```diff
@@ -1,9 +1,8 @@
 import datetime
+from collections import OrderedDict
 from datetime import date, time
 from typing import Dict, List
 
-from typing_extensions import OrderedDict
-
```
GeoWill, line 6: @symroe `typing_extensions` weren't available in 3.10, I could've imported `OrderedDict` from the `typing` lib, but IIUC we're instantiating an new instance of the class, rather than asserting the type in line 11 (`self.dates: Dict[date, List] = OrderedDict()`) and so we want the class rather than the type in the import? FWIW it works either way, which makes me think I don't really understand what's going on.

PR #4823, opened 2022-09-14
Format import script PRs with Black

symroe:
Some import scripts made via the uploader fail CI initially because they're not valid in Black's eyes.

This PR passes the generated script through Black before it creates the PR


GeoWill (approved):
Moved `black` into `requirements/base.txt`

PR #4874, opened 2022-11-21
Improve Circleci caching

GeoWill:


.circleci/config.yml
```diff
@@ -45,16 +50,28 @@ jobs:
             pip install -r requirements/testing.txt
             playwright install
 
+      - run:
+          name: Install HTML Tidy
+          command: |
+            mkdir -p ~/tidy
+            wget --directory-prefix /home/circleci/tidy https://github.com/htacg/tidy-html5/releases/download/5.8.0/tidy-5.8.0-Linux-64bit.deb && sudo dpkg -i /home/circleci/tidy/tidy-5.8.0-Linux-64bit.deb
```
symroe, line 57: I think you want to pass `-nc` to wget, otherwise it will always replace the file, and make caching it pointless.

PR #4950, opened 2023-01-25
Remove ID messaging for En/Sc/Cy

Bekabyx:
New legislation = new messaging :grin:

NI messaging around voter ID is unchanged but messaging for the other nations has been removed - potentially to be replaced by new messaging when we have that figured out. 

Screenshot 2023-01-25 at 11 41 55
Screenshot 2023-01-25 at 11 50 36

### DIFF
--- a/polling_stations/templates/home.html
+++ b/polling_stations/templates/home.html
@@ -48,8 +48,6 @@
                         <p>
                             {% trans "If you are registered to vote, but you don't have your poll card, you can go to the polling station and give them your name and address." %}
                         </p>
-
-                        <p>{% trans "In Northern Ireland, you must bring photo ID." %}</p>
                     </div>
                 </div>
             {% endif %}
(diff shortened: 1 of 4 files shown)


VirginiaDooley (approved):

polling_stations/templates/home.html
```diff
@@ -49,7 +49,7 @@ <h2>{% blocktrans with election_date=election_date|date:"DATE_FORMAT" context="S
                             {% trans "If you are registered to vote, but you don't have your poll card, you can go to the polling station and give them your name and address." %}
                         </p>
 
-                        <p>{% trans "In England, Wales and Scotland, you don't need any form of ID. In Northern Ireland, you must bring photo ID." %}</p>
+                        <p>{% trans "In Northern Ireland, you must bring photo ID." %}</p>
```
symroe, line 52: Actually could you just remove this line too, until we've planned what we're going to say? to me this still implies that you _don't_ need ID in other places. Do you see what I mean?

Bekabyx: Totally understand, it's kind of a weird limbo where we don't have a complete idea of new messaging yet and this could be misleading. Removed in b522ce5

symroe (approved):

Issue #4476, opened 2022-03-25
Notify us if file is 'pending' for longer than about 5 minutes


pmk01:
'Pending' is a bit of a blind spot for us - we only get notifications if upload succeeds or fails.

VirginiaDooley:
@pmk01  Preferred method of notification? Email, slack or ?

pmk01:
Email please!

PR #4953, opened 2023-01-26
Pending Upload Notifications

VirginiaDooley:
Closes https://github.com/DemocracyClub/UK-Polling-Stations/issues/4476

This work creates a management command to check for pending uploads (> 20 mins) then sends and logs an email (per upload) in those cases, if they exist.


VirginiaDooley:

polling_stations/apps/file_uploads/tests/test_check_pending.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertEqual(Upload.objects.first().warning_about_pending_sent, True)
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertEqual(mail.outbox[0].subject, "Possible failed uploads")
+        self.assertEqual(
+            mail.outbox[0].body, f"A file(s) upload was not successful: {qs}"
+        )
+        self.assertEqual(mail.outbox[0].to, [settings.DEFAULT_FROM_EMAIL])
+
+    @freeze_time("2021-04-01 12:03:00")
+    def test_handle_nothing_pending(self):
```
GeoWill, line 34: This name is confusing because there _is_ an upload pending, but it's within the allowed time limit.

polling_stations/apps/file_uploads/management/commands/check_pending.py
```diff
@@ -0,0 +1,31 @@
...
+from django.conf import settings
+from django.utils.timezone import now
+from file_uploads.models import Upload
+from django.core.mail import EmailMessage
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs):
+        from_time = now() - timedelta(minutes=5)
```
GeoWill, line 11: Is there a reason for the 5 minute cut off. I guess I expect it's enough, but if we started doing more stuff it might not be. given we'll probably run this management command hourly I'd be more inclined to have a 15 minute cut off, to reduce the chance of false positives

polling_stations/apps/file_uploads/management/commands/check_pending.py
```diff
@@ -0,0 +1,31 @@
...
+        )
+
+        if len(qs) == 0:
+            print("No failed uploads")
+        else:
+            qs.update(warning_about_pending_sent=True)
+
+            email = EmailMessage(
+                "Possible failed uploads",
+                f"A file(s) upload was not successful: {qs}",
```
GeoWill, line 23: `qs` might contain more than one upload. Will this be confusingly formatted?

polling_stations/apps/file_uploads/management/commands/check_pending.py
VirginiaDooley: @pmk01 Any reason for 5 mins?

symroe: I agree. I picked 5 minutes out of the air, as the current Lambda should have finished within that time.

The [function timeout is 15 minutes](https://github.com/DemocracyClub/wdiv-s3-trigger/blob/master/serverless.yml#L10) though, so we should set this check to `$timeout + [some padding]`

polling_stations/apps/file_uploads/management/commands/check_pending.py
```diff
@@ -0,0 +1,31 @@
...
+from django.utils.timezone import now
+from file_uploads.models import Upload
+from django.core.mail import EmailMessage
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs):
+        from_time = now() - timedelta(minutes=5)
+        qs = Upload.objects.filter(timestamp__lte=from_time).filter(
```
symroe, line 12: Could you add this to [the Upload manger](https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/file_uploads/models.py#L28) as a new method?

This is because it's better to have this query defined in one place (on the model), so we can reuse it in other parts of the project, should we want to (for example, if we decide to show a page with them all on)

polling_stations/apps/file_uploads/management/commands/check_pending.py
```diff
@@ -0,0 +1,31 @@
...
+            warning_about_pending_sent=False
+        )
+
+        if len(qs) == 0:
+            print("No failed uploads")
+        else:
+            qs.update(warning_about_pending_sent=True)
```
symroe, line 19: I think this needs a little bit of rework, depending on how you plan to deal with Will's comment. 

## Options 1

If we are going to send a single email for _all_ uploads in the qs, then we should do something like:

```
qs = Upload.objects.stuck_in_pending()
if not qs:
    self.stdout.write("No uploads found")
    return

[send email]
```

# Option 2

To send one email per upload in the qs:

```
qs = Upload.objects.stuck_in_pending()
for upload in qs:
    [email]
else:
    self.stdout.write("No uploads found")
```

(Note `self.stdout.write` rather than `print` as it's better for testing)


However, I'm not sure we need to print that no uploads are found, as that will be the common case and don't think we need that filling up our logs.

symroe (commented):
Inline comments are getting a bit messy, so I'll add this one here:

You're currently doing:

1. get the QS
2. Update the QS to say emails have been sent
3. sending emails

If the email sending step fails, then we'll have updated the model incorrectly.

There are two things we can do to fix this, and I think we should do both:

1. Wrap the calling function in `@transaction.atomic`. This means that any changes to the database are only written when the function / method actually completes without error
2. Call the update on the model after the email has been sent.

You can do this in a refactor by adding the email sending code to the model (as per https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/file_uploads/models.py#L163), updating the model _within_ that method, and decorating the method with `transaction.atomic`.

Happy to chat about this if I'm not making sense.

polling_stations/apps/file_uploads/management/commands/check_pending.py
pmk01: Yep, no reason for 5 minutes, can be any length of time.

polling_stations/apps/file_uploads/management/commands/check_pending.py
```diff
@@ -0,0 +1,13 @@
...
+
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs):
+        qs = UploadQuerySet.pending_upload_qs(self)
+        if len(qs) == 0:
+            pass
+        else:
+            for upload in qs:
+                upload.send_error_email()
+        return qs
```
GeoWill, line 13: I don't think there's any harm in this, but it looks like the return value is mostly there for testing. In general it would be better to test the `pending_upload_qs` method on the `Upload` model's test case. The obvious issue with that is there's no tests for the `Upload` model. (:raised_hand_with_fingers_splayed: guilty). So it's fine like this for now, but something to come back to.

GeoWill (commented):
I think this looks good, one comment to fix an issue from previous PR - that can happen here or in another PR.

polling_stations/apps/file_uploads/management/commands/check_pending.py
VirginiaDooley: I was tempted to add more tests but didn't want to go too far out of scope. Happy to do that in another PR and adjust this method as suggested then.

polling_stations/apps/file_uploads/models.py
```diff
@@ -160,6 +170,7 @@ def send_confirmation_email(self, user):
 
         return email.send()
 
+    @transaction.atomic
     def send_error_email(self):
         subject = "File upload failed"
         message = f"File upload failure: {self.__str__}. Please investigate further."
```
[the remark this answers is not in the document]
VirginiaDooley: Fixed with c86684f23688c8e2fa844c36cbaa0c50c933601a

polling_stations/apps/file_uploads/tests/test_check_pending.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertEqual(len(mail.outbox), 2)
+        self.assertEqual(mail.outbox[0].subject, "File upload failed")
+        self.assertEqual(mail.outbox[1].subject, "File upload failed")
+        self.assertEqual(
+            mail.outbox[0].body,
+            f"File upload failure: {upload1}. Please investigate further.",
+        )
+        self.assertEqual(
+            mail.outbox[1].body,
+            f"File upload failure: {upload2}. Please investigate further.",
```
GeoWill, line 48: I would hard code these strings - this is why there wasn't a failure when the method was using `self.__str__` - if you hard code the expected string and someone changes the `__str__` method then they'll know what's currently expected.

GeoWill (approved):
One small inliner. Then good to squash to a single commit and merge.

Issue #3051, opened 2021-02-25
Only report validation when all files are processed


pmk01:
Democracy Counts sends two files. When uploading these, the uploader checks the smaller file first, and presents you with an ❌. Only after a further refresh does it check both files, and switch to ✔

pmk01:
Related: sometimes the polling stations file is smaller than 1kb (especially for by-elections), which throws an error.

symroe:
This might be tricky as I don't know how much each Lambda process knows at the time it's processing a file. We might want to deal with this in the interface, and only show the validation if we have a response for all files.

symroe:
See #4929 for the 1kb / by-election comment above.

chris48s:
> This might be tricky as I don't know how much each Lambda process knows at the time it's processing a file

Basically, this is why status is reported at a file level. The lambda trigger is a `s3:ObjectCreated` event, so all lambda knows is that a file just appeared. It doesn't know whether it is supposed to have a friend or not. WDIV doesn't receive the files first either - they're uploaded straight from the client to S3. This is a good thing because it means WDIV doesn't have to receive a >100Mb POST request but this is the tradeoff.

As suggested, you may be able to handle this a bit more intelligently back on WDIV when rendering the admin template. Something to the effect of (pseudo-code):

```
if EMS == "democracy counts" and file_count == 1:
   if uploaded_date < [30 seconds ago]:
       status = "⌛ waiting for second file"
    else:
       status = "❌ only one file uploaded"
```

or whatever.

PR #4955, opened 2023-01-30
Set status based on EMS and file count

VirginiaDooley:
Closes https://github.com/DemocracyClub/UK-Polling-Stations/issues/3051
Ref https://trello.com/c/M4uUNZMl/3229-priority-wdiv-uploader-fixes-needed-before-next-election

- This work adds a status check to each file in the file set during the upload process and returns a more specific status message to the user.
- I've also added a server environment check for prod to send custom error emails as not to confuse anyone in Fresh Desk while we are testing the uploader. 
- Lastly, I've updated the node version in CI as the previous orb was no longer in use. 

Changes can be tested by uploading 1/2 and/or 2/2 EMS files such as http://dev.wdiv.club/uploads/councils/BDF/

Screenshot 2023-02-27 at 3 51 07 PM

REMINDER! Drop this commit once reviewed and approved [12bf87b](https://github.com/DemocracyClub/UK-Polling-Stations/pull/4955/commits/12bf87bbfa9717f9ed5236eb8a59af948ccbac42)

### DIFF
--- a/polling_stations/apps/file_uploads/models.py
+++ b/polling_stations/apps/file_uploads/models.py
@@ -16,7 +16,9 @@ from data_importers.import_script import ImportScript
 
 status_map = {
     "Pending": "⌛",
+    "Waiting": "⌛ waiting for second file",
     "Error": "❌",
+    "Error One File": "❌ only one file uploaded",
     "OK": "✔️",
 }
 
@@ -64,6 +66,15 @@ class Upload(models.Model):
         if not self.file_set.all():
             return "Pending"
         for f in self.file_set.all():
+            if ("Expected 2 files, found 1" in f.errors) and (
+                self.timestamp + timedelta(seconds=180) > now()
+            ):
+                return "Waiting"
+            elif ("Expected 2 files, found 1" in f.errors) and (
+                self.timestamp + timedelta(seconds=180) < now()
+            ):
+                return "Error One File"
+
             if not f.csv_valid:
                 return "Error"
         return "OK"
--- a/polling_stations/apps/file_uploads/templates/file_uploads/council_detail.html
+++ b/polling_stations/apps/file_uploads/templates/file_uploads/council_detail.html
@@ -46,14 +46,14 @@
                             {% endif %}
                         </td>
                         <td>
-                            {{ upload.status }}
-                            {% if upload.status == "Pending" %}
+                            {{ upload.status_emoji }}
+                            {% if upload.status == "Pending" or upload.status == "Waiting" %}
                                 <br><small>(Refresh page to update)</small>
                             {% endif %}
                         </td>
                         <td>
                             {% if upload.status != "Pending" %}
-                                {% if STATIONS  %}✔{% else %}❌{% endif %}
+                                {% if STATIONS %}✔{% else %}❌{% endif %}
                             {% endif %}
                         </td>
                     </tr>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
           key: v1-polling_stations-dependencies-{{ checksum "requirements/base.txt" }}-{{ checksum "requirements/testing.txt" }}-{{ checksum "requirements/cdk.txt" }}
 
       - node/install:
-          node-version: '10.19.0'
+          node-version: '16.13.0'
 
       - run:
           name: Install node node_modules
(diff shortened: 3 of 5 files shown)


VirginiaDooley:

polling_stations/apps/file_uploads/models.py
```diff
@@ -56,7 +56,29 @@ def status(self):
         for f in self.file_set.all():
             if not f.csv_valid:
                 return "Error"
-        return "OK"
+            else:
+                return self.ems_check(f)
```
GeoWill, line 60: The else isn't needed here.

polling_stations/apps/file_uploads/models.py
```diff
@@ -56,7 +56,29 @@ def status(self):
...
+
+    def ems_check(self, file):
+        wait_time = now() - timedelta(seconds=30)
```
GeoWill, line 63: I think 30 seconds is too short. Lambda can take 10-15 seconds, and ems files can be 5-15mb. I think we can't rely that much on council upload speeds.

polling_stations/apps/file_uploads/models.py
```diff
@@ -56,7 +56,29 @@ def status(self):
...
+        if file.ems != "Democracy Counts":
+            pass
```
GeoWill, line 65: I think this means we'll never return `OK` for other EMSs.

polling_stations/apps/file_uploads/models.py
```diff
@@ -56,7 +56,29 @@ def status(self):
         for f in self.file_set.all():
             if not f.csv_valid:
```
GeoWill, line 57: I think if there's a single file in the fileset, and the ems is Democracy Counts, then the lambda handler will have set `csv_valid` to `False`  and added `"Expected 2 files, found 1"` to the `errors` field on the existing file.
So I think you could just look at `f.errors` here:
```python
for f in self.file_set.all():
    if ("Expected 2 files, found 1" in f.errors) and (self.timestamp + timedelta(seconds=180) > now()):
        return "Waiting"
    elif ("Expected 2 files, found 1" in f.errors) and (self.timestamp + timedelta(seconds=180) < now()):
        return "Error One File"
    
    if not f.csv_valid:
        return "Error"
return "OK"
```

GeoWill (changes requested):
This is quite a confusing area, because it's not clear how different fields on the model are getting set. I haven't completely got it in my head, so my suggestion might not be quite right. FWIW I found reading the trigger code helpful to see what might be happening.

polling_stations/apps/file_uploads/models.py
VirginiaDooley: Fixed with https://github.com/DemocracyClub/UK-Polling-Stations/pull/4955/commits/10c545be345bbd157df9c4b8ce7c08db57f08d96

polling_stations/apps/file_uploads/models.py
VirginiaDooley: Fixed with https://github.com/DemocracyClub/UK-Polling-Stations/pull/4955/commits/10c545be345bbd157df9c4b8ce7c08db57f08d96

polling_stations/apps/file_uploads/models.py
VirginiaDooley: Fixed with https://github.com/DemocracyClub/UK-Polling-Stations/pull/4955/commits/10c545be345bbd157df9c4b8ce7c08db57f08d96

polling_stations/apps/file_uploads/templates/file_uploads/council_detail.html
```diff
@@ -46,14 +46,11 @@ <h2>Uploads</h2>
                             {% endif %}
                         </td>
                         <td>
-                            {{ upload.status }}
-                            {% if upload.status == "Pending" %}
-                                <br><small>(Refresh page to update)</small>
-                            {% endif %}
```
GeoWill, line 52: I think it would be nice to keep the note about refreshing the page - but check for `Waiting` as well as `Pending`.

GeoWill (approved):
Small change to keep the note about refreshing. Otherwise I think this looks good to go, once the temp deploy commit is dropped.

polling_stations/apps/file_uploads/templates/file_uploads/council_detail.html
```diff
@@ -47,6 +47,9 @@ <h2>Uploads</h2>
                         </td>
                         <td>
                             {{ upload.status_emoji }}
+                            {% if upload.status == "Pending" or "Waiting" %}
```
GeoWill, line 50: I don't think this does what you want. I think it should be: `if upload.status == "Pending" or upload.status =="Waiting"`.

GeoWill:
Good to merge I think :+1:

Issue #4220, opened 2022-03-07
Assign councils to domain based on more than the councils email address


pmk01:
I would like the ability to assign councils to any domain - for example, Wandsworth and Richmond staff use @wandsworthandrichmond.gov.uk domains for their personal emails. At present the user has to create an account first, and then we attach it to them, which is hard to explain.

pmk01:
Good questions. Not a lot, I suspect. If we need to punt any uploader work, this is probably it.

symroe:
Ok, I suggest we do it in code for now then, as that's a quick win. Can you make a list of domain -> council ID for each non-default domain you know about, and we'll add that to the logic that assigns users to councils?

symroe:
Is this done?

VirginiaDooley:
> Is this done?

Yes I believe so. Completed with https://github.com/DemocracyClub/UK-Polling-Stations/pull/4958

PR #4958, opened 2023-01-31
Assign custom email domains to user

VirginiaDooley:
Ref https://github.com/DemocracyClub/UK-Polling-Stations/issues/4220

This work adds a check and alternative route to `assign_council_to_user` when the user wants to login with an email address that contains a domain which differs from the council website domain. 

As noted in the comments, this is a quick fix for the few cases where this need arises; A PR to customise the list will be required as this functionality is not accessible from the admin site. I've added this instruction inline and to the README.

I've tested the following scenarios in the dev server: 

- Login with a standard domain email and be redirected to the matching council
- Login with a custom domain email and be redirected to the designated council
- Try to login with an email address that is neither custom or standard (error)

I've testing the following scenarios on the staging server: 

- Login with a custom domain email and be redirected to the designated council
- Try to login with an email address that is neither custom or standard (error)

### DIFF
--- a/polling_stations/apps/file_uploads/utils.py
+++ b/polling_stations/apps/file_uploads/utils.py
@@ -3,10 +3,31 @@ import os
 from django.db.models import Q
 
 from councils.models import Council, UserCouncils
+from polling_stations.settings.constants.uploads import CUSTOM_DOMAINS
 
 
 def get_domain(request):
     return os.environ.get("FQDN", request.META.get("HTTP_HOST"))
+
+
+# This is a quick fix: Raise a PR to
+# update CUSTOM_DOMAINS in
+# polling_stations/settings/constants/uploads.py
+# with the the actual list of email domains
+# and councils, as the need arises.
+# TO REMOVE:
+# 1) Delete the entry from the CUSTOM_DOMAIN list
+# 2) In the admin panel, delete the UserCouncil
+# object from the bottom of the Council profile page
+
+
+def assign_approved_custom_domains(user, email_domain):
+    # if email_domain in custom_domains, assign it to the
+    # corresponding council, otherwise do nothing
+    if email_domain := CUSTOM_DOMAINS.get(email_domain, None):
+        for council_id in email_domain:
+            council = Council.objects.get(council_id=council_id)
+            UserCouncils.objects.update_or_create(user=user, council=council)
 
 
 def assign_councils_to_user(user):
@@ -14,6 +35,7 @@ def assign_councils_to_user(user):
     Adds rows to the join table between User and Council
     """
     email_domain = user.email.rsplit("@", 1)[1]
+    assign_approved_custom_domains(user, email_domain)
     councils = Council.objects.filter(
         Q(electoral_services_email__contains=email_domain)
         | Q(registration_email__contains=email_domain)
--- a/polling_stations/settings/constants/uploads.py
+++ b/polling_stations/settings/constants/uploads.py
@@ -10,3 +10,11 @@ GITHUB_USERNAME = "polling-bot-4000"
 GITHUB_EMAIL = "developers@democracyclub.org.uk"
 GITHUB_REPO = "DemocracyClub/UK-Polling-Stations"
 GITHUB_API_KEY = os.environ.get("GITHUB_API_KEY", "")
+
+# update this constant with email domains
+# (i.e. the bit after '@') and the corresponding
+# three letter council_ids in a list as shown below:
+CUSTOM_DOMAINS = {
+    # "example.com": ["EXC"],
+    # "example.org": ["EXO", "EXP"],
+}
--- a/README.md
+++ b/README.md
@@ -156,3 +156,9 @@ If you are interested in helping the project by writing an import script, see th
 ## Install git hooks
 
 If you like you can use the commit hooks defined in `.pre-commit-config.yaml`. Run `pre-commit install && pre-commit install -t pre-push`.
+
+## Custom council domains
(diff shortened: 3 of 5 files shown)


VirginiaDooley:

polling_stations/apps/file_uploads/utils.py
```diff
@@ -12,17 +12,58 @@ def get_domain(request):
...
+def approved_custom_domain(user, email_domain):
+    # This is a quick fix: Peter can raise a
+    # PR to update this list with the the actual
+    # list of email domains and councils as the
+    # need arises.
+
+    custom_domains = {
+        "custom_council.gov.uk": "Foo Council",
+        "custom_council2.gov.uk": "Bar Council",
+    }
```
VirginiaDooley, line 24: Note: Use the `council.gss` with 3 letter code values rather than the `council.name`

polling_stations/apps/file_uploads/utils.py
```diff
@@ -12,17 +13,43 @@ def get_domain(request):
...
 
+# This is a quick fix: Peter can raise a
+# PR to update CUSTOM_DOMAINS with the the actual
+# list of email domains and councils as the
+# need arises.
+def approved_custom_domain(user, email_domain):
+    # if email_domain in custom_domains, assign it to the
+    # corresponding council and return True
+    if email_domain not in CUSTOM_DOMAINS:
+        pass
```
GeoWill, line 24: I think it would be clearer to `return False` here

polling_stations/settings/constants/uploads.py
```diff
@@ -3,3 +3,8 @@
 GITHUB_USERNAME = "polling-bot-4000"
 GITHUB_EMAIL = "developers@democracyclub.org.uk"
 GITHUB_REPO = "DemocracyClub/UK-Polling-Stations"
+
+CUSTOM_DOMAINS = {
+    "custom_council.gov.uk": "FOO",
+    "custom_council2.gov.uk": "BAR",
```
GeoWill, line 9: This should be empty with a comment to explain the expected format (email domain (i.e. the bit after '@') to reg code?) and how to use it preferably with a real world example.

polling_stations/apps/file_uploads/utils.py
```diff
@@ -12,17 +13,43 @@ def get_domain(request):
...
+    else:
+        # get an instance of the council described in the custom_domains dict
+        council = Council.objects.using(LOGGER_DB_NAME).get(
+            name=CUSTOM_DOMAINS[email_domain]
```
GeoWill, line 28: `council_id=`?

GeoWill (changes requested):

polling_stations/apps/file_uploads/utils.py
```diff
@@ -12,17 +13,43 @@ def get_domain(request):
...
     email_domain = user.email.rsplit("@", 1)[1]
-    councils = Council.objects.using(LOGGER_DB_NAME).filter(
-        Q(electoral_services_email__contains=email_domain)
-        | Q(registration_email__contains=email_domain)
-    )
-
-    for council in councils:
-        UserCouncils.objects.using(LOGGER_DB_NAME).update_or_create(
-            user=user, council=council
+    if approved_custom_domain(user, email_domain):
```
symroe, line 41: I _think_ we might want to call both methods in all cases…

If someone at council A manages two areas, we want them to be assigned to two councils when they log in.

We can add `council-A.gov.uk` to your list to assign it to "Council B", but we also want the normal code to assign them to "Council A".

So I think this can just be:
```
assign_approved_custom_domain()
councils = Council.objects…
[etc]
```

polling_stations/apps/file_uploads/utils.py
```diff
@@ -12,17 +13,43 @@ def get_domain(request):
     return os.environ.get("APP_DOMAIN", request.META.get("HTTP_HOST"))
 
 
+# This is a quick fix: Peter can raise a
+# PR to update CUSTOM_DOMAINS with the the actual
+# list of email domains and councils as the
+# need arises.
+def approved_custom_domain(user, email_domain):
```
symroe, line 20: With my other comment...

Can you change this to an action like "assign_approved_custom_domains" so we know it does something rather than just checks something?

And with the early return, we don't need the `else` statement, we can just dedent the rest of the code.

VirginiaDooley: Fixed with 9c15a6cd5cf7a4948cfa5c3c331209fe1a0017c2

polling_stations/apps/file_uploads/utils.py
VirginiaDooley: Fixed with 9c15a6cd5cf7a4948cfa5c3c331209fe1a0017c2

VirginiaDooley:
What is the post login redirect behaviour we want when a user email is assigned to multiple councils? At the moment, the standard assignment path dominates the custom domain assignment. For example, `user@ynysmon.gov.uk` is redirected to `/uploads/councils/AGY/` even when included in a custom domain constant for other councils.

polling_stations/apps/file_uploads/utils.py
```diff
@@ -27,6 +27,10 @@ def assign_approved_custom_domains(user, email_domain):
             UserCouncils.objects.using(LOGGER_DB_NAME).update_or_create(
                 user=user, council=council
             )
+            # remove councils that are no longer in the custom list
+            UserCouncils.objects.using(LOGGER_DB_NAME).filter(user=user).exclude(
+                council__council_id__in=email_domain
+            ).delete()
```
VirginiaDooley, line 33: Remove 30-33 as it won't be triggered unless line 24 returns true.

Add to comment to remove the UserCouncil when removing an entry from the CUSTOM_DOMAINS

GeoWill (approved):
Looks good. Can be merged.

PR #4982, opened 2023-02-15
Script to bulk COPY data in to Postgres

symroe:
Downloads data from S3, splits it in to chunks and loads it into Postgres using n processes.


symroe:
I've not tried, but this should work with the processes UPRN to Council file too.

GeoWill:
Rebased on master and done some fixups. Testing so possibly more to come.

GeoWill (approved):

PR #5030, opened 2023-02-24
Hotfix/extra page css

GeoWill:


polling_stations/templates/postcode_view.html
```diff
@@ -10,26 +10,6 @@
...
-            margin-bottom: 1rem;
-            font-size: inherit;
-            line-height: 1.6;
-            text-rendering: optimizeLegibility;
-        }
-
-        .callout.success {
-            margin-top: 2rem;
-        }
-    </style>
```
GeoWill, line 32: I'm not 100% sure this is redundant, but I'm also not 100% sure how to check. When I delete it nothing changes, but perhaps it's for mobile or something? Any ideas?

polling_stations/templates/archived_site.html
```diff
@@ -1,9 +0,0 @@
-<div class="callout warning" style="overflow:auto">
-    <div class="columns large-6 large-centered">
-        <h5>Archived site</h5>
-        <p>
-            This site contains information about polling stations for the General Election in June 2017.
-            We will be updating the site in time for the local elections in May 2018.</p>
-        <a href="https://democracyclub.org.uk/blog/">Follow our blog for news on what we're up to.</a>
-    </div>
-</div>
```
GeoWill, line 9: Pretty sure this is good to go.

polling_stations/templates/postcode_view.html
GeoWill: @VirginiaDooley Reckons these are no longer in use.  Apart from in a template that is dropped  in 0941b62b9f3b3.

VirginiaDooley (approved):

PR #5050, opened 2023-03-01
Update upload email confirmation text

VirginiaDooley:
### PR Checklist

When creating a Pull Request please delete items as necessary, leaving those that are relevant.

- [ ] References a specific issue or if not, describes the bug or feature in detail
- [ ] Note what card in Trello this work relates to
- [ ] Instructions for how reviewers can test the code locally
- [ ] Tests have been added and/or updated
- [ ] Screenshot of the feature/bug fix (if applicable)
- [ ] If any new text is added, it's internationalized
- [ ] Any new elements have aria labels
- [ ] No unintentional console.logs left behind after debugging
- [ ] Did I use the clear and concise names for variables and functions?
- [ ] Did I explain all possible solutions and why I chose the one I did?
- [ ] Added any comments to make new functions clearer
- [ ] Did I added or updated any new dependencies? Explain why.
- [ ] Did I update the package.json and/or Pipfile.lock version if relevant?
- [ ] Have I rebased with the latest version of master?
- [ ] Added PR labels
- [ ] Update any history/changelog file
- [ ] Update any documentation
- [ ] Update dev handbook


VirginiaDooley:
[Uploader confirmation email](https://trello.com/c/d8EvAMLe/3294-uploader-confirmation-email)

polling_stations/apps/file_uploads/api.py
```diff
@@ -59,9 +59,9 @@ def save(validated_data):
             if (
                 file_set.first().ems == "Democracy Counts" and file_set.count() == 2
             ) or file_set.first().ems != "Democracy Counts":
-                user = self.context["request"].user
                 upload.make_pull_request()
-                upload.send_confirmation_email(user=user)
+                user = upload.gss.users.first()
```
VirginiaDooley, line 63: I don't feel like this is the best way forward and would welcome suggestions for a more reliable alternative. I had hoped to pull in the request object and get the user that way, but struggled to do this.

polling_stations/apps/file_uploads/models.py
```diff
@@ -54,6 +55,7 @@ class Upload(models.Model):
 
     objects = UploadQuerySet.as_manager()
     warning_about_pending_sent = models.BooleanField(default=False)
+    upload_user = models.ForeignKey(User, on_delete=models.CASCADE, null=True)
```
GeoWill, line 58: Drive by... Do we definitely want to `CASCADE` the deletes here. I'm not sure - in some ways a council 'owns' the upload object more than the user. I think perhaps `SET_NULL` would be more appropriate - but would be good to know what you think.

VirginiaDooley: I agree, this needs a change. I think if we change to `SET_NULL` then we should add a `Upload.upload_user.email` check where we are sending emails, right?

polling_stations/apps/file_uploads/models.py
```diff
@@ -157,30 +157,55 @@ def pr_body(self):
...
+                subject,
+                message,
+                settings.DEFAULT_FROM_EMAIL,
+                [to],
+                reply_to=[settings.DEFAULT_FROM_EMAIL],
+                headers={"Message-ID": subject},
+            )
 
-        return email.send()
+            return email.send()
```
GeoWill, line 208: This works, but we were chatting about how it could be DRY-ed up.
To achieve that I would re-write it as an `if`/`else` statement and just define `to` and `subject` which are the parts that change - rather than joining it to the previous `if`. We can then construct our `EmailMessage` instance.

I've also removed the `context` you're passing to `render_to_string` - as far as I can see `subject` is not being used in the template - this means it's quite readable just inlining the method call. 

Finally I don't think `return` is doing anything here - you don't use it in `polling_stations/apps/file_uploads/api.py`, so I've left it out. 

```python
# if we're in production, and the upload user exists, send them an email
if server_env == "production" and self.upload_user != None:
    to = self.upload_user.email
    subject = f"Your file upload for {self.gss.short_name} ({self.election_date}) was successful"
# for all other environments, send the email to the default
# from email with a subject line that makes it clear
# we are not in production and testing is taking place
else:
    to = settings.DEFAULT_FROM_EMAIL
    subject = f"**NB triggered from {server_env} instance** Your file upload for {self.gss.short_name} ({self.election_date}) was successful"

email = EmailMessage(
    subject,
    render_to_string(template_name="file_uploads/email/upload_confirmation.txt"),
    settings.DEFAULT_FROM_EMAIL,
    [to],
    reply_to=[settings.DEFAULT_FROM_EMAIL],
    headers={"Message-ID": subject},
)
email.send()
```

cdk/lambdas/wdiv-s3-trigger/tests/test_github_helper.py
```diff
@@ -40,6 +41,7 @@ def test_raise_issue_without_key(self):
     def test_raise_new_issue_with_key(self):
         repo = "chris48s/does-not-exist"
         key = "f00b42"
+        os.environ["SERVER_ENVIRONMENT"] = "production"
```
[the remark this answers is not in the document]
symroe: Oh Gah, yes, I didn't see the file name. Sorry, ignore everything I just said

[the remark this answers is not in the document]
symroe: Can you make an issue about this to fix it post-elections? I think it will cause problems in the future, and we want something like [this](https://stackoverflow.com/questions/2059482/temporarily-modify-the-current-processs-environment/34333710#34333710) at some point

polling_stations/apps/file_uploads/models.py
VirginiaDooley: Sorry, I thought we didn't want to send an email if the user didn't exist, even if in prod?

VirginiaDooley: This change suggests to me that if we are in prod and the user doesn't exist, an email will be sent to DC.

GeoWill: I think you've deleted it, but before you had an `if` statement that caught that situation and returned early - meaning an email wouldn't be sent. I was suggesting replacing the block I selected to be replaced. so you'd have:
```python
if prod and not upload.user:
    return
if prod and upload.user:
    to = upload.user.email
    subject = "..."
else:
    to = DEFAULT
    subject = "TESTING..."
sendemail()
```

GeoWill: > I would re-write it

here my 'it' was referring to the codeblock I selected. Not that clear, sorry.

VirginiaDooley: Ah ok, got it - sorry for the confusion. I've revised as agreed and updated tests.

PR #5059, opened 2023-03-02
2023 EONI importer

symroe:
_I have manually run this against the staging RDS_

---

Take the script form the previous deploy and gets it to run using `RunOncePerTagRunCommandClient` (AWS RunCommand).

This is an initial bit of work to get it on staging for them (and us) to QA. We need to figure out how we want to drive this in terms of how we map our staging on to their staging.

Option 1 is that we give them access to the staging bucket and import any data on there in our staging, same for production.

Option 2 is that our staging is only for us, and "production" is the only thing they see. We manage staging imports by giving them access to the production bucket, but import form a "staging" and "production" folder on the bucket. 

I can see some complications with both, but I think option 1 is easier for all of us, especially as they'll not use the staging bucket very much (maybe a couple of times before each major NI election).

https://trello.com/c/XFvRPMEz/3291-eoni-data-staging-site-import

### TODO

```[tasklist]
### Tasks
- [ ] Document process from EONI's point of view
- [ ] Document how we import data, what the commands do
- [ ] (depending on the above options) take the bucket name form the environment
```

### DIFF
--- a/cdk/lambdas/ssm_run_command_once/command_runner.py
+++ b/cdk/lambdas/ssm_run_command_once/command_runner.py
@@ -52,6 +52,10 @@ class RunOncePerTagRunCommandClient:
             InstanceIds=[self.instance_id],
             DocumentName="AWS-RunShellScript",
             Parameters={"commands": [command]},
+            CloudWatchOutputConfig={
+                "CloudWatchLogGroupName": "/aws/ssm/command_runner",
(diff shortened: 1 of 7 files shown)


GeoWill (approved):
Messed about a bit to get it all working. Decision is to run give them access to dev(and/or) stage bucket.

Issue #4724, opened 2022-05-03
Running import scripts in PRs using CI


alexdutton:
@GeoWill mentioned it'd be useful to automatically run import scripts in CI as soon as they're made, and as they're updated.

There is apparently an RDS instance with the prerequisite data (e.g. AddressBase) set up in AWS that could be connected to.

To connect there are a couple of options:

* expose the RDS port on the public internet
* use SSM and a bastion host to tunnel through to the RDS instance within the VPC

The latter would be considered more secure. You'd need programmatic credentials for your SSM client to access SSM and install an SSH key on your bastion host, and then credentials for the RDS instance.

In another place, we've done [something like this](https://gist.github.com/alexdutton/5010a3b7842f510eb81a579b58dcf430) to set up a tunnel through a bastion server to the RDS instance. The bastion server can be created using an AWS CDK [BastionHostLinux](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.BastionHostLinux.html) construct.

If you use GitHub Actions you can use the [concurrency parameter on a workflow](https://docs.github.com/en/actions/using-jobs/using-concurrency) to ensure only import script is being tested at once.

Also quite possible you'd want the RDS instance and bastion host in its own AWS account?

PR #5119, opened 2023-03-07
Run imports in CI

GeoWill:
The aim here is to run import scripts as they get merged in CI 
Closes https://github.com/DemocracyClub/UK-Polling-Stations/issues/4724

## Trello
https://trello.com/c/0bSKpmSv

## Run it locally
`./manage.py run_new_imports -f 4078f6ac -t fcf57110a`

Thoughts on the approach in #x-will. Key point is I made it a management command because it was easier to pinch code from other bits, and seemed better organised and easier to test to me.

```[tasklist]
### PR Checklist

- [ ] References a specific issue or if not, describes the bug or feature in detail
- [ ] Note what card in Trello this work relates to
- [ ] Instructions for how reviewers can test the code locally
- [x] Tests have been added and/or updated
- [ ] Did I use the clear and concise names for variables and functions?
- [ ] Did I explain all possible solutions and why I chose the one I did?
- [ ] Have I rebased with the latest version of master?
- [ ] Added PR labels
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,270 +1,34 @@
-
 version: 2.1
 
+# this allows you to use CircleCI's dynamic configuration feature
+setup: true
+
+# the path-filtering orb is required to continue a pipeline based on
+# the path of an updated fileset
 orbs:
-  node: circleci/node@5.0.2
-  slack: circleci/slack@4.1
-
-jobs:
-  build_and_test:
-    docker:
-      - image: cimg/python:3.10.8-browsers
-        environment:
-          CIRCLECI: true
-          PGHOST: 127.0.0.1
-      - image: cimg/postgres:14.5-postgis
-        environment:
-          POSTGRES_USER: postgres
-          POSTGRES_DB: polling_stations
-
-    working_directory: ~/repo
-
-    steps:
-      - checkout
-
-      - restore_cache:
-          key: v1-polling_stations-dependencies-{{ checksum "requirements/base.txt" }}-{{ checksum "requirements/testing.txt" }}-{{ checksum "requirements/cdk.txt" }}
-
-      - node/install:
-          node-version: '16.13.0'
-
-      - run:
-          name: Install node node_modules
-          command: npm install
-          cache-path: ~/repo/node_modules
-          override-ci-command: npm install
-
-      - run:
-          name: install app dependencies
-          command: |
-            sudo apt update && sudo apt install -y gdal-bin
-            pyenv local $(pyenv versions --bare | grep 3.10 | head -n 1)
-            python -m venv .venv
-            . .venv/bin/activate
-            python -m pip install --upgrade pip
-            python -m pip install wheel
-            python -m pip install coveralls
-            python -m pip install -r requirements/testing.txt
-            playwright install
-
-      - run:
-          name: install CDK Python dependencies
-          command: |
-            . .venv/bin/activate
-            pip install -r requirements/cdk.txt
-
-      - save_cache:
-          paths:
-            - ./.venv
-            - ./node_modules
-            - ./home/circleci/.cache
-          key: v1-polling_stations-dependencies-{{ checksum "requirements/base.txt" }}-{{ checksum "requirements/testing.txt" }}-{{ checksum "requirements/cdk.txt" }}
-
-      - run:
-          name: Install HTML Tidy
-          command: wget https://github.com/htacg/tidy-html5/releases/download/5.8.0/tidy-5.8.0-Linux-64bit.deb && sudo dpkg -i tidy-5.8.0-Linux-64bit.deb
-
-      - run:
-          name: Print versions
-          command: |
-            . .venv/bin/activate
-            python --version
-            python manage.py --version
-
-      - run:
-          name: Pre-test checks
-          command: |
-            . .venv/bin/activate
-            black --check .
-            python -m pip check
-            python manage.py check
-            python manage.py makemigrations --check
-
-      - run:
-          name: Pytest
-          command: |
-            . .venv/bin/activate
-            python manage.py collectstatic --no-input
-            pytest --flakes --cov-report= --cov=polling_stations
-
-      - run:
-          name: Submit coverage
-          command: |
-            . .venv/bin/activate
-            coveralls
-
-  cdk_synth:
-    machine:
-      image: ubuntu-2204:2022.10.2
-    working_directory: ~/repo
-    parameters:
-      dc-environment:
-        type: enum
-        enum: [ development, staging, production ]
-    environment:
-      DC_ENVIRONMENT: "<<parameters.dc-environment>>"
-
-    steps:
-    - checkout
-    - restore_cache:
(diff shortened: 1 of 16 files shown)


GeoWill:
I seem to have broken coveralls :roll_eyes:

polling_stations/apps/data_importers/base_importers.py
```diff
@@ -194,6 +204,23 @@ def get_base_folder_path(self):
...
+        """
+        if not hasattr(self, "elections"):
+            return False
+        for date_str in self.elections:
+            try:
+                election_date = datetime.date.fromisoformat(date_str)
+                if election_date >= datetime.date.today():
+                    return True
+            except ValueError:
+                pass
```
symroe, line 221: Is it worth raising a warning here?

polling_stations/apps/data_importers/management/commands/run_new_imports.py
```diff
@@ -0,0 +1,167 @@
...
+            try:
+                call_command(script_path.stem, **opts)
+                self.summary.append(
+                    (
+                        "INFO",
+                        f"Ran import script: {script_path.name}",
+                    )
+                )
+            except:
+                # usually we want to handle a specific exception, but in this situation
```
symroe, line 98: Is this true? Why don't you want to fail CI if the import script can't run?

symroe: Also, it's a good idea to at least catch `Exception`. This means that other

polling_stations/apps/data_importers/management/commands/run_new_imports.py
```diff
@@ -0,0 +1,167 @@
...
+            "include_past_elections": False,
+        }
+
+        sys.stdout.write(
+            f"Comparing repo between {from_sha} and {to_sha}\n"
+            f"\tFrom: https://github.com/DemocracyClub/UK-Polling-Stations/commit/{git_rev_parse(from_sha)}\n"
+            f"\tTo: https://github.com/DemocracyClub/UK-Polling-Stations/commit/{git_rev_parse(from_sha)}\n"
+        )
+
+        # __import__("ipdb").set_trace()
```
symroe: Clean up

polling_stations/apps/data_importers/management/commands/run_new_imports.py
```diff
@@ -0,0 +1,167 @@
...
+        if has_imports and not has_application:
+            sys.stdout.write("Only import scripts have changed\n")
```
symroe: I think you want `self.stdout` in these. `sys.stdout.write` is the same as `print`, but it's useful to use whatever `self.stdout` is set to in management commands

polling_stations/apps/data_importers/tests/test_run_new_imports.py
```diff
@@ -0,0 +1,134 @@
...
+        "polling_stations/apps/data_importers/management/commands/import_barrow_in_furness.py",
+        "polling_stations/apps/data_importers/management/commands/import_eoni.py",
+        "polling_stations/apps/data_importers/management/commands/import_wigan.py",
+        "polling_stations/settings/constants/councils.py",
+        "polling_stations/templates/base.html",
+        "polling_stations/templates/postcode_view.html",
+        "requirements/base.txt",
+    ]
+    assert expected_empty == get_paths_changed(
+        "1012c398b13d2e9c87f718b87e07ee9cd1c26222",
```
symroe, line 52: I see what you've done here, but something feels…odd about this code checking it's own git history. I wonder what happens if, for whatever reason, we change the hashed (for example, what if we pull import scripts into a new repo? Maybe this PR is moot in that case, but you get my point).

To mitigate changes, can you leave a comment as to how you found these magic values?

.circleci/config.yml
```diff
@@ -233,6 +273,18 @@ workflows:
         min-size: 1
         max-size: 2
         desired-capacity: 1
+    - post_deploy_tasks:
+        name: "Development: Post-Deploy Tasks"
+        requires:
+        - "Development: AWS CodeDeploy"
+        context: [CircleCIAPIUser, deployment-development-wdiv]
+        filters: { branches: { only: [ main, master ] } }
+    - pre_deploy_tasks:
```
symroe, line 282: Would you mind ordering these in the way the tasks are ordered? Like, move `pre` before all other steps, and `post` after the others?

symroe (commented):
I think you need to implement `--exit_early` and call this on all jobs, otherwise the deploy will always happen?

polling_stations/apps/data_importers/management/commands/run_new_imports.py
GeoWill: rebase fail - I did that here 83ccb291a637243967da56b12e27ce37f7e4a667

polling_stations/apps/data_importers/management/commands/run_new_imports.py
GeoWill: ditto.

polling_stations/apps/data_importers/base_importers.py
GeoWill: Good shout.

polling_stations/apps/data_importers/management/commands/run_new_imports.py
GeoWill: Right, so my thinking here was that we don't want failed import scripts to block deploys. But I think I actually agree with you. Most of our import script changes are single commits, so it's simple enough to just drop that commit if there's an issue. I think I might go with your recommendation of catching `Exception` as a starting point, then refactor further to take it out completely.

GeoWill:
> I think you need to implement --exit_early and call this on all jobs, otherwise the deploy will always happen?

So I haven't implemented that because I didn't think it was needed. 

### Only import scripts
- Command is called as in the `pre-deploy` job. This:
  - runs the import scripts
  - sets the ssm parameter
  - quits the circle jobs

### Only app
- No scripts are run
  - Checking time seems negligible

### Mix of the two
- pre-deploy
  - Scripts don't run because --post-deploy flag isn't set
- post-deploy
  - scripts run, but nothing to exit from.

symroe:
> quits the circle jobs

`inigo-montoya.gif`.

So, `args = ["circleci-agent", "step", "halt"]` doesn't stop the _workflow_, it stops the _step_ cleanly. However the workflow continues, meaning that, as things stand, a deploy will happen each time.

[This](https://github.com/DemocracyClub/circle-ci-conditional-job-test/blob/new-new-stroud/magic.py#L54-L62) was my attempt at stopping the workflow, but that was the approach that made it look like all the jobs had failed, so we chose to [add a check to each job](https://github.com/DemocracyClub/circle-ci-conditional-job-test/blob/new-new-stroud/.circleci/config.yml#L23-L25)

symroe:
> doesn't stop the workflow, it stops the step cleanly

Maybe think of it like and early return of a method rather than a `sys.exit()`?

GeoWill:
OK, I think I've got it now. Clearly wishful thinking as I convinced myself you'd solved that with the `step halt` approach!

PR #5275, opened 2023-03-15
Feature/Photo ID Wording

Bekabyx:
[Trello](https://trello.com/c/UvMnJkuS/3207-voter-id-messaging)

Updated the home page and the postcode lookup to include information about photo ID for voters, based off of Peter's suggestions in [this document.](https://docs.google.com/document/d/1QHfjTaSUQ-p7thwyCOo4-dpd27N4b1sdM9u7VRqhCZg/edit)

This PR also sets the value for `NEXT_CHARISMATIC_ELECTION_DATE` so we're ready for May 4th.

For local testing, you'll need to run the project both with and without the election date set, once to see the election season text and once to see the not-election-season text. When testing the postcode lookup page, ensure that you're using a postcode which does have an upcoming election.

**Homepage, election date unset:**

Screenshot 2023-03-15 at 11 51 17

**Homepage, election date set:** 

Screenshot 2023-03-15 at 16 45 30

**Postcode lookup, election requires voter ID:**

Screenshot 2023-03-15 at 11 51 36

**There are no changes to the postcode lookup page for elections which do not require ID on any of their associated ballots**

```[tasklist]
### PR Checklist

- [x] Tests have been added and/or updated
- [x] Screenshot of the feature/bug fix (if applicable)
- [x] If any new text is added, it's internationalized
- [x] Any new elements have aria labels
- [x] No unintentional console.logs left behind after debugging
- [x] Did I use the clear and concise names for variables and functions?
- [x] Added any comments to make new functions clearer
```


polling_stations/apps/data_finder/tests/playwright/test_postcode_lookup.py
Bekabyx, line 1: This Playwright test will need to be updated (and more added) to cover the new ID rules. 

The reason this part of the test has been removed is because I've been having an absolute mare trying to get it working on my machine and I felt like I had exceeded a sensible timebox.

New tests haven't yet been added because I'm not sure how to work with the legacy cassettes to achieve it.

polling_stations/templates/fragments/you_dont_need_poll_card.html
```diff
@@ -1,16 +1,27 @@
 {% load i18n %}
-{% if voter_id_pilot %}
```
GeoWill, line 2: Since we're deleting this, does it make sense to also delete the method in ee_wrapper, and where it's added to  `context` in `data_finder/views.py`?

polling_stations/templates/fragments/you_dont_need_poll_card.html
```diff
@@ -1,16 +1,27 @@
...
+            <p>
+                {% trans "You must vote at your assigned polling station." %}
+            </p>
+
+            {% if territory == 'NI' %}
+                <p><a href="http://www.eoni.org.uk/Vote/Voting-at-a-polling-place" target="_top">
+                    {% trans "Read more about voting in Northern Ireland" %}
+                </a></p>
+            {% else %}
+                <p><a href="https://www.gov.uk/voting-in-the-uk/polling-stations" target="_top">
```
GeoWill, line 21: I know we're not getting these translated for these elections, but since you're putting the scaffolding in now, is it possible to blank out the links to the welsh language equivalents of the gov uk and electoral commission websites? if not straightforward , then don't worry about it.

polling_stations/apps/data_finder/tests/playwright/test_postcode_lookup.py
GeoWill: That sounds frustrating! Shame as playwright was mean to be easier to work with than our previous integration tests.
Can you make an issue to revisit this - I wonder if we can move to a mixture of mocking the API calls, and using HTTP Archive Files - which playwright seems to be set up to work with. ([API mocking](https://playwright.dev/python/docs/mock) and [HAR files](https://playwright.dev/python/docs/network#record-and-replay-requests) docs)

GeoWill (commented):
Bit pressed for time, so just a speed read through, and got it running locally. Basically looks good, with a couple of thoughts/questions inline.

polling_stations/templates/fragments/you_dont_need_poll_card.html
Bekabyx: On our translated version of the postcode page, we don't link to the CY "voting in the UK" page (https://www.gov.uk/voting-in-the-uk/polling-stations) in this section. The text reads "Darllen mwy am bleidleisio ym Mhrydain Fawr" but it always links to the English language version.

The new link added for the EC website on the home page also doesn't go through to the CY version, and it has no Welsh text translation either. 
Screenshot 2023-03-17 at 10 36 23

I don't think there's anything to do here, but I hope I haven't misunderstood your comment 🙊

polling_stations/templates/fragments/you_dont_need_poll_card.html
Bekabyx: The context assignment has been removed now 🙂

GeoWill: Do you know if we're still expecting NI voter id information to be in the metadata? Looking at https://elections.democracyclub.org.uk/api/elections.json?postcode=BT15%203JX&future=1&current=1%27 metadata is `null` for all ballots. So if we know EE isn't using the `metadata` field to communicate about voter id - and we're going to use the `requires_voter_id` key, then I think we should drop that logic too. If it is the case that the information might come by either route (and that wouldn't be an error) then I can see we should leave it in.

polling_stations/templates/fragments/you_dont_need_poll_card.html
GeoWill: So I know we were only linking to the English version, but I think think there's a way to have 'translated links', so when we come to get translations done, we are reminded to also go and search out the CY versions of the links - I was imagining that it'd by annoying to be reading a site in Welsh, then get linked to the English language version of another site that also has a welsh translation. However not a priority for us!

GeoWill: Let's ignore this for now.

polling_stations/templates/fragments/you_dont_need_poll_card.html
GeoWill: I confirmed in tech surgery that we're not expecting voter id info for Northern Ireland in the `metadata` object, so I think all that can be removed as part of this work:

Bekabyx: Grand, I assume that same confirmation also applies to the 2019 voter id trials so I shall remove it

polling_stations/apps/data_finder/tests/playwright/test_postcode_lookup.py
```diff
@@ -109,6 +109,3 @@ def test_northern_ireland(page, live_server):
         expect(
             page.locator("text=The Electoral Office for Northern Ireland")
         ).not_to_be_empty()
-        expect(
-            page.locator("text=You will need photographic identification")
-        ).not_to_be_empty()
```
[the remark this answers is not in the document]
Bekabyx: Thank you! I'll update this and check it in CI to be sure but this looks like it'll probably fit the bill 😁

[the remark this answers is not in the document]
Bekabyx: This should be sorted and neatened up now. What I ended up with wasn't quite an exact match for what you've got here but is functionally the same.

GeoWill (approved):
Looks good. Can be merged whenever.

PR #5543, opened 2023-04-13
Fix markdown filter

GeoWill:
Spent 15mins trying to get a playwright test to work and failed, however this fixes it locally for me. I am confused about whether this has been broken for a long time though.

I think this has been removed (unitentionally?) from dc_django_utils in this [commit](https://github.com/DemocracyClub/dc_django_utils/commit/698ddc54d79cfcd9b21a1293bf617f39f8a78d3d). 

This puts it back as a stop gap.

### DIFF
--- a/polling_stations/settings/base.py
+++ b/polling_stations/settings/base.py
@@ -163,7 +163,6 @@ INSTALLED_APPS = (
     "rest_framework.authtoken",
     "rest_framework_gis",
     "django_extensions",
-    "markdown",
     "corsheaders",
     "pipeline",
     "dc_signup_form",
--- a/polling_stations/templatetags/markdown_filter.py
+++ b/polling_stations/templatetags/markdown_filter.py
@@ -1,0 +1,14 @@
+from django import template
+from django.utils.safestring import mark_safe
+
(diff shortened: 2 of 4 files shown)


polling_stations/settings/base.py
```diff
@@ -163,7 +163,6 @@
     "rest_framework.authtoken",
     "rest_framework_gis",
     "django_extensions",
-    "markdown",
```
GeoWill, line 166: Not sure what this was doing WCIVF doesn't have it.

polling_stations/templatetags/markdown_filter.py
```diff
@@ -0,0 +1,14 @@
...
+
+register = template.Library()
+
+
+@register.filter(name="markdown")
+def markdown_filter(text):
+    return mark_safe(markdown.markdown(text))
+
+
+markdown_filter.is_safe = True
```
GeoWill, line 14: Lifted from WCIVF

symroe (approved):
Not 100% sure on what went on, but this seems like a decent quick fix :+1:

PR #5600, opened 2023-04-18
Support multiple charismatic election dates

GeoWill:


polling_stations/apps/data_finder/helpers/every_election.py
```diff
@@ -105,18 +105,29 @@ def get_future_election_dates(self):
 
     def _get_next_election_date(self):
         ballots = self.get_all_ballots()
-        next_charismatic_election_date = getattr(
-            settings, "NEXT_CHARISMATIC_ELECTION_DATE", None
-        )
+        # if no ballots, return early
```
GeoWill, line 108: Check this is right - do we want to return `NEXT_CHARISMATIC_ELECTION_DATE` when there's no ballots for this area? (as is current behaviour)

PR #5603, opened 2023-04-18
Support multiple charismatic election dates

GeoWill:
```[tasklist]
### PR Checklist

- [x ] Instructions for how reviewers can test the code locally
- [ x] Tests have been added and/or updated
- [x] Did I explain all possible solutions and why I chose the one I did?
- [x] Added any comments to make new functions clearer
```

**testing locally**
Make sure `settings/local.py` contains:
```
EVERY_ELECTION = {"CHECK": True}
NEXT_CHARISMATIC_ELECTION_DATES = ["2023-05-04", "2023-05-18"]
``` 
Try a postcode with an election on another date (eg EN3 6QT), try postcode from NI, try both with bits commented out. etcetc.

**Docs**

Suggest adding the following to the wiki on the what to do for an election section:
Set `NEXT_CHARASMATIC_ELECTION_DATES` to be a list of dates (yyy-mm-dd).
This will mean that users are only told about an upcoming election if their postcode has an election that falls on one of the dates contained in the list. Otherwise we will say 'we don't know about any upcoming elecions. Generally only major elections should be in this list, and probably only one. However we needed 2 when the NI locals were on a different day to the England locals (2023-05-18 and 2023-05-04) .


Bekabyx:
I've noticed that the homepage messaging is no longer updating for election day with this change. I modified the `NEXT_CHARISMATIC_ELECTION_DATES` list to also contain today's date and I'd expect to see the card which says "Polling stations are open from 7am to 10pm today". 

The logic for this doesn't hook directly in to the old `NEXT_CHARISMATIC_ELECTION_DATE` but does leverage `datafinder > views.py > get_context_data()` which is still looking for the singular form charismatic date.

GeoWill:
Good spot!

Bekabyx:
I've had another play about and as far as I can tell everything is looking as I'd expect it to 😁

Bekabyx (approved):

PR #5687, opened 2023-04-27
Serve CF from www subdomain

symroe:
First part of the work to redirect from WWW to non-WWW, but at least this fixes some links that are in the wild at the moment.

Note that I've applied these changes manually. A deploy isn't going to work as there'll be a duplicate A record. We need to delete that before deploying.


GeoWill (approved):

GeoWill:
Don't merge till quiet time.

GeoWill:
https://github.com/DemocracyClub/UK-Polling-Stations/pull/5703

PR #5691, opened 2023-04-27
Hotfix/countermanded poll edits

VirginiaDooley:
Ref https://trello.com/c/xJEReIxB/3334-%E2%98%B9-countermanded-poll-issues-%E2%98%B9

https://wheredoivote.co.uk/address/100010661523/

This work repairs the broken markdown filter and adds a link to the corresponding cancelled election on WCIVF. 

To test: WDIV pulls data from EE so rather than manipulating objects in the shell, I edited [this method](https://github.com/DemocracyClub/UK-Polling-Stations/blob/7cc8d2fbf7282572992ed6ae7c88f3132bd2547b/polling_stations/apps/data_finder/helpers/every_election.py#L150) as follows: 

```
def get_cancelled_election_info(self):
        rec = {
            "cancelled": True,
            "name": "Test Election",
            "rescheduled_date": "March 4th",
            "metadata": "This is test data",
        }
```
and changed conditionals in templates until I was able to manufacture a cancelled election. 

Screenshot 2023-04-27 at 4 07 46 PM

```[tasklist]
### PR Checklist

- [ ] Tests have been added and/or updated
- [x] Screenshot of the feature/bug fix (if applicable)

### DIFF
--- a/polling_stations/templates/fragments/cancelled_election.html
+++ b/polling_stations/templates/fragments/cancelled_election.html
@@ -14,12 +14,11 @@
     {% else %}
         <p>{% blocktrans with cancelled_election_name=cancelled_election.name %}The poll for <strong>{{ cancelled_election_name }}</strong> has been cancelled.{% endblocktrans %}</p>
     {% endif %}
+    <p>To learn more, please visit <a href="https://whocanivotefor.co.uk/elections/{{ cancelled_election.slug }}/">Who Can I Vote For?</a></p>
 {% endif %}
 
 {% if cancelled_election.metadata.cancelled_election.detail %}
-    {% filter markdown %}
-        {{ cancelled_election.metadata.cancelled_election.detail }}
-    {% endfilter %}
+    {% filter markdown %}{{ cancelled_election.metadata.cancelled_election.detail }}{% endfilter %}
 {% else %}
     {% if council.electoral_services_phone_numbers %}
         <p>
--- a/.circleci/deploy-config.yml
+++ b/.circleci/deploy-config.yml
@@ -342,14 +342,14 @@ workflows:
         requires:
         - "CDK Synth"
         context: [deployment-development-wdiv, slack-secrets]
-        filters: { branches: { only: [ development ] } }
+        filters: { branches: { only: [ development, hotfix/superuser-conf-email ] } }
         dc-environment: development
     - code_deploy:
         name: "Development: AWS CodeDeploy"
         requires:
         - "Development: CDK Deploy"
         context: [deployment-development-wdiv, slack-secrets]
-        filters: { branches: { only: [ development ] } }
+        filters: { branches: { only: [ development, hotfix/superuser-conf-email ] } }
         dc-environment: development
         min-size: 1
         max-size: 2

### CODE REVIEW

**Comment by VirginiaDooley** [comment]
[☹ Countermanded poll issues ☹](https://trello.com/c/xJEReIxB/3334-%E2%98%B9-countermanded-poll-issues-%E2%98%B9)

File: polling_stations/templates/fragments/cancelled_election.html [hunk 1]
```diff
@@ -14,12 +14,11 @@ <h2>{% trans "Cancelled Election" %}</h2>
     {% else %}
         <p>{% blocktrans with cancelled_election_name=cancelled_election.name %}The poll for <strong>{{ cancelled_election_name }}</strong> has been cancelled.{% endblocktrans %}</p>
     {% endif %}
+    <p>To learn more, please visit <a href="https://whocanivotefor.co.uk/elections/{{ cancelled_election.slug }}/">Who Can I Vote For?</a></p>
```
GeoWill, line 17: There's no `slug` property on `cancelled_election`, so this won't work. I think it would be better to direct them to `whocanivotefor.co.uk/elections/{{ postcode }}`. Of course this might be confusing with split postcodes, but it's probably better than nothing. If (as seems imminent) WCIVF gets an address picker, then we might need to be clever about sending a postcode or a uprn depending on whether the postcode is split or not. But maybe let's cross that bridge later.

The reason I don't think we can use an election slug is that there might be more than one, (eg local and mayor). For context: we arrive at the `cancelled_election` fragment if the `context["has_election"]` is false  ([link](https://github.com/DemocracyClub/UK-Polling-Stations/blob/7cc8d2fbf7282572992ed6ae7c88f3132bd2547b/polling_stations/templates/postcode_view.html#L36)) and there is a cancelled election ([link](https://github.com/DemocracyClub/UK-Polling-Stations/blob/7cc8d2fbf7282572992ed6ae7c88f3132bd2547b/polling_stations/templates/fragments/no_election.html#L4)). 

So we might want to think about handling the case where there is an uncancelled and a cancelled election by telling people about the cancelled election in a card below their polling station information. However I'm hesitant about that, as it feels like scope creep for WDIV.

polling_stations/templates/fragments/cancelled_election.html
```diff
@@ -14,12 +14,11 @@ <h2>{% trans "Cancelled Election" %}</h2>
...
 {% endif %}
 
 {% if cancelled_election.metadata.cancelled_election.detail %}
-    {% filter markdown %}
-        {{ cancelled_election.metadata.cancelled_election.detail }}
-    {% endfilter %}
+    {% filter markdown %}{{ cancelled_election.metadata.cancelled_election.detail }}{% endfilter %}
```
GeoWill, line 21: If your pipe solution works maybe it's better as it can't be regressed by someone deciding to split this line into three again?

GeoWill (changes requested):

GeoWill:
A general point about test driving these changes. If you change:
```
EVERY_ELECTION = {"CHECK": False, "HAS_ELECTION": True}
```
to
```
EVERY_ELECTION = {"CHECK": True}
```
in `local/settings.py`, then your local install will go and talk to `elections.democracyclub.org.uk`, so as long as you use an actual cancelled election (which we currently have) then you can see what it will look like when deployed. However that's only true while we have current cancelled elections. Outside of election season this won't work, and some mocking/patching will be needed.

GeoWill:
https://github.com/DemocracyClub/UK-Polling-Stations/pull/5703

PR #5844, opened 2023-08-10
Ruff ruff

GeoWill:


pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[tool.ruff]
+line-length = 88
+ignore = [
+    "E501",
+    "E203", #https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#slices
```
symroe, line 5: I've not seen this in other projects, do you think we need to add it to them?

symroe (changes requested):
Can you: 

1. Update `CONTRIBUTING.md`?
2. Delete `pytest.ini`, and ensure it's fully replaced by `pyproject.toml`?

Otherwise, this is looking good!

symroe (approved):

pyproject.toml
GeoWill: I didn't actually add it based on something breaking, rather reading the black docs about integration - black and ruff were fighting over something else (I'm not sure exactly what). This seemed like a good idea™, but doesn't have a specific purpose. So happy to take out if you prefer.

PR #5848, opened 2023-08-14
Django upgrade <4.3

VirginiaDooley:
This work bumps Django<4.3 as well as CI orb versions.

`appnope` which was added automatically following an install and specific to Macs. https://pypi.org/project/appnope/

To test, follow instructions for installing requirements in `README` and run tests and run server and search for a test postcode such as `EX1 1JN`. 

```[tasklist]
### PR Checklist
- [x] Note what card in Asana this work relates to
- [x] Did I added or updated any new dependencies? Explain why.
```

---
  - https://app.asana.com/0/0/1205090401342430


GeoWill:
I've tried to fix how dependencies are managed in #5851. I think you should be able to reset this branch to new `master`, edit `requirements/base.in` to `django<4.3` and then run `invoke requirements --upgrade`. It should Just Work™... but I'll be interested to hear how it breaks ;)

README.md
```diff
@@ -48,14 +48,16 @@ pip install -r requirements/base.txt -r requirements/testing.txt -r requirements
...
+pip install -r requirements/base.txt -r requirements/testing.txt -r requirements/local.txt -r requirements/cdk.txt -r cdk/lambdas/wdiv-s3-trigger/requirements.txt
 ```
 
 - Just Running Tests
 ```
 pip install -r requirements/base.txt -r requirements/testing.txt
 ```
 
+- Update requirements
+Bump version in relevant requirements file, e.g. ```base.in``` then run ```python -m invoke requirements --upgrade```.
```
GeoWill, line 60: Only single backticks required. Also I've a card in asana to document this properly, but this is better than it just being in my head - so thanks!

README.md
```diff
@@ -48,14 +48,16 @@ pip install -r requirements/base.txt -r requirements/testing.txt -r requirements
 
 - Local Dev with CDK libs
 ```
-pip install -r requirements/base.txt -r requirements/testing.txt -r requirements/local.txt -r requirements/cdk.txt
+pip install -r requirements/base.txt -r requirements/testing.txt -r requirements/local.txt -r requirements/cdk.txt -r cdk/lambdas/wdiv-s3-trigger/requirements.txt
```
GeoWill, line 51: Can you add this to the lines above and below.

.circleci/deploy-config.yml
```diff
@@ -112,7 +112,7 @@ jobs:
           command: |
             . .venv/bin/activate
             python manage.py collectstatic --no-input
-            pytest --ruff --cov-report= --cov=polling_stations
+            pytpytestruff --cov-report= --cov=polling_stations
```
GeoWill, line 115: I don't think this is intentional

GeoWill (changes requested):
Looking good. Thanks! 
I'm a bit wondering about d5b59b2e0a36204f08a5eaaab72d927255367080 and 8f8b5b9a519938c90e9f701407b6a317c43092be. 
First off, the commit messages are a bit confusing I think you mean 'invoke requirements --upgrade'. 

More interesting is it looks like you ran it twice and got two lots of changes? Is this the case? 

Either way I reckon they can be squashed and the message amended.

VirginiaDooley:
> 8f8b5b9



> Looking good. Thanks! I'm wondering about [d5b59b2](https://github.com/DemocracyClub/UK-Polling-Stations/commit/d5b59b2e0a36204f08a5eaaab72d927255367080) and [8f8b5b9](https://github.com/DemocracyClub/UK-Polling-Stations/commit/8f8b5b9a519938c90e9f701407b6a317c43092be). First off, the commit messages are a bit confusing I think you mean 'invoke requirements --upgrade'.
> 
> More interesting is it looks like you ran it twice and got two lots of changes? Is this the case?
> 
> Either way I reckon they can be squashed and the message amended.

You're right, will clarify and squash. One result was pre-rebase with master, the other afterwards.

PR #5903, opened 2023-09-14
Addressbase polygon lookups

GeoWill:


polling_stations/apps/addressbase/management/commands/create_addressbase_lookup.py
```diff
@@ -242,9 +243,9 @@ def create_outcode_lookup(self, outcode, cursor):
     def create_sharded_lookup(self):
         (self.dst_dir / "outcodes").mkdir(exist_ok=True, parents=True)
         outcodes = self.get_outcodes()
-        with connection.cursor() as cursor:
+        with ThreadPoolExecutor(max_workers=4) as executor:
             for outcode in outcodes:
-                self.create_outcode_lookup(outcode, cursor)
+                executor.submit(self.create_outcode_lookup, outcode)
```
GeoWill, line 248: This does not result in any real speed up - I experimented with 4-8 workers (8 cores on my laptop), it pins all my cores to 100%, which the non threadpooled implementation doesn't, but the work doesn't get done faster. Not sure why not, but something in don't understand. My RAM (16GB) doesn't get maxed out by either.

GeoWill:
First commit is ready for review, I haven't tested it, but have used it to produce the lookups for the new constituencies. It needs files in quite specific formats, but the validation checks should make that clear if they're wrong.

PR #5948, opened 2023-10-17
Cdk 2

GeoWill:
I think this Just Works™... I've deployed to dev and it all went smoothly. I can't get `cdk synth` to run locally - but that was true before the update, and is something to do with how I've installed cdk I think.


symroe (approved):
Well this is good news!

`cdk synth` _does_ work for me. I'll leave the actual review to the CloudFormation gods

PR #6100, opened 2024-01-31
Spike/data event model

GeoWill:


polling_stations/apps/councils/models.py
```diff
@@ -130,6 +131,27 @@ def import_script_path(self):
...
+        return self.dataevent_set.filter(event_type=event_type).latest()
+
+    @property
+    def live_upload(self):
+        if not self.has_polling_stations_in_db:
+            return None
+        latest_import = self.latest_data_event(DataEventTypes.IMPORT)
+        latest_teardown = self.latest_data_event(DataEventTypes.TEARDOWN)
+
+        if latest_teardown.created > latest_import.created:
```
symroe, line 150: if either of these are None, this will raise

polling_stations/apps/file_uploads/templates/file_uploads/council_detail.html
```diff
@@ -52,8 +52,10 @@ <h2>Uploads</h2>
                             {% endif %}
                         </td>
                         <td>
-                            {% if upload.status != "Pending" %}
-                                {% if STATIONS %}✔{% else %}❌{% endif %}
+                            {% if upload.pk == live_upload.pk %}
```
GeoWill, line 55: I think this will raise if `live_upload` is `None`

polling_stations/apps/data_finder/views.py
```diff
@@ -27,6 +29,20 @@
 from .helpers.every_election import EmptyEveryElectionWrapper
 
 
+def polling_station_current(station):
+    if not station:
+        return False
+    if not hasattr(station, "elections"):
+        return False
```
GeoWill, line 36: This means we're only showing data when there's an `IMPORT` type `DataEvent` object  with the `election_dates` field populated.
Is this correct?

polling_stations/apps/data_finder/views.py
```diff
@@ -143,6 +159,26 @@ def get_directions(self):
...
+            latest_election_date = max(elections)
+            if latest_election_date >= timezone.now().date():
+                return True
+        except DataEvent.DoesNotExist:
+            # Do we want this to be the default?
+            return False
+        except AttributeError as err:
+            if err.args[0] == "'NoneType' object has no attribute 'election_dates'":
+                return False
+            raise
```
GeoWill, line 176: These exceptions also mean we're only showing data when there's an `IMPORT` type `DataEvent` object with the `election_dates` field populated.
Is this correct?

polling_stations/apps/addressbase/models.py
```diff
@@ -63,6 +65,29 @@ def polling_station(self):
...
+        for the relevant council
+        """
+        try:
+            elections = Subquery(
+                DataEvent.objects.filter(
+                    council_id=OuterRef("council_id"),
+                )
+                .order_by("-created")
+                .values("election_dates")[:1]
+            )
```
GeoWill, line 82: Should this subquery be filtering on `event_type`

symroe: I think so, to guard against any future event types. It's not going to add any overhead to the query, is it?

polling_stations/apps/data_finder/views.py
symroe: I don't know right now how much we enforce the election dates, but I think we should do so. In that case, this is ok, I think!

polling_stations/apps/councils/models.py
```diff
@@ -140,6 +142,36 @@ def import_script_path(self):
...
         return str(import_script_path)
 
+    @property
+    def has_polling_stations_in_db(self):
+        if self.pollingstation_set.count() > 0:
+            return True
+        return False
+
+    def latest_data_event(self, event_type):
+        data_event_model = apps.get_model("data_importers", "DataEvent")
```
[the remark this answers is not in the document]
GeoWill: That's helpful, as I hadn't thought about the string option. 
There is this line: https://github.com/DemocracyClub/UK-Polling-Stations/blob/c426b89f5eb48827900fa6cc546778bd8000db98/polling_stations/apps/data_importers/models.py#L28 in the `data_importers/models.py` module, which would still need the `apps.get_model` to avoid the circular import issue.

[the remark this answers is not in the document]
GeoWill: But is maybe a better place for the complexity to live.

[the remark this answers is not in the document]
GeoWill: Just went on a bit of tangent trying to override the `Council.save` method as:
 
```
def save(self, **kwargs):
    super().save()
    DataQuality.objects.get_or_create(council_id=self.council_id)
```

This didn't work, when creating council records using the `CouncilFactory`.

Not worth getting my head around it. So just going to shift `apps.get_model` to the `data_importers/models.py` and leave it for now. Getting rid of the signal is a nice to have, but not needed.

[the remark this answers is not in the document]
GeoWill:
```
def council_saved(sender, **kwargs):
    DataQuality.objects.get_or_create(council=kwargs["instance"])


council_model = apps.get_model("councils", "Council")
post_save.connect(council_saved, sender=council_model)
```

Doesn't work either because `raise AppRegistryNotReady("Models aren't loaded yet.")`.

I think I'm going to sack it off as Too Difficult and leave it as is.

PR #6124, opened 2024-02-07
Feature/unpublish polling stations

GeoWill:
### Admin list view
image

### Admin detail view
image

### Council List of Stations (viewed as superuser with links)
nb published/unpublished will be visible to councils.
image


symroe:
This looks good.

Two thoughts: 

1. Do we want to explain in the admin help text that a station will be re-published if/when there's another import? Like, this is something that needs to be kept track of, and un-publishing is a bit of a hack? Obviously the user base for this feature is >4 people, but still...
2. Do you want to show the published status on the table in the uploader interface? So that a council can see that it's no longer published.

symroe:
Could use the "source" pattern from WCIVF to add a note about unpublished stations 
image

GeoWill:
> Could use the "source" pattern from WCIVF to add a note about unpublished stations

I like this. Will do something similar.

I didn't bother with the explainer on the basis that I was hoping to get something in using events to keep track of what's been done. But maybe that's optimistic.

polling_stations/apps/pollingstations/models.py
```diff
@@ -35,6 +35,13 @@ def __unicode__(self):
         return "%s (%s)" % (name, self.council)
 
 
+class VisibilityChoices(models.TextChoices):
+    # Visible on website and in API.
+    PUBLISHED = "PUBLISHED", "Published"
+    # Only visible to upload user and admin. I.e. not on website or API.
+    UNPUBLISHED = ("UNPUBLISHED", "Unpublished")
```
symroe, line 42: [formatting] is this a Ruff thing? Looks odd having braces for one and not the other!

polling_stations/apps/pollingstations/admin.py
```diff
@@ -1 +1,65 @@
...
+        "council__name",
+        "council__council_id",
+        "internal_council_id",
+    ]
+
+    # Filter on published status or council name
+    list_filter = ["visibility", "council__name"]
+
+    # Human-readable name for list view
+    list_display = ["display_name"]
```
symroe, line 25: Could council and visibility by useful here?

polling_stations/apps/file_uploads/views.py
```diff
@@ -247,13 +247,16 @@ def get_context_data(self, **kwargs):
             .values_list("polling_station_id", "uprn")
         )
 
-        for station in council_from_default_db.pollingstation_set.all():
+        for station in council_from_default_db.pollingstation_set.all().annotate(
+            pk=F("id")
+        ):
```
symroe, line 252: Can you explain what this is doing? `pk` should already be set as a property on the object, or if it's not meant to be the built in `pk` then that name might b confusing

polling_stations/apps/data_importers/event_helpers.py
```diff
@@ -1,10 +1,50 @@
...
 
 
 def record_teardown_event(council_id):
     DataEvent.objects.create(
         council=Council.objects.get(council_id=council_id),
         event_type=DataEventType.TEARDOWN,
     )
+
+
+def set_station_visibility(event):
```
symroe, line 39: Can this have the same type annotation as below?

symroe (commented):
Some inline comments, plus...

I think some more type hinting might be useful, generally scattered around.

Am I right in thinking that the only admin way to publish/unpublish is via the list display / queryset modification? Do you plan on modifying the form to allow changing one station?

symroe:
> Am I right in thinking that the only admin way to publish/unpublish is via the list display / queryset modification? Do you plan on modifying the form to allow changing one station?

No of course I'm not! Ignore!

polling_stations/apps/pollingstations/admin.py
GeoWill: Currently:
```python
    @admin.display(description="Station Display Name")
    def display_name(self, obj):
        name = obj.address.split("\n")[0]
        council_name = obj.council.name
        internal_council_id = obj.internal_council_id
        return f"{name} - {council_name} ({internal_council_id})"
```

I'll make it:
```python
return f"{name} - {council_name} ({internal_council_id}) ({visibility})"
```

polling_stations/apps/file_uploads/views.py
GeoWill: I'm not sure!  
```python
        for station in council_from_default_db.pollingstation_set.all():
            context["STATIONS"].append(
                {
                    "address": station.address,
                    "postcode": station.postcode,
                    "location": "✔️" if station.location else "❌",
                    "example_uprn": example_uprn_map.get(station.internal_council_id),
                    "pk": station.id,
                }
            )
```
works fine!

polling_stations/apps/pollingstations/admin.py
symroe: So adding items to the list creates columns in the table. 

It's a minor thing I know, so do ignore if you like, but I think it's more readable to have each field in a column. At least 

`list_display = ["display_name", "internal_council_id", "visibility"]` is a quick win, you'd have to make `council_name` another `@admin.display` method.

GeoWill: Neat! I hadn't clocked how to make columns.

GeoWill:
I'd like to deploy this to dev and check it works there. But so far this is working locally. So I'd say ready for another look.

GeoWill:
Apologies for squashing all the fixups, but I was getting too confused!

PR #6135, opened 2024-02-10
Mgmt command to generate a csv showing councils and latest EMS

GeoWill:


polling_stations/apps/councils/models.py
```diff
@@ -21,6 +22,21 @@ def with_polling_stations_in_db(self):
...
+        upload_subquery = (
+            Upload.objects.filter(gss=OuterRef("council_id"))
+            .annotate(ems=latest_ems_subquery)
+            .order_by("-timestamp")
+        )
+
+        return self.using(DEFAULT_DB_ALIAS).annotate(
+            latest_ems=Subquery(upload_subquery.values("ems")[:1]),
+            latest_upload_id=(upload_subquery.values("id")[:1]),
+        )
```
GeoWill, line 38: Based off of [this](https://stackoverflow.com/questions/2074514/django-query-that-get-most-recent-objects-from-different-categories/43926433#43926433) I'm sure there's a neater way of doing this with `Prefetch`. But I couldn't figure it out, and it's not a big table, or a frequently used query, so I think this is probably fine.

symroe: I think this is actually better than Prefetch, but you might be right that it's possible. I agree it doesn't matter :+1:

symroe (approved):

PR #6143, opened 2024-02-12
Write instance sentry DSN to global env file

symroe:
By writing here, each service on the system will get the DSN set. This means that (with a EE reboot) the onboard EE will write to the DNS for the account / application.


GeoWill (changes requested):
TIL about `/etc/environment`! 
My comment won't matter if we know the file is going to be empty, but then we can just use `'w'` instead of `'a'`.

GeoWill (approved):

deploy/hooks/afterinstall/write_global_env_vars.py
```diff
@@ -0,0 +1,23 @@
...
+    try:
+        value = client.get_parameter(Name=var)["Parameter"]["Value"]
+    except botocore.exceptions.ClientError:
+        sys.exit(f"{var} not found")
+    vars[var] = value
+
+
+with open("/etc/environment", "a") as env_file:
+    for key, value in vars.items():
+        env_file.write(f"""\n{key}="{value}"\n""")
```
GeoWill, line 23: This will leave us with blank lines between every key/value pair. I doubt this matters, but don't know that it doesn't!

PR #6415, opened 2024-03-05
Feature/council uploader list filters

GeoWill:


polling_stations/apps/councils/models.py
```diff
@@ -25,19 +25,37 @@ def with_polling_stations_in_db(self):
...
+            .annotate(all_valid=Func("csv_valid", function="BOOL_AND"))
+            .values("all_valid")
+        )
+        upload_subquery = (
+            Upload.objects.future()
+            .filter(gss=OuterRef("council_id"))
+            .annotate(all_files_valid=Subquery(all_valid_subquery))
+            .order_by("-timestamp")
+        )
+        return self.using(DEFAULT_DB_ALIAS).annotate(
```
symroe, line 41: Can you explain why you're using `using` here?

GeoWill: I mostly did it because it's what was there on the similar `with_ems_from_uploads` method, and then got distracted by everything else and never questioned it!

Without it I think it will hit RDS because the file_upload models are routed there. If that's the case it will break in a confusing way if replication isn't working, so maybe better to just let it go to RDS, though I think that will slow page loads a little. 

However, I don't actually know for sure how the routing works, and it could be the case that it would go to the local db anyway, since `councils.Council` isn't special cased in  `db_routers.py`. Do you know?

polling_stations/apps/file_uploads/filters.py
```diff
@@ -0,0 +1,51 @@
...
+        pollingstations = PollingStation.objects.filter(council_id=OuterRef("pk"))
+        if value == "1":
+            queryset = queryset.filter(Exists(pollingstations))
+        if value == "0":
+            queryset = queryset.filter(~Exists(pollingstations))
+        return queryset
+
+    def filter_upload(self, queryset, name, value):
+        queryset = Council.objects.with_future_upload_details()
+        return queryset.filter(latest_upload_status=value)
```
GeoWill, line 51: I don't know if they need it, but I wasn't that sure how to test this class.

PR #6614, opened 2024-03-15
Have a sleep if 409 on push to gh

GeoWill:
https://democracy-club-gp.sentry.io/issues/4925452960/?project=4504713373876224&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=3#

### DIFF
--- a/polling_stations/apps/file_uploads/models.py
+++ b/polling_stations/apps/file_uploads/models.py
@@ -1,3 +1,4 @@
+import time
 from datetime import timedelta
 
 from commitment import GitHubClient, GitHubCredentials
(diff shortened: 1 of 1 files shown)


symroe:
This seems like a reasonable workaround for the GitHub API, but can you have a think about what the total response time is likely to be for the request / response cycle that triggers this code?

Gunicorn will, I think, time out after 30 seconds, and I don't know if the Lambda handler will kill the request sooner.

If GitHub takes a few seconds to raise the 409, then we wait 10 and the subsequent request takes a few more seconds, are we going to time out the parent request somewhere?

GeoWill:
So the lambda handler has a timeout of 10 minutes: https://github.com/DemocracyClub/UK-Polling-Stations/blob/effba563f5da6de5ce83a5b2d14fdac6f83cecd1/cdk/stacks/wdiv_s3_trigger_stack.py#L57. 

This is started from when a file lands in the uploads bucket: 
https://github.com/DemocracyClub/UK-Polling-Stations/blob/effba563f5da6de5ce83a5b2d14fdac6f83cecd1/cdk/stacks/wdiv_s3_trigger_stack.py#L63-L65

The PR to gh is made from the instance:
https://github.com/DemocracyClub/UK-Polling-Stations/blob/effba563f5da6de5ce83a5b2d14fdac6f83cecd1/polling_stations/apps/file_uploads/api.py#L60

So I think the 30 second gunicorn timeout ([docs](https://docs.gunicorn.org/en/stable/settings.html#timeout) - I think we're using the default) is the limit we need to worry about. I think I've written it such that it will only retry once, so I think this shouldn't fall foul of that. 

Does that all sound reasonable?

symroe:
Yep that's OK. We might not get a sentry warning from the django side if we do get timeouts, but should get a warning from Lambda. I think the Lambda will retry too, so should be ok

symroe (approved):

PR #6769, opened 2024-03-22
Feature/polling station accessibility

GeoWill:
Testing this locally is tricky if you don't have a working copy of WDIV. So far I've only tried it out with one council (Cardiff) so to test locally would need to have Cardiff imported, and then import the test data I've been using which I'll share in slack. 

To see the changes go to the dev's site and to see the uploader visit:
 `/uploads/councils/CRF`, there should be a button to upload that you can click on. 

To see what user's see navigate to the example pages for the various stations. Most don't have any 'getting to the station' or 'at the station' text.
NB Stations have been imported and geocoded by postcode (which wouldn't happen on prod) to show how this looks with maps. 
Particular examples:
- With map and placeholder text `Tredegarville Primary School` (example url: `/address/200002934041/`). 
- Without map or text `Butetown Community Centre` (`address/200001800992/`)
- Without map, with text `Highfields Church` (`address/200002934250/`)

The admin page for each station has an editable form for the accessibility information. 

As you can see in the screenshots below I need to get a bunch of text translated, but I think I've done all the scaffolding needed in terms of template tags and gettext calls. 

### Bunch of screenshots:

Screenshot from 2024-04-05 07-40-32
Screenshot from 2024-04-05 07-40-41
Screenshot from 2024-04-05 07-40-49
Screenshot from 2024-04-05 07-41-01
Screenshot from 2024-04-05 07-43-50
Screenshot from 2024-04-05 07-43-39
Screenshot from 2024-04-05 07-44-56
Screenshot from 2024-04-05 07-44-34
image


polling_stations/apps/file_uploads/views.py
```diff
@@ -403,70 +398,23 @@ def test_func(self):
...
+        uploaded_file = self.request.FILES["csv_file"]
+        rows: list[str] = [
+            line.decode("utf-8").replace("\n", "") for line in uploaded_file
+        ]
+        file_handler = AccessibilityInformationHandler(self.council)
+        file_handler.handle(rows)
+        if file_handler.errors:
+            pass  # do stuff with errors
+        if file_handler.infos:
+            pass  # do stuff with infos
```
GeoWill, line 413: At the moment there's not reporting of infos (eg duplicate rows) or errors (eg missing ids) to the uploading user. I did find the console output useful in terms of debugging though. My plan had been to that if there were any infos or errors to keep the user on the upload page and display them using `ds-status-error` and `ds-status-message` and the django messages framework.
However I'm leaning towards shoving them on the model, so that the display can be more persistent rather than one time after upload. I will chat this through with @pmk01

PR #6775, opened 2024-03-22
Hotfix/dc utils footer

VirginiaDooley:
This change implements the footer changes with tags from dc_utils and fixes the right margin on the logo with a new version of design-system.

Dependent on https://github.com/DemocracyClub/dc_django_utils/pull/65 and https://github.com/DemocracyClub/design-system/pull/115
Screenshot 2024-03-25 at 11 00 10 AM

Screenshot 2024-03-22 at 7 13 51 PM


polling_stations/apps/pollingstations/tests/test_urls.py
```diff
@@ -26,6 +26,8 @@ def test_trailing_slashes(self):
         urls = [rec["url"] for rec in data]
         urls.remove("/admin/<url>")
         for url in urls:
+            print(url)
             if self.is_exception(url):
                 continue
-            assert url[-1] == "/", url + " does not end with /"
+            if url[-1] != "/":
+                print("this url is broken", url)
```
GeoWill, line 33: I like the approach :grinning:.  Can you explain why it's necessary - i.e. what's triggering the failing test, and why this is the right way to fix it.

polling_stations/templates/base.html
```diff
@@ -60,62 +60,33 @@
     {% block page_content %}
     {% endblock page_content %}
 {% endblock content %}
+{% block mailing_list %}{% endblock mailing_list %}
```
GeoWill, line 63: What does this line do?
Should it be an `include`?

polling_stations/templates/base.html
```diff
@@ -60,62 +60,33 @@
...
+            <li><a href="https://github.com/DemocracyClub/">{% trans "GitHub" context "link to Democracy Club GitHub page" %}</a></li>
+        </ul>
+    </div>
+{% endblock footer_menu %}
+{% block extra_footer_text %}
+    <div class="ds-copyright">
+        <p>
+            {% blocktrans %}Contains OS data © Crown copyright and database right {{ current_year }}{% endblocktrans %}<br>
+            {% blocktrans %}Contains Royal Mail data © Royal Mail copyright and database right {{ current_year }}{% endblocktrans %}<br>
+            {% blocktrans %}Contains National Statistics data © Crown copyright and database right {{ current_year }}{% endblocktrans %}<br>
```
GeoWill, line 85: Judging from your screen shot, I don't think `{{ current_year }}` is doing anything.

GeoWill (changes requested):

polling_stations/templates/base.html
VirginiaDooley: It removes the mailing list which is now added by default in dc_utils.

polling_stations/templates/base.html
VirginiaDooley: I've pushed the missing variable `{% now "Y" as current_year %}` so works now. Screenshot updated, too.

GeoWill:
Thanks for the changes/explanation. I missed this first time round:
`Democracy Club` changes colour when I go to Welsh. Is that expected? 
image

VirginiaDooley:
> Thanks for the changes/explanation. I missed this first time round: `Democracy Club` changes colour when I go to Welsh. Is that expected? image

Oh interesting, this might be a case (span element inside list item) that wasn't considered when we fixed url colors in a previous design system change. I wonder if I should add this to the design system or override the style here given that I think it's a unique case?

GeoWill (approved):

PR #6889, opened 2024-04-09
coords from GRT council

awdem:
coords from GRT council: https://app.asana.com/0/1205951895246773/1206787020488734/f

### DIFF
--- a/polling_stations/apps/data_importers/management/commands/import_guildford.py
+++ b/polling_stations/apps/data_importers/management/commands/import_guildford.py
@@ -41,6 +41,128 @@ class Command(BaseHalaroseCsvImporter):
 
             return rec
 
+        # coordinates from council:
+        rec = super().station_record_to_dict(record)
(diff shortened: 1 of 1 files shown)


GeoWill (changes requested):
Bit of a nit, but I don't think the brackets are doing anything in the if statements. 
I.e this:
```python
if (rec["internal_council_id"]) == "4-st-marys-church-community-centre":
```
is the same as this:
```python
if rec["internal_council_id"] == "4-st-marys-church-community-centre":
```

awdem:
> Bit of a nit, but I don't think the brackets are doing anything in the if statements. I.e this:
> 
> ```python
> if (rec["internal_council_id"]) == "4-st-marys-church-community-centre":
> ```
> 
> is the same as this:
> 
> ```python
> if rec["internal_council_id"] == "4-st-marys-church-community-centre":
> ```

Ah that's because I made a little python script to format the statements and it went through several iterations ha.

GeoWill (approved):

PR #6906, opened 2024-04-10
fix from BNH council

awdem:
fix from BNH council:  https://app.asana.com/0/1205951895246773/1207045750656123


GeoWill:
:grimacing: I got there, but didn't realise there was an asana card... https://github.com/DemocracyClub/UK-Polling-Stations/pull/6905
Sorry.

PR #6973, opened 2024-04-23
Confirm before deleting prod

symroe:
psql:
```COMMENT ON DATABASE wdiv IS 'env:production';```

Then run `teardown --all`


symroe:
Of course this is a proof of concept, I think it needs some cleaning up into something more generic, and maybe it needs to be added to some other places too.

PR #7007, opened 2024-05-01
station change from RCH council

awdem:
https://app.asana.com/0/1205951895246773/1207213161209835/f

### DIFF
--- a/polling_stations/apps/data_importers/management/commands/import_rochdale.py
+++ b/polling_stations/apps/data_importers/management/commands/import_rochdale.py
@@ -63,4 +63,15 @@ class Command(BaseXpressDemocracyClubCsvImporter):
             "5515",
         ):
             record = record._replace(polling_place_postcode="OL12 7JG")
+
+        # Station change from council:
+        # old: Mobile Unit on the Car Park of Ogden Baptist Church, Cedar Lane, Newhey  OL16 4LD
+        # new: Mobile Unit on the Car Park of Elm Grove, OL16 4LH
+        if record.polling_place_id == "5422":
+            record = record._replace(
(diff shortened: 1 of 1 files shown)


polling_stations/apps/data_importers/management/commands/import_rochdale.py
```diff
@@ -63,4 +63,15 @@ def station_record_to_dict(self, record):
...
+
+        # Station change from council:
+        # old: Mobile Unit on the Car Park of Ogden Baptist Church, Cedar Lane, Newhey  OL16 4LD
+        # new: Mobile Unit on the Car Park of Elm Grove, OL16 4LH
+        if record.polling_place_id == "5422":
+            record = record._replace(
+                polling_place_address_1="Elm Grove",
+                polling_place_address_2="",
+                polling_place_postcode="OL16 4LH",
+            )
```
GeoWill, line 75: Does this replace all the parts of the old address that need replacing? Which columns contain the info are sometimes inconsistent between rows, so maybe safer to just replace all the `polling_place_address_n` fields.

GeoWill (commented):
See inline comment

polling_stations/apps/data_importers/management/commands/import_rochdale.py
awdem: Yes it should! I printed the whole record to compare it to the new one and then replaced or removed lines as needed.

awdem: I also checked it in qgis afterwards I should add.

GeoWill (approved):

PR #7070, opened 2024-05-24
Fail CI if import scripts can't run

symroe:
Small PR to solve a simple problem:

Previously, when this line was hit (imports didn't run because a full deploy was needed), CI would see an exit code of 0 from this command. CI uses exit codes to detect if the program has run properly, so it would look like the deploy / import worked when it didn't.

I think it's better to fail CI when the import scripts can't run, as this could cause confusion otherwise.


GeoWill:
I think this might block CI from running to completion by failing early.

PR #7103, opened 2024-05-27
SPIKE: What not calling EE might look like

symroe:
This is actually not as bad as I thought it might be, but there's still stuff to do and think about.

Reasoning: 

EE doesn't currently allow us to offer an address picker if a postcode is split over two ballots. This is because EE only has ONSPD, not full AddressBase.

In other, unrelated, work, we have "baked" the list of current elections into Parquet files with a row per UPRN. This was an optimisation for an API user who only wanted elections and candidates (not polling stations), and didn't want to deal with split addresses when the address was only split over polling districts (e.g split over elections causes many fewer address pickers than split over polling district).

This spike converts WDIV to using this elections-by-address endpoint. 

In theory this has a number of advantages:

1. We don't need embedded EE / to scale EE at all. 
2. The pre-baked files are static so even though we're using the API at the moment, we could just sync them to be on-instance
3. It's much easier doing this than trying to get AddressBase into EE (and all the faff around deployment and updating that would bring)

There are some problems at the moment that need to be addresses before we can rely on this though:

1. The elections parquet files are made manually right now. This means that the elections don't go away until we manually update them. This code has some checking for future elections, but we can't just ship this and forget about it.
2. Explainers aren't in the ballot files right now. We might want to add them, or we might want to find some other way of getting them in the response. They're only used on the frontend, so at the moment we could still call the on-board EE for them. Or cache them in a JSON file on instance or something.
3. Calling the elections endpoint will cause a log line to be written. This will end up double counting postcode lookups, so we need to find a way to exclude those requests. A `?log=false` param maybe?

....And so on. I'm sure there are a load of other problems. It might not be possible to progress this before the general, but we should try as many people want us to deal with splits over constituencies.


symroe:
(And of course tests will fail as I've made no attempt to update them)

symroe:
I just smashed in the static elections client code from https://github.com/DemocracyClub/aggregator-api/blob/master/api/endpoints/v1/elections/static_elections_client.py

It worked nicely as a drop in, meaning that all we'd need to do is sync the parquet files to the instances on deploy. I've not run any benchmarking, but it seems _a lot_ faster than calling out to the API.

polling_stations/apps/data_finder/helpers/routing.py
```diff
@@ -80,6 +82,19 @@ def addresses_have_single_station(self):
             return bool(list(self.polling_stations)[0])
         return False
 
+    def lookup_elections(self):
```
GeoWill, line 95: This method (which sets `self.elections_response`) is only called in the `route_type` cached property.
Which works for how it is then patched into polling_stations/apps/api/postcode.py, but could be confusing.

Would it make sense to do something like: 

```python
def __init__(self, postcode):
     ...
    self._elections_response = None
    ...

@property
def elections_response(self):
    if not self._elections_response:
        self.lookup_elections()
    return self._elections_response
```

polling_stations/apps/data_finder/helpers/baked_data_helper.py
```diff
@@ -0,0 +1,170 @@
...
+class RemoteBakedElectionsStrategy(BaseBakedElectionsStrategy):
+    def __init__(self, api_key=None, **kwargs):
+        self.api_key = api_key or getattr(settings, "DEVS_DC_API_KEY", None)
+        if not self.api_key:
+            raise ValueError("API key required for remote backend")
+        super().__init__(**kwargs)
+
+    def get_response_for_postcode(self, postcode: Postcode):
+        # TODO: pull into a constant
+        devs_dc_base = "https://developers.democracyclub.org.uk"
```
GeoWill: Want to do this now? I don't mind.

polling_stations/apps/data_finder/helpers/baked_data_helper.py
```diff
@@ -0,0 +1,170 @@
...
+            params={"auth_token": self.api_key},
+        )
+        req.raise_for_status()
+        return req.json()
+
+
+class LocalParquetElectionsStrategy(BaseBakedElectionsStrategy):
+    def __init__(self, elections_parquet_path: Path = None, **kwargs):
+        self.elections_parquet_path = elections_parquet_path or getattr(
+            settings, "ELECTION_PARQUET_DATA_PATH", None
```
GeoWill, line 58: This is happening via parameter store?

polling_stations/apps/data_finder/helpers/baked_data_helper.py
```diff
@@ -0,0 +1,170 @@
...
+            Type[BaseBakedElectionsStrategy]
+        ] = LocalParquetElectionsStrategy,
+        **kwargs,
+    ):
+        self.strategy = strategy(**kwargs)
+
+
+if __name__ == "__main__":
+    helper = BakedElectionsHelper(strategy=None, api_key="asd")
+    helper.strategy.get_response_for_postcode(Postcode("GL5 1NA"))
```
GeoWill: delete? (Unless it's helpful)

polling_stations/apps/api/postcode.py
```diff
@@ -57,6 +60,8 @@ def generate_advance_voting_station(self, routing_helper):
     def get_ee_wrapper(self, postcode, rh, query_params):
         if rh.route_type == "multiple_addresses":
             return EmptyEveryElectionWrapper()
+        if rh.elections_response:
+            return StaticElectionsAPIElectionWrapper(rh.elections_response)
         kwargs = {}
         query_params = parse_qs_to_python(query_params)
         if include_current := query_params.get("include_current", False):
```
GeoWill, line 65: I don't think there's anything we can do about this, but as long as there are parquet files I think these lines won't be hit. Do we rely on the `include_current` flag somewhere else when we're calling the WDIV API?

GeoWill (commented):
At the moment this is looking good on the API (`api/beta/address/77145275/` vs `api/beta/address/10070392100/`), and nothing is showing on the front end (as expected). 
Few questions inline.

polling_stations/apps/api/postcode.py
symroe: Decided to park until later

polling_stations/apps/data_finder/helpers/baked_data_helper.py
symroe: Done, although the `RemoteBakedElectionsHelper` isn't actually uses anywhere at the moment. I'll keep it in for now, but we might never need it

PR #7236, opened 2024-05-30
Initial system for patching council contact details based on election ID

symroe:
This allows us to change contact details based on a ballot paper ID.

Designed for short-term patching of councils in Scotland.


symroe:
Clearly we need the actual content, but I think this deals with the actual logic

GeoWill (approved):

PR #7729, opened 2024-06-11
Post to Asana

symroe:
Totially tested this on a _local_ database, like, no problems with that at all.

Might be worth testing on, er, a different local database with some dummy data.


symroe:
Need to add `asana` as a dependancy, but unsure of the best way

polling_stations/settings/constants/asana.py
```diff
@@ -10,6 +10,105 @@
 
 ASANA_REPORT_TYPE_FIELD_ID = "1207543086662142"
 
+ASANA_OPT_FIELDS = [
```
symroe, line 13: FWIW I don't know what these do, it was just copy paste from an example!

PR #7764, opened 2024-06-13
WiP

GeoWill:


polling_stations/apps/data_importers/data_types.py
```diff
@@ -375,11 +383,72 @@ def check_split_postcodes_are_split(self, split_postcodes):
...
+        self, polling_station_id, uprns_assigned_to_station
+    ):
+        # At this stage we want to know if the station has the right council id.
+        # There are three cases:
+        # 1. All the addresses are in the council named in the import script.
+        #    So the station will have the correct council_id and no action is necessary.
+        # 2. All the addresses are in a different council.
+        #    In this case we need to update the council_id on the polling station in the pollingstations table
+        # 3. The addresses assigned to this station are in different council areas.
+        #    In this case we need to duplicate the station making sure there are a record for each council_id.
```
GeoWill, line 413: This approach won't work, because the polling stations aren't saved in the DB yet.

PR #8077, opened 2024-08-01
trigger a build

chris48s:


chris48s:
Hmm. Does this project not run tests on PRs? :confused:

chris48s:
I reckon this might be https://github.com/pypa/pip/issues/12884

chris48s:
OK, so turns out this is actually fixed in rtree==1.3.0
New solution: upgrade rtree, leave pip as it was.

requirements/constraints.txt
```diff
@@ -1691,12 +1689,6 @@ sentry-sdk==2.1.1 \
     # via
     #   -r requirements/../cdk/lambdas/wdiv-s3-trigger/requirements/base.in
     #   -r requirements/base.in
-setuptools==69.2.0 \
```
chris48s, line 1694: I re-generated the requirements files using:

```
pip-compile --allow-unsafe --generate-hashes --output-file=requirements/constraints.txt --strip-extras requirements/constraints.in --upgrade-package rtree
```

to re-lock `requirements/constraints.txt`

and then

```
pip-compile --allow-unsafe --generate-hashes --output-file=requirements/base.txt --strip-extras requirements/base-constrained.in --upgrade-package rtree
```

to re-lock `requirements/base.txt`

I'm not sure why pip-tools took `setuptools` out of the lockfile here. Tbh, your pip-tools setup on this repo is pretty fiendish and also a bit under-specified. I think I'm going to need a bit of a primer on this before I can really do substantial work on the dependency tree on this repo. From memory, I feel like whether to include setuptools in lock files is something pip-tools went back on forth on a bit in different releases. There is no specific pip-tools version specified in the requirements or docs. I locked this using `pip-tools==7.4.1`. Let me know if you want me to re-lock using a different `pip-tools` version. Also if I didn't use the right combination of commands..

symroe (approved):

PR #8086, opened 2024-08-15
add additional CI checks

chris48s:


chris48s:
bah also need to load this into https://github.com/DemocracyClub/UK-Polling-Stations/pull/8082

PR #8137, opened 2024-10-25
reduce cache length on election intersecting local auth query

chris48s:
Refs https://app.asana.com/0/1204880927741389/1208612753966995/f

This should mean we're not making this query too often, but if we hit a state where we've cached an unhelpful value it won't stick around for an inconveniently long time


GeoWill (approved):
Looks good.

PR #8147, opened 2024-11-06
Show different registration info/polling station opening times for City of London local elections

chris48s:
Refs
- https://app.asana.com/0/1204880927741389/1208540419887296/f
- https://app.asana.com/0/1204880927741389/1208687275764618/f
- https://app.asana.com/0/1204880927741389/1208687275764619/f

This PR is part of the big pile of "deal with City of London edge cases" issues.

This is quite hard to do well on WDIV because it has the least nuanced view of "what election(s) are we telling the user about?"

In this PR I am looking at electoral registration details and polling station opening times but not ID requirements.

### DIFF
--- a/polling_stations/apps/data_finder/helpers/every_election.py
+++ b/polling_stations/apps/data_finder/helpers/every_election.py
@@ -287,6 +287,18 @@ class EveryElectionWrapper:
             return len(uncancelled_ballots) > 1
         return False
 
+    @property
+    def has_city_of_london_ballots(self):
+        # City of London local elections have some edge cases e.g:
+        # diffrent polling station opening times, different registration rules
+        if not self.request_success:
+            return False
+        uncancelled_ballots = [b for b in self.ballots if not b["cancelled"]]
+        for b in uncancelled_ballots:
+            if b["election_id"].startswith("local.city-of-london"):
+                return True
+        return False
+
 
 class EmptyEveryElectionWrapper:
     """
@@ -321,6 +333,10 @@ class EmptyEveryElectionWrapper:
 
     def get_cancelled_election_info(self):
         return {}
+
+    @property
+    def has_city_of_london_ballots(self) -> bool:
+        return False
 
 
 class StaticElectionsAPIElectionWrapper:
@@ -445,3 +461,13 @@ class StaticElectionsAPIElectionWrapper:
             if b["election_id"] not in settings.ELECTION_BLACKLIST
         ]
         return sorted(ballots, key=lambda k: k["poll_open_date"])
+
+    @property
+    def has_city_of_london_ballots(self):
+        # City of London local elections have some edge cases e.g:
+        # diffrent polling station opening times, different registration rules
+        uncancelled_ballots = [b for b in self.ballots if not b["cancelled"]]
+        for b in uncancelled_ballots:
+            if b["election_id"].startswith("local.city-of-london."):
+                return True
+        return False
--- a/polling_stations/apps/data_finder/helpers/every_election.py
+++ b/polling_stations/apps/data_finder/helpers/every_election.py
@@ -294,27 +294,23 @@ class EmptyEveryElectionWrapper:
     when we are going to show an address picker.
 
     This class allows us to swap out the EveryElectionWrapper while keeping
-    exisitng code the same.
+    existing code the same.
     """
 
-    @staticmethod
-    def has_election() -> bool:
+    def has_election(self) -> bool:
         return False
 
-    @staticmethod
-    def get_metadata() -> None:
+    def get_metadata(self) -> None:
         return None
 
-    @staticmethod
-    def get_ballots_for_next_date() -> List:
+    def get_ballots_for_next_date(self) -> List:
         return []
 
-    @staticmethod
-    def get_all_ballots() -> List:
+    def get_all_ballots(self) -> List:
         return []
 
-    @staticmethod
-    def multiple_elections():
+    @property
+    def multiple_elections(self):
         return False
 
     def get_explanations(self):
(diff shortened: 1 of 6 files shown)


polling_stations/apps/data_finder/helpers/every_election.py
chris48s, line 1: Everything in this file made me very sad.

- We've got 3 different classes with the same methods but no way to sure they stay in sync
- Two of them have nearly the exact same code copy & pasted between them (again with opportunities to drift)
- There's a completely random mashup of properties and methods, leading to confusion (see #8148 !)
- The methods don't seem to behave very consistently. For example, some are looking at all future ballots but some are looking at ballots for next date. I suspect this isn't for good reasons.

I've resisted the urge to tinker with it too much, but I'd really like to rethink the approach to this. For the moment I've just put this fire with the other fire, which is basically how we've ended up here.

polling_stations/templates/fragments/register_to_vote.html
chris48s, line 1: Currently we aren't showing any register to vote information on WDIV because it is commented out
https://github.com/DemocracyClub/UK-Polling-Stations/blob/b9e5eb5fe2c43ff317b2ee47a349613394a274f2/polling_stations/templates/postcode_view.html#L42
On WDIV we add/remove this manually. I've just added this text alongside the standard CTA if the user has >1 City of London local elections upcoming.

polling_stations/templates/fragments/polling_station_unknown.html
```diff
@@ -29,4 +29,10 @@ <h2>{% blocktrans %}Contact {{ council }}{% endblocktrans %}</h2>
     {% endif %}
 </p>
 
-<p><strong>{% blocktrans %}Polling stations are open from 7am to 10pm.{% endblocktrans %}</strong></p>
+<p><strong>
+    {% if has_city_of_london_ballots %}
+        {% blocktrans %}Polling stations are open from 8am to 8pm.{% endblocktrans %}
+    {% else %}
+        {% blocktrans %}Polling stations are open from 7am to 10pm.{% endblocktrans %}
+    {% endif %}
```
chris48s, line 37: This approach to polling station opening hours may end up being wrong in situations where there are multiple election types upcoming. For example, a City of London by-election is scheduled a week before GLA elections or something.

I think we really need to move to WDIV being more specific about what election(s) we're talking about to properly fix this, which is a bigger job.

polling_stations/apps/data_finder/helpers/every_election.py
```diff
@@ -287,6 +287,18 @@ def multiple_elections(self):
...
 
+    @property
+    def has_city_of_london_ballots(self):
+        # City of London local elections have some edge cases e.g:
+        # diffrent polling station opening times, different registration rules
+        if not self.request_success:
+            return False
+        uncancelled_ballots = [b for b in self.ballots if not b["cancelled"]]
+        for b in uncancelled_ballots:
+            if b["election_id"].startswith("local.city-of-london"):
```
GeoWill, line 298: This will catch elections for `https://elections.democracyclub.org.uk/organisations/local-authority/LND-alder/1189-01-01/` and `https://elections.democracyclub.org.uk/organisations/local-authority/LND/1189-01-01/` I assumed this is desired, but thought it worth double checking

polling_stations/templates/fragments/polling_station_unknown.html
GeoWill: I agree this is isn't the place to fix this, but we do get a lot of people looking at our site using City of London postcodes even though they don't live there. Do you think it's worth changing the text to make it clear that the 8 to 8 rule is CoL specific?

Something like:

> Polling stations for City of London elections are oprn from 8am to 8pm. Elections to other organisations (eg GLA or Parliament) polling stations will be open from 7am to 10pm.

polling_stations/apps/data_finder/helpers/every_election.py
chris48s: Yes. Both Alderman and Common Council have the same rules here.

polling_stations/templates/fragments/polling_station_unknown.html
chris48s: I've added some extra clarifying text based on this suggestion in https://github.com/DemocracyClub/UK-Polling-Stations/pull/8147/commits/02e2ccee4b7a0fbd3e111efd9d461ec604ddd15a

chris48s:
There is actually something I meant to mention when I originally wrote this PR that slipped my mind.

There is another place we show polling station opening times on WDIV. When polling day is today, we show this on the home page:

https://github.com/DemocracyClub/UK-Polling-Stations/blob/31f62e60421e7a0236ccdec1bef4d7979f2fd472/polling_stations/templates/home.html#L49-L57

This includes the opening hours, but because we haven't collected a postcode/address yet we can't tailor it based on the ballot(s). I wasn't really sure what to do with this.

One option would just be to bin it and only show opening times once the user has performed a search.
We could leave it as-is, which will mean if you're in the City of London it is wrong on the home page but right after you search.
Or we could show both sets of opening times the home page (despite the fact that this special case for City of London will be irrelevant to 99% of users and possibly just cause confusion).

GeoWill:
I think either leave it as is or get rid. 
I err slightly to leaving it, but don't feel very strongly about it.

GeoWill (approved):

chris48s:
Just having thought this through a bit: The next scheduled City of London elections aren't on the same date as the main County/Mayor ones so we can special case this. If we aren't going to set 2025-03-20 as a "charismatic" date, this doesn't matter. If we are then we probably want to do something special. I'm going to punt this to an Asana card.

PR #8153, opened 2024-11-14
Bump uk-geo-utils

GeoWill:
I've shoved this on dev environment and successfully imported onspd and addressbase.

Once this is merged we should run those commands.

### DIFF
--- a/polling_stations/apps/addressbase/management/commands/update_addressbase.py
+++ b/polling_stations/apps/addressbase/management/commands/update_addressbase.py
@@ -100,6 +100,11 @@ class Command(BaseCommand):
             "--uprntocouncil-s3-uri",
             help="S3 URI for UPRN to Council data file",
         )
+        parser.add_argument(
+            "--database",
+            default=PRINCIPAL_DB_NAME,
+            help="Database name. Defaults to PRINCIPAL_DB_NAME - i.e. RDS if you're on EC2",
+        )
 
     def teardown(self):
         self.stdout.write(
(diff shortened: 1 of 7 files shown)


chris48s:
Do you think it makes sense to convert `update_addressbase` to use a `database` param so it has a consistent interface/behaviour with the commands from the uk-geo-utils package, or just note in the docs when we need `--database` and when we don't?

GeoWill:
It's a good shout. I was being lazy/not fixing what wasn't broken. 
I've always struggled to get multi db testing working in WDIV. I had another bash, but stopped after what felt like a proportionate amount of time. If you think it does need more testing maybe we can have a look together. For now I've stashed the work locally.

polling_stations/apps/addressbase/management/commands/update_addressbase.py
```diff
@@ -76,8 +76,8 @@ class Command(BaseCommand):
         "Updates both Addressbase and UPRN to Council mapping tables from local files"
     )
 
-    def __init__(self, stdout=None, stderr=None, no_color=False, force_color=False):
-        super().__init__(stdout, stderr, no_color, force_color)
+    def __init__(self):
+        super().__init__()
```
chris48s, line 80: I think I'd make this

```suggestion
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
```

chris48s:
I've left one comment inline.

I don't think we need to worry too much about the tests for the sake of merging this.

More generally..

The first thing we would need to do is make sure our test settings file has multiple DB connections configured in it. So
https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/pyproject.toml#L29
https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/settings/testing.py

Then the second thing is telling django about the multiple DBs you are using:
https://docs.djangoproject.com/en/4.2/topics/testing/tools/#multi-database-support

IF you've already got that far and hit other problems, happy to have a look at it sometime but maybe lets make it another ticket for another day.

PR #8161, opened 2024-11-19
Wrap the import of eoni uprns in a transaction

GeoWill:


symroe (approved):

GeoWill:
This doesn't actually fix the bug I hoped it would. So I'm going to close it. It might be useful in another way. Sorry!

GeoWill:
Actually this might be the fix to the sentry issues. I was thrown off by the command runner cloud watch logs.

PR #8168, opened 2024-11-21
Feature/import eoni slack report

GeoWill:
Need to set `SLACK_TOKEN` in parameter store.  But tested in development env and posted to `#bot-testing` channel in slack successfully.

### DIFF
--- a/polling_stations/apps/core/slack_client.py
+++ b/polling_stations/apps/core/slack_client.py
@@ -1,0 +1,63 @@
+import sys
+
+from slack_sdk import WebClient
+from slack_sdk.errors import SlackApiError
+
+import os
+from typing import Optional, Dict, Any
+
+
+class SlackChannelNotFoundError(Exception):
+    pass
+
+
+class SlackClient:
+    def __init__(self, token: Optional[str] = None, channel: str = "bots", stdout=None):
+        self.stdout = stdout or sys.stdout
+        self.token = token or os.getenv("SLACK_TOKEN")
+        if not self.token:
+            raise ValueError(
+                "Slack token not provided and SLACK_TOKEN not found in environment"
+            )
+
+        self.client = WebClient(token=self.token)
+        self.channel = channel
+        self.channel_id = self.get_channel_id_from_name(self.channel)
+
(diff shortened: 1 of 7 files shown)


polling_stations/apps/core/slack_client.py
```diff
@@ -0,0 +1,59 @@
...
+    def get_channel_id_from_name(self, channel_name: str) -> Optional[str]:
+        conversation_id = None
+        try:
+            conversations_response = self.client.conversations_list()
+            for channel in conversations_response.data["channels"]:
+                if channel["name"] == channel_name:
+                    conversation_id = channel["id"]
+                    return conversation_id
+        except SlackApiError as e:
+            print(f"Error: {e}")
```
symroe, line 33: Tiny nit-pick here, but might be a useful pattern to think about. It's tricky to test this output as it's going to `stdout`. It might be nicer to set `self.stdout = sys.stdout` in the `__init__` and allow passing a stdout into the class. Then in tests we can pass a `StringIO` and read from it. All calls to `print` can be converted to `self.stdout.write`.

polling_stations/apps/core/tests/test_slack_client.py
```diff
@@ -1,5 +1,6 @@
 from unittest.mock import Mock, patch
 import pytest
+from six import StringIO
```
symroe, line 3: Wrong import here...we don't want to rely on `six`, so I think this is `io.StringIO` now

polling_stations/apps/core/slack_client.py
```diff
@@ -10,7 +12,8 @@ class SlackChannelNotFoundError(Exception):
 
 
 class SlackClient:
-    def __init__(self, token: Optional[str] = None, channel: str = "bots"):
+    def __init__(self, token: Optional[str] = None, channel: str = "bots", stdout=None):
+        self.stdout = sys.stdout
```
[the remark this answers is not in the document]
GeoWill: I went for `self.stdout = stdout or sys.stdout` which I think is right

symroe (approved):

PR #8169, opened 2024-11-21
Keep LAST_IMPORT_SHA on master for development env

GeoWill:
This should fix and issue we were having where dev deploys would 'fail' CI because the 'LAST_IMPORT_SHA' in parameter store had been updated to a value which wasn't in the tree anymore, meaning we couldn't git diff against it.

### DIFF
--- a/cdk/lambdas/wdiv-s3-trigger/tests/test_handler.py
+++ b/cdk/lambdas/wdiv-s3-trigger/tests/test_handler.py
@@ -22,7 +22,7 @@ trigger_payload_str = """{
     {
       "eventVersion": "2.0",
       "eventSource": "aws:s3",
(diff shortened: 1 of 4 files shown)


GeoWill:
Good points. 
This does what you suggest, which is also easier to test! 
I changed the region we use in tests to eu-west-2 which meant touching the `wdiv-s3-trigger` tests too.

GeoWill:
Few more tidy ups. Now when the command tries to get the `LAST_IMPORT_SHA` from parameter store it checks if the sha exists in the tree. If it doesn't and the environment is `development` it just returns the sha for the HEAD of master branch. If the sha doesn't exist in the tree and the environment is something else, it throws an error.

symroe (approved):

PR #8186, opened 2024-12-11
Feature/show eoni stations when theres no election

GeoWill:
This should mean we show polling stations in Northern Ireland when we know about them, but there's no election. 
When there is an election we should have the same behaviour in Northern Ireland as the rest of UK. 

I guess another change would be to explain which scheme NI stations are, when there is an election. But I would probably punt that to another PR.

We can turn the feature on and off by using the  `SHOW_EONI_STATIONS_ALL_THE_TIME` setting. At the moment this is set to `True`, but  would suggest we initially merge this to master with it set to `False`. We can then deploy to the dev site andchange the setting there, to get thumbs up from EONI before progressing. Alternatively we just deploy this branch to dev without merging.

### DIFF
--- a/locale/cy/LC_MESSAGES/django.po
+++ b/locale/cy/LC_MESSAGES/django.po
@@ -70,6 +70,14 @@ msgstr "Cod post:"
 msgid "Enter your postcode"
 msgstr "Rhowch eich cod post"
 
+#: polling_stations/apps/data_finder/views.py:186
+msgid "local Councillor"
+msgstr ""
+
+#: polling_stations/apps/data_finder/views.py:187
+msgid "Member of the Legislative Assembly, and Member of Parliament"
+msgstr ""
+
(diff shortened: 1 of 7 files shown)


chris48s:
Having done a bit of testing with this locally, I think this is good as it stands.

One thing I just want to double check here is: What we've done affects the WDIV frontend only. The API is still going to say `"polling_station": null` if you don't have an upcoming election. I think this is correct, but just want to double check we agree that is the desired behaviour.

In the course of reviewing this, I did spot a couple of other things which I've chucked into the backlog:
- https://app.asana.com/0/1204880927741389/1208958646411295/f
- https://app.asana.com/0/1204880927741389/1208958646411297/f

but they're not really strictly related to this and shouldn't block from merging

chris48s (approved):

PR #8187, opened 2024-12-12
Add flag to teardown NI stations

GeoWill:
This just gives us a short hand way of tearing down the Northern Irish Councils. 
The rational for this is that if there are unexpected scheme changes it would be nice to have a quick way of saying 'we don't know' while we make sure the right data is up.

### DIFF
--- a/polling_stations/apps/data_importers/management/commands/teardown.py
+++ b/polling_stations/apps/data_importers/management/commands/teardown.py
@@ -32,7 +32,7 @@ class Command(BaseCommand):
             "-c",
             "--council",
             nargs=1,
-            help="Council ID to clear in the format X01000001",
+            help="Council ID to clear, uses three letter council codes, e.g. 'ABC'",
         )
 
         group.add_argument(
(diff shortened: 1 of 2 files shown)


polling_stations/apps/data_importers/management/commands/teardown.py
```diff
@@ -59,26 +87,11 @@ def handle(self, *args, **kwargs):
...
-                dq.num_stations = 0
-                dq.save()
-
-                record_teardown_event(council_id)
-                print("..done")
+                self.teardown_council(council_id)
+
+        elif kwargs.get("eoni"):
+            for council_id in NIR_IDS:
+                self.teardown_council(council_id)
```
chris48s, line 94: In the help for the `--council` flag it says

`Council ID to clear in the format X01000001`

but `NIR_IDS` is a list of 3 letter codes: ABC, AND, ANN...

This seems like something is slightly amiss as we are passing the same thing into `teardown_council` either way. Have we modified teardown so that `--council` now expects a 3 letter code? If so, lets update the help.

GeoWill: Good shout. 
This was a side effect of starting to use reg codes as council ids, and moving gss codes to the `identifiers` field on the council geography model.

chris48s (approved):

PR #8191, opened 2024-12-17
Fix slack setting in import eoni from s3 cmd

GeoWill:
The settings object was a LazySettings, and the constants weren't showing up.

I also renamed them to hopefully make typos harder.


chris48s:
I am always really confused about when the lazy settings thing is/isn't a problem, but this looks like it will be not-broken

chris48s (approved):

PR #8192, opened 2024-12-17
Run import eoni as polling_stations user

GeoWill:
Last fix wasn't enough.
I've Just tried this out in dev env and it ran to completion.

### DIFF
--- a/polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py
+++ b/polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py
@@ -8,6 +8,7 @@ from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
 from pyproj import Transformer
 
+from core.slack_client import SlackClient
 from data_importers.s3wrapper import parse_s3_uri, S3Wrapper
 from data_importers.management.commands.import_eoni import Command as EONI_Importer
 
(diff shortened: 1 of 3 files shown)


polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py
```diff
@@ -420,15 +434,22 @@ def call_import_eoni_command(self, options):
...
+        slack_client = SlackClient(channel=self.slack_channel)
+        try:
+            response = slack_client.send_message(
+                message=":warning: *import_eoni_from_s3 command failed*",
+            )
+            slack_client.send_message(
+                message=f"Error: {str(exception)}", thread_ts=response.get("ts")
+            )
+        except Exception as slack_e:
+            self.stderr.write(f"Error posting to Slack: {str(slack_e)}")
```
chris48s, line 451: I think we want to `raise slack_e` here?

polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py
```diff
@@ -196,6 +197,12 @@ def handle(self, *args, **options):
             # Run the import
             self.call_import_eoni_command(options)
 
+        except Exception as e:
```
chris48s, line 200: If you really want to catch generic exception, you at least need to exclude `KeyboardInterrupt` and `SystemExit`. The pattern I'd usually use for this is:

```py
try:
    # generic thing that might fail in an unknown way
except (KeyboardInterrupt, SystemExit):
    raise
except Exception as err:
    # exception behaviour
    # maybe re-raise
```

same for the other place

polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py
GeoWill: Yes, your right here. Then the command will be marked as failed in the AWS command runner console,.

PR #8214, opened 2025-01-23
bin concept of custom finders

chris48s:
Refs https://app.asana.com/0/1204880927741389/1208958646411295/f


polling_stations/apps/api/postcode.py
```diff
@@ -129,10 +116,9 @@ def retrieve(self, request, postcode=None, format=None, geocoder=geocode, log=Tr
             if ret["polling_station"] and not ret["council"]:
                 ret["council"] = ret["polling_station"].council
 
-        # get custom finder (if no polling station)
+        # this feature doesn't exist any more, but the key is
+        # still in the API response for legacy compatibility
         ret["custom_finder"] = None
```
chris48s, line 121: For the sake of not making a breaking API change, I am going to keep the key in the object but it is just always null in 100% of cases.

chris48s:
Properly removing this has knock-on effects in a bunch of places that have code to consume the value which we can also remove.

Again for the sake of not making a breaking change, I think I want to leave the key in the object on devs.DC but I do want to remove stuff like https://developers.democracyclub.org.uk/api/v1/#notes-and-tips-2 and just say "this field used to be a thing, but now it is always null" or whatever.

polling_stations/apps/api/postcode.py
symroe: Don't forget that this is basically not a supported API. If the devs.DC code doesn't consume this, then there's nothing to break. If it does, then I suggest we come back and remove this when that code doesn't rely on it.

symroe:
I've not checked everywhere in the code here, but this looks fine. 

If we're only keeping the field in the API for public reasons, then I'd rather we do that in the devs.DC code rather than here. 

There are general problems with shoving an object from this service into devs.DC and then re-broadcasting it, so let's not make things harder for ourselves by supporting legacy properties across them both

symroe (approved):

polling_stations/apps/api/postcode.py
chris48s: Yeah I think I want to get rid of any code that might be expecting this value to exist in devs.DC first then come back and remove it. In some ways, I may have approach this backwards.

polling_stations/apps/api/sandbox-responses/AA12AA.json
```diff
@@ -11,7 +11,6 @@
         },
         "type": "Feature"
     },
-    "custom_finder": null,
```
chris48s, line 14: Probably not one to deal with in this PR, but I wonder if it is actually useful to keep maintaining sandbox responses on this repo now it is private?

PR #8224, opened 2025-02-04
Explicitly specify DB connection when we need to, and don't when we don't need to

chris48s:
Refs https://app.asana.com/0/1204880927741389/1208838937595665/f

This PR is a bit of a mashup of several different things. There are 3 core things I'm doing in this PR.

Prior to #8221 our "solution" to this problem was to pepper `.using()` and `using=` across scripts we thought we might want to run via SSH. #8221 should make this less necessary (but not completely unnecessary - we'll come on to this) so one of the things I'm doing in this PR is removing a lot of that stuff from places where we had done it explicitly and relying on the DB router to route queries to the right DB.

Then that said.. Django's DB routers are a somewhat leaky abstraction. They apply to ORM operations, but that doesn't mean they apply to all database I/O. Specifically there are 2 things we need to worry about:

**connection.cursor**

This one is kind of obvious if you think about it. The DB router expects a model class as one of the inputs so anything where you're running raw SQL doesn't make sense in that world. There is also no way to enforce `allow_relation` if we're working with raw SQL so if you think about it, it is obvious that the DB routers won't work if you're using raw SQL.
There's a thread about this on the django community forum https://forum.djangoproject.com/t/allow-database-router-to-override-default-connection/34529
Anyway, I can't see this changing soon. Also, we use raw SQL in several places in this application. Usually for valid reasons. So we do need to think about this.
In this PR, I've broadly done 2 things:
- Where it is possible to remove raw SQL and replace with ORM calls I have done that. Replacing raw SQL where we don't strictly need it puts us back on the happy path and allows the DB routers to do their thing.
- Where it is not possible to remove raw SQL (either because the ORM doesn't implement stuff we need or for performance reasons) I've made sure we are explicitly specifying the DB connection when using `.cursor()` to execute SQL because if we don't explicitly specify it, django just uses the default DB connection.

**transaction.atomic**

This one is perhaps less obvious.
Regardless of what your database routers say, transaction atomic blocks use the default DB connection unless you explicitly specify it with `using=`.

https://github.com/django/django/blob/5a2c1bc07d126ce32efaa157e712a8f3a7457b74/django/db/transaction.py#L182-L183
https://github.com/django/django/blob/5a2c1bc07d126ce32efaa157e712a8f3a7457b74/django/db/transaction.py#L18-L25

In our case, because the default DB is the replica rather than the primary in production this means that when we write code like this:

```py
@transaction.atomic
def do_the_things():
    this = This(foo='bar')
    that = That(foo='bar')
    this.save()
    that.save()
```

The transaction applies to database IO over the replica connection but the writes we are doing are happening on the primary.
So it looks like that code is safely wrapped in a transaction, but it isn't really when we run it in prod :scream:

Maybe this is telling us we should actually make the primary DB the default connection? (I accept there are tradeoffs either way round). For now I have chosen to just explicitly pass a `using=` to every `transaction.atomic` so that transactions actually working as intended·

Anyway, the point is: If we're using raw SQL or transactions it is up to us to manually specify what connection we're using or django will just use the default and it doesn't matter what we put in DB routers.

What I **haven't** attempted to do in this PR is prevent us from writing wrong code in future. That is another issue/PR.

The scope of this PR is only to make sure all the code we have already written is talking to the correct DB and remove code that is no longer needed following #8221

I have attempted to split this down into commits which are easy to digest one at a time.

### DIFF
--- a/polling_stations/apps/data_importers/data_types.py
+++ b/polling_stations/apps/data_importers/data_types.py
@@ -8,9 +8,9 @@ from collections import namedtuple
 
 from addressbase.models import Address, UprnToCouncil, get_uprn_hash_table
 from councils.models import Council
-from django.db import connection
 from pollingstations.models import PollingDistrict, PollingStation
 from uk_geo_utils.helpers import Postcode
+from polling_stations.db_routers import get_principal_db_connection
 
 Station = namedtuple(
     "Station",
@@ -116,7 +116,7 @@ class DistrictSet(CustomSet, AssignPollingStationsMixin):
         self.saved = True
 
     def get_uprns_by_district(self):
-        cursor = connection.cursor()
+        cursor = get_principal_db_connection().cursor()
         cursor.execute(
             """
                 SELECT a.uprn, d.polling_station_id
@@ -133,7 +133,7 @@ class DistrictSet(CustomSet, AssignPollingStationsMixin):
         return cursor.fetchall()
 
     def get_uprns_by_district_join_stations(self):
-        cursor = connection.cursor()
+        cursor = get_principal_db_connection().cursor()
         cursor.execute(
             """
                 SELECT a.uprn, s.internal_council_id
--- a/polling_stations/apps/file_uploads/models.py
+++ b/polling_stations/apps/file_uploads/models.py
@@ -8,7 +8,6 @@ from django.contrib.auth.models import User
 from django.contrib.gis.db import models
 from django.contrib.postgres.aggregates import BoolAnd, StringAgg
 from django.core.mail import EmailMessage
-from django.db import transaction
 from django.db.models import Case, Exists, Func, OuterRef, Q, Value, When
 from django.template.loader import render_to_string
 from django.utils import timezone
@@ -219,7 +218,6 @@ class Upload(models.Model):
         )
         email.send()
 
-    @transaction.atomic
     def send_error_email(self):
         subject = "File upload failed"
         message = f"File upload failure: {self}. Please investigate further."
--- a/polling_stations/db_routers.py
+++ b/polling_stations/db_routers.py
@@ -3,6 +3,7 @@ import os
 from django.conf import settings
 from django.db import DEFAULT_DB_ALIAS
 from django_middleware_global_request import get_request
+from django.db import connections
 
 
 PRINCIPAL = settings.PRINCIPAL_DB_NAME
@@ -54,3 +55,7 @@ def get_principal_db_name():
     if PRINCIPAL in settings.DATABASES:
         return PRINCIPAL
     return REPLICA
+
+
+def get_principal_db_connection():
+    return connections[get_principal_db_name()]
--- a/polling_stations/apps/addressbase/management/commands/create_uprn_council_lookup.py
+++ b/polling_stations/apps/addressbase/management/commands/create_uprn_council_lookup.py
@@ -1,7 +1,7 @@
 from pathlib import Path
 
 from django.core.management.base import BaseCommand
-from django.db import connection
+from polling_stations.db_routers import get_principal_db_connection
 
 
 class Command(BaseCommand):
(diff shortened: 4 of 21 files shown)


polling_stations/db_routers.py
```diff
@@ -29,3 +30,7 @@ def get_principal_db_name():
     if settings.PRINCIPAL_DB_NAME in settings.DATABASES:
         return settings.PRINCIPAL_DB_NAME
     return DEFAULT_DB_ALIAS
+
+
+def get_principal_db_connection():
+    return connections[get_principal_db_name()]
```
chris48s, line 36: A couple of notes on this.

1. I added this function as discussed, but given we have to pass the connection name in as a string when we use `transaction.atomic()` I'm not sure it is actually that useful. It might be more consistent to just use the string name only for both purposes.
2. To make this PR something we can review independently of the commits in #8221 I've slightly falsified the commit history here. We'll need to resolve a conflict with that PR if we keep the function. When resolved, the "real" diff for this file will be:

```diff
diff --git a/polling_stations/db_routers.py b/polling_stations/db_routers.py
index 0a03ff41..dd41e3ca 100644
--- a/polling_stations/db_routers.py
+++ b/polling_stations/db_routers.py
@@ -3,6 +3,7 @@ import os
 from django.conf import settings
 from django.db import DEFAULT_DB_ALIAS
 from django_middleware_global_request import get_request
+from django.db import connections
 
 
 PRIMARY = settings.PRINCIPAL_DB_NAME
@@ -54,3 +55,7 @@ def get_principal_db_name():
     if PRIMARY in settings.DATABASES:
         return PRIMARY
     return REPLICA
+
+
+def get_principal_db_connection():
+    return connections[get_principal_db_name()]
```

polling_stations/apps/file_uploads/models.py
```diff
@@ -219,7 +218,6 @@ def send_confirmation_email(self):
         )
         email.send()
 
-    @transaction.atomic
     def send_error_email(self):
         subject = "File upload failed"
         message = f"File upload failure: {self}. Please investigate further."
```
chris48s, line 223: As far as I can tell from reading this function, this transaction wasn't really doing anything so I just removed it.

polling_stations/apps/data_importers/data_types.py
```diff
@@ -116,7 +116,7 @@ def save(self):
         self.saved = True
 
     def get_uprns_by_district(self):
-        cursor = connection.cursor()
+        cursor = get_principal_db_connection().cursor()
         cursor.execute(
             """
                 SELECT a.uprn, d.polling_station_id
```
chris48s, line 122: Potentially the 2 SQL statements in this file could be rewritten using the ORM. I think you could probably replace
`get_uprns_by_district()` with either

```py
def get_uprns_by_district(self):
    districts = PollingDistrict.objects.filter(
        council_id=self.council_id, area__contains=OuterRef("location")
    ).values("polling_station_id")[:-1]

    return (
        Address.objects.filter(
            uprntocouncil__lad=self.gss_code,
        )
        .annotate(polling_station_id=Subquery(districts))
        .values_list("uprn", "polling_station_id")
    )
```

OR

```py
def get_uprns_by_district(self):
    districts = PollingDistrict.objects.filter(
        council_id=self.council_id, area__contains=OuterRef("location")
    ).values("polling_station_id")

    return (
        Address.objects.filter(
            uprntocouncil__lad=self.gss_code,
        )
        .annotate(polling_station_id=Subquery(districts))
        .values_list("uprn", "polling_station_id")
    )
```

but I couldn't quite get my head round which and there's about 10 years of thinking about edge cases that I don't want to untangle here. I settled for just specifying the connection and sticking with raw SQL for these two.

chris48s:
I wanted to flag a couple of places where I explicityl chose not to change anything:

1. There are a handful of places in the councils app where we have explicit `.using()` and `using=` calls which I haven't unpicked in this PR. I think I'm going to leave those to query with Will when he is available. Leaving them how they are doesn't break anything.
2. I've left the uses of connection.cursor() in https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/dashboard/views.py as they are. We don't enable these view in prod and reading from the replica is fine anyway because we're not reading something we might have just written, so.. meh

chris48s:
Refs https://github.com/DemocracyClub/uk-geo-utils/pull/37

symroe (approved):

PR #8228, opened 2025-02-06
revise cancelled elections template

chris48s:
Refs https://app.asana.com/0/1209332771640377/1209290265508276/f
Refs https://app.asana.com/0/1209332771640377/1209332771640386/f

This is the PR when I am making the most changes.. bit of a general revamp/improvement to the cancelled elections messaging.

I started off thinking the best way to get rid of "cancelled" would be to properly implement cancellations reasons. I've decided that unpicking what is going on with the ballot cache and the 2 different EE backends is just too much to deal with right now. Lets circle back and look at https://app.asana.com/0/1209332771640377/1209332771640406/f after we've re-thought that under the banner of "remove embedded EE". I'm hoping we will simplify a lot of the things that are making this hard in that process :crossed_fingers:

### DIFF
--- a/polling_stations/templates/fragments/cancelled_election.html
+++ b/polling_stations/templates/fragments/cancelled_election.html
@@ -1,32 +1,22 @@
 {% load i18n_with_welsh %}
-{% load markdown_filter %}
 {% if cancelled_election.metadata.cancelled_election.title %}
     <h2>{{ cancelled_election.metadata.cancelled_election.title }}</h2>
 {% else %}
     <h2>{% trans "Cancelled Election" %}</h2>
 {% endif %}
 
-{% if cancelled_election.name %}
-    {% if cancelled_election.rescheduled_date %}
-        <p>{% blocktrans with cancelled_election_name=cancelled_election.name %}The poll for <strong>{{ cancelled_election_name }}</strong> has been rescheduled.{% endblocktrans %}</p>
-        {% comment %}Translators: 'it' is a postponed election{% endcomment %}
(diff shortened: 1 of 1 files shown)


polling_stations/templates/fragments/cancelled_election.html
```diff
@@ -1,32 +1,22 @@
 {% load i18n_with_welsh %}
-{% load markdown_filter %}
```
chris48s, line 2: Uniquely, WDIV was actually set up to expect markdown. I've switched to to use `| linebreaks`

polling_stations/templates/fragments/cancelled_election.html
```diff
@@ -1,32 +1,22 @@
...
 {% if cancelled_election.metadata.cancelled_election.title %}
     <h2>{{ cancelled_election.metadata.cancelled_election.title }}</h2>
 {% else %}
     <h2>{% trans "Cancelled Election" %}</h2>
```
chris48s, line 5: If we are here, we know that the election isn't going to take place and we know it hasn't been rescheduled _yet_ (i.e: there is no replacement ballot attached yet), but we don't actually know if it is postponed or cancelled.

I've changed the inline text to "..will not go ahead as planned".
In terms of the heading here, we could also change it to
"This election will not go ahead as planned"
which is a little wordy but gets rid of cancelled. if you've got any other ideas for generic words, I'm open to ideas.

This will be shown more frequently than the corresponding version on WCIVF

PR #8233, opened 2025-02-10
WIP a DB backend that throws an exception if you try and do anything

chris48s:
This isn't really something to review or merge yet, but I wrote this code and want to shelve it somewhere for now..

This PR adds a DB backend that throws an exception if you try and do anything, which might be useful for unit-testing code that does a lot of DB writes to make sure we're using the right connection. The thinking here is that we can set the default DB connection to `"core.db.error_backend"` as a way of asserting that all our database I/O goes to the primary connection.

The big issue this doesn't help with though is:

```py
def db_for_read(self, model, **hints):
    if model._meta.label == "my_app.Foo":
        return "my_special_db"
    return "default"

...

with transaction.atomic(): # implicitly, this is with transaction.atomic(using="default")
    f = Foo()
    f.save()
```

because although this is definitely **wrong code that looks right**, the empty transaction is a no-op on the default DB so doesn't trigger the exception.

This code might still be useful, but its not a silver bullet for the problem we are trying to solve.

### DIFF
--- a/polling_stations/apps/core/db/error_backend/base.py
+++ b/polling_stations/apps/core/db/error_backend/base.py
@@ -1,0 +1,68 @@
+"""
+A django DB backend that throws an error if you try to do anything. This is
+useful for testing database routing. We can use it if we want to configure a
+DB connection and ensure that it was definitely not used.
+
+Based on
+https://github.com/django/django/tree/main/django/db/backends/dummy
+"""
+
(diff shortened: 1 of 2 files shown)


chris48s:
If we want to go down this route, https://docs.djangoproject.com/en/5.1/topics/db/multi-db/#:~:text=If%20the%20concept,connection.ConnectionDoesNotExist%20exception. might be a better solution

PR #8366, opened 2025-03-05
split EE wrapper in two, update code for querying parquet files

chris48s:
Refs https://app.asana.com/0/1204880927741389/1209455180573677

The core things happening in this PR are:

- Split the old `EveryElectionWrapper` into two classes

  - `EEFetcher` (which requests data from EE)
  - `EEWrapper` (which provides some higher level convenience functions over the data)

  This then allows `EEWrapper` to be constructed with data queried straight from the EE API, or data we got from `LocalParquetElectionsHelper` and follow-up requests to EE. It doesn't care where the data came from and it means we don't need to have 2 slightly different versions of the EE Wrapper class

- Delete code we don't need any more

- Allow querying data from parquet files by UPRN and use that when we're querying by UPRN. The parquet files were previously only being used in quite a small subset of cases.

- Implement a bunch of error handling for invalid conditions that may occur

- Assume we are always dealing with "current" elections (regardless of whether this comes from parquet files or we're querying from EE directly) and filter them down to future only in the wrapper.

- Update all the existing tests

- Add a test suite for `LocalParquetElectionsHelper`, which was previously not under test

- Implement handling for the current spec/assumptions in the parquet files

---

Obviously there is quite a lot going on here and I wanted to keep this changeset relatively constrained. As such, there are also a number of things I explicitly haven't done:

- Improvements to `get_full_ballots()`: See https://github.com/DemocracyClub/UK-Polling-Stations/pull/8366#discussion_r1991541110 for details

- This PR lays down most of the code changes needed to consume these parquet files, but the upshot of merging this PR will be that we are still directly querying EE in production. There are some further jobs needed to actually enable it in production:

  - Set a `ELECTION_PARQUET_DATA_PATH` env var in production
  - Set `USE_LOCAL_PARQUET_ELECTIONS` True in `polling_stations/settings/constants/elections.py`
  - Update `sync_baked_elections_parquet.sh` to get data from the right place (having decided where that is)
  - Ensure dev/staging/prod instances have the necessary permissions to pull data from the relevant S3 bucket
  - Maybe other things? If I have missed anything, let me know

Those two things do need doing, but will happen in follow up PRs.

As long as `USE_LOCAL_PARQUET_ELECTIONS` stays `False`, I think we can just merge this PR without consequence.

---

Then there's a couple of things I don't think we change this side of May:

- There are a number of things I don't really love about the EE Wrapper class:

  - The API surface is really inconsistent. For example, we have a random mash-up of `ee.prop` and `ee.get_prop()` for no real reason
  - Some of the helper methods are implementing logic that may be a bit outdated one way or another and needs revisiting
  - We should probably split the functionality used by the uploader out from the functionality used by the finder - I don't think they really belong in the same class

  I was tempted to try and clean up some of that in this PR while I was refactoring that code, but I decided that just keeping the current API "warts and all" is the MVP here. Then if we want to tackle those issues we can do it in another PR.

- Currently the default is to query EE live. While there are now unit tests covering `LocalParquetElectionsHelper`, all the integration tests that test the application behaviour at a higher level are using (or mocking) `EEFetcher` as the way of querying EE. I haven't really attempted to do anything about this here, but it does mean there is a difference between how we run in production and they system under test (and in dev). I don't think that is something we should try to change right now, but if we decide this is the way forward, lets revisit the way we test and develop locally. For now, lets just be aware of it.


polling_stations/apps/data_finder/helpers/baked_data_helper.py
```diff
@@ -67,56 +50,45 @@ def __init__(self, elections_parquet_path: Path = None, **kwargs):
...
-    def get_full_ballots(self, ballot_data):
+    def get_full_ballots(self, ballot_ids):
         result = []
-        for ballot in ballot_data:
-            ballot_id = ballot["ballot_paper_id"]
+        for ballot_id in ballot_ids:
             url = ballot_paper_id_to_ee_url(ballot_id)
             req = session.get(url)
             result.append(req.json())
         return result
```
chris48s, line 68: There are several problems to note here:

1. There is no error handling at all here. Only happy path handling has been implemented.
2. These requests are being done in sequence, not in parallel. This means if the user has 4 upcoming ballots we make 4 HTTP requests one after another. We're re-using the same connection, but fundamentally it is still sequential.

This does need improving, but I'm going to deal with this topic as a follow up PR.

polling_stations/apps/data_finder/helpers/baked_data_helper.py
```diff
@@ -67,56 +50,45 @@ def __init__(self, elections_parquet_path: Path = None, **kwargs):
...
 
-    def ballot_list_to_dates(self, ballot_list):
```
chris48s, line 93: Removing this keeps the data format the same whether it came from `EEFetcher` or `LocalParquetElectionsHelper`. We are always just dealing with a list of ballot objects

polling_stations/apps/data_finder/helpers/baked_data_helper.py
```diff
@@ -67,56 +50,45 @@ def __init__(self, elections_parquet_path: Path = None, **kwargs):
...
-            # If the file isn't found it should mean that there are no current
-            # elections for this outcode. Just return an empty dates list.
-            return False, []
+            # ERROR
+            # In theory this shouldn't happen
+            # every outcode should exists as a parquet file
+            message = f"Expected file {parquet_filepath} not found"
+            logger.error(message)
+            capture_message(message, level="error")
+            return {"address_picker": False, "ballot_ids": [], "request_success": False}
```
chris48s, line 85: In all these cases where something bad happened:

- Plough on as best we can, but tell `EEWrapper` to follow the `request_success=False` path
- Scream about it in Sentry

symroe (approved):

chris48s:
Quick note here: We talked on a call about slightly simplifying the way we configure this. So something more like setting

```py
BALLOT_FETCHER = EEFetcher
```

or

```
BALLOT_FETCHER = LocalParquetElectionsHelper
```

in settings, and then in code just doing something like

```
ballots, request_success = BALLOT_FETCHER(params)
wrapper = EEWrapper(ballots, request_success)
```

so that we "centralise" more of the conditionality. At the moment we've got several places where this is conditional at the call site. One of the issues with this is that the constructors for `EEFetcher` and `LocalParquetElectionsHelper` don't take the same params at the moment.

The conclusion from that was: Lets roll with this for now and review this once we've trialled the approach a bit more. But just wanted to note it here on the PR so we can refer back to it.

PR #8399, opened 2025-03-10
use coordinate columns from halarose exports

awdem:
Halarose exports seem to have consistent polling station coordinate columns now so we might as well use them if they're populated.

---
  - https://app.asana.com/0/0/1209626808958245

### DIFF
--- a/polling_stations/apps/data_importers/ems_importers.py
+++ b/polling_stations/apps/data_importers/ems_importers.py
@@ -339,8 +339,13 @@ class BaseHalaroseCsvImporter(
 
     def get_station_point(self, record):
         location = None
+        # Try coords first
+        x_coord = float(record.pollingvenueeasting)
+        y_coord = float(record.pollingvenuenorthing)
+        if x_coord > 0 and y_coord > 0:
+            location = Point(x_coord, y_coord, srid=27700)
 
-        # try UPRN first, if available
+        # try UPRN next, if available
         if (
             hasattr(record, self.station_uprn_field)
             and getattr(record, self.station_uprn_field).strip()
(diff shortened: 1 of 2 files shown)


chris48s:
Looks like you need to update some tests to account for this.
Also, can you have a look at
- Chorley
- Watford
- Gravesham
- Lincoln
(those are the 4 areas we have already had data from and written a Halarose import script)
Do they run cleanly? Did any of those 4 areas have co-ordinates in them? If so, is the data good, or are there any issues? Do we need to re-run them?

awdem:
> Looks like you need to update some tests to account for this.
> 
> Also, can you have a look at
> 
> * Chorley
> * Watford
> * Gravesham
> * Lincoln
> 
> (those are the 4 areas we have already had data from and written a Halarose import script)
> 
> Do they run cleanly? Did any of those 4 areas have co-ordinates in them? If so, is the data good, or are there any issues? Do we need to re-run them?

Updated

> Looks like you need to update some tests to account for this.
> 
> Also, can you have a look at
> 
> * Chorley
> * Watford
> * Gravesham
> * Lincoln
> 
> (those are the 4 areas we have already had data from and written a Halarose import script)
> 
> Do they run cleanly? Did any of those 4 areas have co-ordinates in them? If so, is the data good, or are there any issues? Do we need to re-run them?

Update test fixture in 7e82ec7eeeed97c158e01d8405fe064b9fc66bad

I'm going to check those today before merging this.

chris48s:
I'll leave it with you to have a look at those importers next then :+1:

awdem:
> I'll leave it with you to have a look at those importers next then 👍

No changes needed for those 4 so going ahead with the merge!

PR #8487, opened 2025-03-20
log/handle errors 2

chris48s:
Refs https://app.asana.com/0/1204880927741389/1209455180573675/f

This PR is a the follow up to https://github.com/DemocracyClub/UK-Polling-Stations/pull/8366#discussion_r1991541110
and replaces https://github.com/DemocracyClub/UK-Polling-Stations/pull/8468

The core things happening in this PR are:

- If we get an error from EE fetching any ballots handle this cleanly
- Improve logging so we have proper visibility on errors
- If we fail fetching data from EE, throw an error on the API rather than return `ballots: []`

I'll leave some more comments inline

### DIFF
--- a/cdk/lambdas/wdiv-s3-trigger/requirements.txt
+++ b/cdk/lambdas/wdiv-s3-trigger/requirements.txt
@@ -129,9 +129,9 @@ requests-paginator==0.2.0 \
     # via
     #   -c cdk/lambdas/wdiv-s3-trigger/requirements/../../../../requirements/constraints.txt
     #   -r cdk/lambdas/wdiv-s3-trigger/requirements/base.in
-sentry-sdk==2.1.1 \
-    --hash=sha256:95d8c0bb41c8b0bc37ab202c2c4a295bb84398ee05f4cdce55051cd75b926ec1 \
-    --hash=sha256:99aeb78fb76771513bd3b2829d12613130152620768d00cd3e45ac00cb17950f
+sentry-sdk==2.23.1 \
+    --hash=sha256:2288320465065f3f056630ce55936426204f96f63f1208edb79e033ed03774db \
+    --hash=sha256:42ef3a6cc1db3d22cb2ab24163d75b23f291ad9892b1a8c44075ce809a32b191
     # via
     #   -c cdk/lambdas/wdiv-s3-trigger/requirements/../../../../requirements/constraints.txt
     #   -r cdk/lambdas/wdiv-s3-trigger/requirements/base.in
--- a/polling_stations/apps/api/address.py
+++ b/polling_stations/apps/api/address.py
@@ -19,6 +19,7 @@ from drf_spectacular.utils import (
 )
 from pollingstations.models import AdvanceVotingStation
 from rest_framework import serializers
+from rest_framework.exceptions import APIException
 from rest_framework.permissions import IsAuthenticatedOrReadOnly
 from rest_framework.response import Response
 from rest_framework.viewsets import ViewSet
@@ -128,15 +129,18 @@ class AddressViewSet(ViewSet, LogLookUpMixin):
         if getattr(settings, "USE_LOCAL_PARQUET_ELECTIONS", False):
             helper = LocalParquetElectionsHelper()
             resp = helper.get_response(Postcode(address.postcode), address.uprn)
+            if not resp["request_success"]:
+                raise APIException("failed to get list of ballots")
             return EEWrapper(
                 resp["ballots"],
                 request_success=resp["request_success"],
                 include_current=include_current,
             )
 
-        return EEWrapper(
-            **EEFetcher(point=address.location).fetch(), include_current=include_current
-        )
+        ee_response = EEFetcher(point=address.location).fetch()
+        if not ee_response["request_success"]:
+            raise APIException("failed to get list of ballots")
+        return EEWrapper(**ee_response, include_current=include_current)
 
     @extend_schema(
         parameters=[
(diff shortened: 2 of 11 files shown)


cdk/lambdas/wdiv-s3-trigger/requirements.txt
```diff
@@ -129,9 +129,9 @@ requests-paginator==0.2.0 \
     # via
     #   -c cdk/lambdas/wdiv-s3-trigger/requirements/../../../../requirements/constraints.txt
     #   -r cdk/lambdas/wdiv-s3-trigger/requirements/base.in
-sentry-sdk==2.1.1 \
-    --hash=sha256:95d8c0bb41c8b0bc37ab202c2c4a295bb84398ee05f4cdce55051cd75b926ec1 \
-    --hash=sha256:99aeb78fb76771513bd3b2829d12613130152620768d00cd3e45ac00cb17950f
+sentry-sdk==2.23.1 \
+    --hash=sha256:2288320465065f3f056630ce55936426204f96f63f1208edb79e033ed03774db \
+    --hash=sha256:42ef3a6cc1db3d22cb2ab24163d75b23f291ad9892b1a8c44075ce809a32b191
```
chris48s, line 134: I upgraded sentry here because the version we were using didn't match the documentation, which was making things harder.

polling_stations/apps/data_finder/helpers/baked_data_helper.py
```diff
@@ -80,8 +107,9 @@ def get_ballot_list(self, postcode: Postcode, uprn: Optional[str] = None):
             # In theory this shouldn't happen
             # every outcode should exists as a parquet file
             message = f"Expected file {parquet_filepath} not found"
-            logger.error(message)
-            capture_message(message, level="error")
+            scope = get_current_scope()
+            scope.fingerprint = ["parquet:expected_parquet_file_not_found"]
+            logging.error(message)
```
chris48s, line 112: Setting a "fingerprint" for each class of error helps Sentry to group related errors as different "events" for the same "error".
When I hooked this up to a real Sentry DSN, I also realised here that I don't need to be calling both `logging.error` and `capture_message` for these. Sentry's log handler will pick up error logs.

polling_stations/apps/data_finder/helpers/baked_data_helper.py
```diff
@@ -54,17 +55,43 @@ def get_response(self, postcode: Postcode, uprn: Optional[str] = None):
...
 
     def get_full_ballots(self, ballot_ids):
         result = []
         for ballot_id in ballot_ids:
             url = ballot_paper_id_to_ee_url(ballot_id)
-            req = session.get(url)
-            result.append(req.json())
+            response = session.get(url, timeout=10)
+            response.raise_for_status()
+            result.append(response.json())
```
chris48s, line 94: This code implements the revised approach we agreed: If we fail to fetch any ballot, bail.

polling_stations/apps/api/address.py
```diff
@@ -128,15 +129,18 @@ def get_ee_wrapper(self, address, query_params):
         if getattr(settings, "USE_LOCAL_PARQUET_ELECTIONS", False):
             helper = LocalParquetElectionsHelper()
             resp = helper.get_response(Postcode(address.postcode), address.uprn)
+            if not resp["request_success"]:
+                raise APIException("failed to get list of ballots")
```
chris48s, line 133: Inside WDIV (where we only care about polling stations), we have some handling for the `request_success=False` case. We can make some assumptions.
The API can't communicate this though, so instead of saying `ballots: []` we now throw an error, but only in the API. The devs.DC will handle these errors by wrapping them as a `UpstreamApiError` to forward on.
https://github.com/DemocracyClub/aggregator-api/blob/e016298ea2fa0f6186ed00fcc9c2a5482be2ffc1/api/endpoints/v1/voting_information/elections_api_client.py#L46-L47

polling_stations/apps/api/tests/test_address.py
```diff
@@ -202,3 +205,60 @@ def test_bad_slug(self):
         )
 
         self.assertEqual(404, response.status_code)
+
+
+class GetEEWrapperTest(TestCase):
```
chris48s, line 210: Previously, the `get_ee_wrapper` methods of the postcode and API endpoints were not under test at all because we were just mocking out the whole method in the main tests to avoid having to mock individual API calls out to EE. This wasn't a huge issue when this method wasn't doing very much (just constructing a class and returning it), but there is now enough going on in those methods that they need to be under test. These tests are perhaps not perfect, but they do give use a starting point and cover the cases I am introducing here.

symroe:
I think this is looking good, but I'd like to see some sort of retry feature, or at least hear about why we shouldn't implement that now.

Looks like there's a pattern documented for this here: https://docs.python-requests.org/en/latest/user/advanced/#example-automatic-retries

chris48s:
Yes I had forgotten about retries. Good reminder.

If we're doing the requests in sequence rather than parallel and also doing a retry I think the "sad" path could get quite long if we are having problems. I wouldn't want to both have a 10 second timeout on each request and also attempt retries on the timeouts if we're making multiple requests.

I think it probably would make sense to retry if we get a 502 Bad Gateway, but for example if we got a 404 or a 500 from upstream I wouldn't really expect that to change if we try again. Let me have a think about how to approach when we do/don't retry and how long we wait.

symroe:
5 seconds is more than enough I think. Really the main consideration here is network fluctuations rather than EE actually being down :+1:

symroe (approved):

chris48s:
One more tiny change (it is only waffer thin).

Latest commit tells the custom HTTP Transport to also do 1 retry for connection-related errors (so `requests.exceptions.ConnectionError` for example) as well as 502/3/4s

symroe (approved):

PR #8603, opened 2025-04-15
Creates a file that looks like addressbase_cleaned.csv from onspd

GeoWill:
This gets all the postcodes that aren't in addressbase, aren't large users, and, are on mainland GB (i.e. no Northern Ireland, Channel Islands, Isle of Man) and creates a csv that has the same shape as addressbase_cleaned.csv. However there's a dummy uprn in the form:
  ONSPD:TE11ST
Also the address field is blank. This should let us chuck it into the data-baker pipeline to end up in the parquet.

The next step will be to generate the csv and add it to the s3 bucket that databaker produces the parquet files from.

### DIFF
--- a/polling_stations/apps/addressbase/management/commands/create_dummy_cleaned_addresses_from_onspd.py
+++ b/polling_stations/apps/addressbase/management/commands/create_dummy_cleaned_addresses_from_onspd.py
@@ -1,0 +1,97 @@
+import csv
+from pathlib import Path
+
+from django.core.management.base import BaseCommand
+from django.db.models import F, Value, CharField, OuterRef, Subquery, Func
+from django.db.models.functions import Concat, Replace
+
+from addressbase.models import Address
+from uk_geo_utils.models import Onspd
+
+
+class AsEWKT(Func):
+    """Calls the postgis ST_AsEWKT function to get strings like: SRID=4326;POINT(0.088849 51.43053)"""
+
+    function = "ST_AsEWKT"
+    output_field = CharField()
+
+
+class Command(BaseCommand):
+    help = (
+        "Export postcodes that exist in ONSPD but not in AddressBase to a CSV file."
+        "It's a good idea to name the output file according to the version of onspd it's derived from."
+        "eg: python manage.py create_dummy_cleaned_addresses_from_onspd ONSPD_AUG_2024_addressbase_cleaned.csv"
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument("output_file", type=str, help="Path to output CSV file")
+
+    def handle(self, *args, **options):
+        output_file = options["output_file"]
+
+        output_path = Path(output_file)
+        output_path.parent.mkdir(parents=True, exist_ok=True)
+
+        # First, create a queryset for postcodes that exist in the Address model
+        existing_postcodes = Address.objects.filter(postcode=OuterRef("pcds")).values(
+            "postcode"
+        )
+
+        # Main query
+        results = (
+            Onspd.objects.annotate(
+                # Create "ONSPD:{postcode}" with whitespace removed
+                uprn=Concat(
+                    Value("ONSPD-", output_field=CharField()),
+                    Replace("pcds", Value(" "), Value(""), output_field=CharField()),
+                    output_field=CharField(),
+                ),
+                address=Value("", output_field=CharField()),
+                postcode=F("pcds"),
+                location_wkt=AsEWKT("location"),
+                address_type=Value("", output_field=CharField()),
+            )
+            .filter(
+                doterm="",  # Active postcodes (not terminated)
+                usertype="0",  # Standard geographic postcodes
+            )
+            .exclude(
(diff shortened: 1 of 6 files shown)


polling_stations/apps/addressbase/management/commands/create_dummy_cleaned_addresses_from_onspd.py
```diff
@@ -0,0 +1,97 @@
...
+                # Create "ONSPD:{postcode}" with whitespace removed
+                uprn=Concat(
+                    Value("ONSPD-", output_field=CharField()),
+                    Replace("pcds", Value(" "), Value(""), output_field=CharField()),
+                    output_field=CharField(),
+                ),
+                address=Value("", output_field=CharField()),
+                postcode=F("pcds"),
+                location_wkt=AsEWKT("location"),
+                address_type=Value("D", output_field=CharField()),
```
chris48s, line 52: Do these actually correspond to Type D addresses?
My instinct is it would be better to use something like `Z` here that definitely isn't a valid value in the same way that we decided to make "uprn" obviously not a UPRN

GeoWill: Yeah, I guess it probably should be something different. I'm not sure it matters in practice, because the value is only there to make it correspond with the glue table schema. I'm sort of tempted to make it blank, similar to the address value. Or maybe set both to `N/A`?

chris48s:
TODO: bring ONSPD Geocoder behaviour into line with this so all postcodes that we consider "valid" in WDIV should also appear in the files

polling_stations/apps/data_finder/helpers/geocoders.py
```diff
@@ -46,7 +46,6 @@ def geocode(self):
         ]
         usertype = geocoder.get_code("usertype")
         exclude_prefixes = [
-            "BT",  # Northern Ireland
```
GeoWill: At the moment the `AddressbaseGeocoder` in `uk-geo-utils` throws a `NorthernIrelandException` for `BT` postcodes ([link](https://github.com/DemocracyClub/uk-geo-utils/blob/8c8d5e8cfce5e37e8bf5d9ab4618df87e639950a/uk_geo_utils/geocoders.py#L64)). 

This means we do need ONSPD to geocode NI postcodes for now.
Given we're moving to always having an approved set of NI addresses in the DB and in the parquet files I think we should change this in `uk-geo-utils`

polling_stations/templates/fragments/errors.html
```diff
@@ -3,7 +3,7 @@
 
 <div class="ds-text-centered">
     <h2>{% blocktrans %}Sorry, we can't find {{ postcode }}{% endblocktrans %}</h2>
-    <p>{% trans "This doesn't appear to be a valid postcode." %}</p>
+    <p>{% trans "We don't hold data for this postcode, or it is an invalid postcode." %}</p>
```
GeoWill, line 6: I haven't remade the translation files, because when I ran `makemessages` there were loads of things that had changed. I think it's better that we just deal with it all together after May 1st.

polling_stations/apps/data_finder/helpers/geocoders.py
```diff
@@ -39,12 +39,22 @@ def geocode(self):
...
+        exclude_prefixes = [
+            "GIR",  # Banks
+        ]
+        if (
+            not local_auth
+            or local_auth in lad_error_values
+            or usertype == "1"  # large users
+            or self.postcode[:2]
+            in exclude_prefixes  # things we don't want to geocode from onspd
+        ):
```
chris48s, line 57: `self.postcode[:2]`
is the first 2 characters
but `GIR` is 3 characters
Given the Girobank postcode is a large user anyway, I think we could just completely ditch this specific check.

polling_stations/templates/fragments/errors.html
chris48s: Yep. Agreed.
I was planning on getting https://app.asana.com/1/1204880536137786/project/1205090401342417/task/1208714863410843?focus=true done before this election cycle tbh, but it never fitted in.

chris48s: Can we say "we don't hold information..." here rather than "data". It feels a bit less techy

polling_stations/apps/data_finder/helpers/geocoders.py
chris48s: If you view uk-geo-utils as a generic python library that helps people work with addressbase, I think the current behaviour is right and we should override/customise it in WDIV because we're doing wierd stuff in this application (we're adding records that are unambiguously not in addressbase to the address table).
If you view uk-geo-utils as a bit of WDIV that unhelpfully lives in another repo, then yeah.
I think I'm pretty much of the opinion that uk-geo-utils is mostly not a useful abstraction at this point, but lets come back to that..

I think for now we can just take no action here for the moment and nothing gets more broken, right.

chris48s (approved):

PR #8605, opened 2025-04-16
councils app: do all the reads/writes on the right DBs

chris48s:
Refs https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1209717073277337?focus=true

The problems here turned out to be a bit of a mixed bag.
Some of it is stuff I messed up in https://github.com/DemocracyClub/UK-Polling-Stations/pull/8224
Some of it is code that was never doing quite what we thought it was

One of the key things going on here is that there's a number of places where we wrote something like.

```py
thing = Thing.objects.using(db).get_or_create(this='that')
```

From reading the code, I think our assumption/hope was that this would do something like:

```py
try:
    thing = Thing.objects.using(db).get(this='that')
except Thing.DoesNotExist:
    thing = Thing(this='that')
    thing.save(using=db)
```

but it actually behaves more like:

```py
try:
    thing = Thing.objects.using(db).get(this='that')
except Thing.DoesNotExist:
    thing = Thing(this='that')
    thing.save(using=None) # crucial difference!!
```

### DIFF
--- a/deploy/files/conf/crontab
+++ b/deploy/files/conf/crontab
@@ -2,8 +2,6 @@
 # HOWEVER: these commands will NOT run on instances booted from the EE AMI
 # If you want to have a command run only once per env / once globally, use SSM Run Command
 MAILTO=developers@democracyclub.org.uk
-# Import councils from EC hourly
-1 * * * * polling_stations /usr/bin/manage-py-command import_councils --only-contact-details --database default
 # Sync elections from EE
 */5 * * * * every_election /usr/local/bin/ee-manage-py-command sync_elections
 # Per instance cloudwatch custom metrics
--- a/polling_stations/apps/councils/management/commands/import_councils.py
+++ b/polling_stations/apps/councils/management/commands/import_councils.py
@@ -10,7 +10,6 @@ from django.core.management.base import BaseCommand
 from django.db import DEFAULT_DB_ALIAS, transaction
 from requests.exceptions import HTTPError
 from retry import retry
-from polling_stations.db_routers import get_principal_db_name
 
 
 from polling_stations.settings.constants.councils import (
@@ -18,8 +17,6 @@ from polling_stations.settings.constants.councils import (
     NIR_IDS,
     WELSH_COUNCIL_NAMES,
 )
-
-DB_NAME = get_principal_db_name()
 
 
 def union_areas(a1, a2):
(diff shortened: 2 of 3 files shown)


polling_stations/apps/councils/management/commands/import_councils.py
```diff
@@ -276,40 +279,43 @@ def handle(self, **options):
...
+                        id_field=OLD_ID_FIELD,
+                        include=OLD_BOUNDARIES_GSS_CODES,
+                    )
+
+            # Clean up old councils that we've not seen in the EC data
+            Council.objects.using(self.database).exclude(
+                council_id__in=self.seen_ids
+            ).delete()
+
+            self.stdout.write("..done")
```
chris48s, line 321: It looks like I changed a lot here because I've intented this whole function in one tab.
What I've actually done here is moved the `transaction.atomic` from being a decorator round this function to a `with` block so that I can reference `this.database`. One of the problems here was that the transaction was always on the principal DB regardless of the value passed in `--database`
You can look at the diff ignoring whitespace at https://github.com/DemocracyClub/UK-Polling-Stations/pull/8605/files?w=1 which is a bit more managable

deploy/files/conf/crontab
```diff
@@ -2,8 +2,6 @@
 # HOWEVER: these commands will NOT run on instances booted from the EE AMI
 # If you want to have a command run only once per env / once globally, use SSM Run Command
 MAILTO=developers@democracyclub.org.uk
-# Import councils from EC hourly
-1 * * * * polling_stations /usr/bin/manage-py-command import_councils --only-contact-details --database default
```
chris48s, line 6: As discussed, I've just disabled this job that is running against the replicas.
We've got a scheduled job to do this on the primary
https://github.com/DemocracyClub/UK-Polling-Stations/blob/3a0bb9dff904bf79030df4492f2148e04ef75a65/cdk/stacks/command_runner_stack.py#L50-L54
and we replicate from there

chris48s:
As future work, I think I'm just going to bin the `--database` flag on this command: https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1210000883029015?focus=true

polling_stations/apps/councils/management/commands/import_councils.py
GeoWill: TiL `?w=1`. It also works on the commit diff view.

Issue #7317, opened 2024-05-31
Make it clear that the example page is an example page


pmk01:
https://wheredoivote.co.uk/example/

A recent feedback on this page: "I live in Edinburgh, Scotland central belt, i find it somewhat unlikely that my polling station is in Bristol - misinformation like this is unhelpful"

Maybe we should add (this is an example page).

pmk01:
https://wheredoivote.co.uk/admin/bug_reports/bugreport/756/change/
https://wheredoivote.co.uk/admin/bug_reports/bugreport/626/change/

PR #8623, opened 2025-04-29
mark example page as example

awdem:
Closes https://github.com/DemocracyClub/UK-Polling-Stations/issues/7317

This PR adds a message to the example page indicating its an example and directing users to the home page for the real postcode lookup.

image

### DIFF
--- a/polling_stations/templates/robots.txt
+++ b/polling_stations/templates/robots.txt
@@ -4,6 +4,7 @@ Disallow: /address/
 Disallow: /address_select/
 Disallow: /email/
 Disallow: /embed/
+Disallow: /example/
 Disallow: /i18n/
 Disallow: /league_table/
 Disallow: /nus_wales/


awdem:
@pmk01 Thoughts on wording?

pmk01:
> @pmk01 Thoughts on wording?

Wording looks good to me, thanks!

awdem:
I've also just added some commit to fix the wonky error messaging

chris48s:
I've pushed another commit to your branch adding a Disallow rule for /example to robots.txt

On the error handling, it feels like we're making more of a meal of this than we need to. This snippet is used in 2 places. If it needs to behave differently in those two places, its probably just 2 different bits of code not a shared snippet. But what's here is an improvement on what is deployed to production. Lets merge this now and get the site looking right since lots of eyes are on the site right now. I'll come back to it tomorrow.

chris48s (approved):

PR #8642, opened 2025-05-28
Walsall polling districts review

chris48s:
This PR implements experimental changes to allow us to show a custom search/results page for Walsall Council's upcoming review of polling stations

I've tried to minimise the changes to the core WDIV app to enable this, but inevitable I have had to make some. I've split things up so the changes to the core are in one commit and the Walsall specific stuff is separated.

### DIFF
--- a/polling_stations/apps/data_finder/review_views.py
+++ b/polling_stations/apps/data_finder/review_views.py
@@ -1,0 +1,67 @@
+import data_finder.views as data_finder_views
+from django.shortcuts import get_object_or_404
+from addressbase.models import Address
+from django.urls import reverse
+from django.http import HttpResponseRedirect
+
+
+class HomeView(data_finder_views.HomeView):
+    template_name = "reviews/WLL/20250609/home.html"
+    namespace = "reviews:"
+
+
+class PostcodeView(data_finder_views.PostcodeView):
+    template_name = "reviews/WLL/20250609/postcode_view.html"
+    namespace = "reviews:"
+
+    def get(self, request, *args, **kwargs):
+        resp = super().get(request, *args, **kwargs)
+        council = getattr(self, "council", None)
+        if council and council.council_id != "WLL":
+            return HttpResponseRedirect(reverse("postcode_view", kwargs=kwargs))
+        return resp
+
+    def we_know_where_you_should_vote(self):
+        return self.get_station()
+
+    def get_ee_wrapper(self, *args, **kwargs):
+        ee = super().get_ee_wrapper(*args, **kwargs)
+        ee.has_election = lambda: True
+        return ee
+
+
+class AddressView(data_finder_views.AddressView):
+    template_name = "reviews/WLL/20250609/postcode_view.html"
+    namespace = "reviews:"
+
+    def get(self, request, *args, **kwargs):
+        address = get_object_or_404(
+            Address.objects.select_related("uprntocouncil"), uprn=self.kwargs["uprn"]
+        )
+        if address.uprntocouncil.lad != "E08000030":
+            return HttpResponseRedirect(reverse("address_view", kwargs=kwargs))
+        return super().get(request, *args, **kwargs)
+
+    def we_know_where_you_should_vote(self):
+        return self.get_station()
+
+    def get_ee_wrapper(self, *args, **kwargs):
+        ee = super().get_ee_wrapper(*args, **kwargs)
+        ee.has_election = lambda: True
+        return ee
+
+
+class WeDontKnowView(data_finder_views.WeDontKnowView):
+    template_name = "reviews/WLL/20250609/postcode_view.html"
+    namespace = "reviews:"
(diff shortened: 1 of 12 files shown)


polling_stations/urls.py
```diff
@@ -79,6 +84,38 @@
     ),
 ]
 
+boundary_review_patterns = (
+    [
+        re_path(
+            r"^WLL/20250609/postcode/(?P<postcode>.+)/$",
```
chris48s, line 90: At the moment everything is hard-coded to Walsall, but I've thought ahead a little about how we might generalise this if we were to take this idea further and made it so we could generalise the URLs and template locations in future if this has legs.

polling_stations/apps/data_finder/review_views.py
chris48s, line 1: I ended up writing a bit more code than I would have ideally liked in this file, but the basic point here is:

Override the core views with the following changes:

- Namespace is `reviews:` so any routing we do moves you round within the sub-app
- Bypass the has_election/station_valid checks - always show a result if we have data
- If someone enters a result in not-Walsall, redirect to the "main" journey
- Override some of the templates with custom ones

If we wanted to generalise this to handle other councils, we could. Also there are a few things we might clean up.

polling_stations/templates/reviews/WLL/20250609/home.html
```diff
@@ -0,0 +1,21 @@
...
+        review and is asking for your views on the draft proposals.
+        Please provide your feedback by no later than Sunday 20 July 2025.
+      </p>
+      <p>
+        Use the tool below to find your proposed polling station and complete
+        the survey on Walsall Council's website
+        <a href="https://www.walsall.gov.uk/PollingDistrictReview">
+          https://www.walsall.gov.uk/PollingDistrictReview
+        </a>
+      </p>
```
chris48s, line 18: This text was supplied by the council

polling_stations/templates/reviews/WLL/20250609/postcode_view.html
```diff
@@ -0,0 +1,62 @@
...
+                        </p>
+                        <p>
+                            Please provide your feedback by no later than Sunday 20 July 2025.
+                        </p>
+                        <p>
+                            For more information please visit
+                            <a href="https://www.walsall.gov.uk/PollingDistrictReview">
+                                https://www.walsall.gov.uk/PollingDistrictReview
+                            </a>
+                        </p>
```
chris48s, line 43: This text was supplied by the council

polling_stations/apps/data_importers/management/commands/import_walsall.py
```diff
@@ -1,44 +1,70 @@
...
+    elections = []
+
+    def district_record_to_dict(self, record):
+        if record[1] == "PEL4AB":
+            return {"internal_council_id": "PEL4/AB", "name": "PEL4/AB"}
+        if record[1] == "ASC5/AB":
+            return {"internal_council_id": "ACS5/AB", "name": "ACS5/AB"}
+        return {"internal_council_id": record[1], "name": record[1]}
+
+    def station_record_to_dict(self, record):
```
chris48s, line 18: A couple of things to note here.

1. There are 3 districts where they want us to show 2 different options for the station location
2. They wanted us to show some extra text

The way I've chosen to do this is by jamming all of it in the address field as markdown and then rendering the markdown in the template.
Obviously, this is not good code. Lets not do more of this.
That said, given this is something we might do once and abandon, I decided I didn't want to add more DB fields and make changes to the base importer classes to accommodate this.
If we decide to develop this feature further this is one of the main things I would want to change.
This mess won't be exposed via the main frontend or API because has_election will always be false, but I will tear down this data as soon as their consultation is over.

GeoWill:
In general looking good. I want to have another read over, but I'm stopping for the day soon, and wanted to flag this. Where the `Ward_` field on the stations contains a line break, it messes up the markdown rendering. eg `WS3 1BU`
image

polling_stations/apps/data_finder/review_views.py
GeoWill: I'm not sure what the right thing to do when someone enters looks for a not-Walsall address/postcode. Just playing with it, it's a bit confusing to go from the 'Walsall Polling Districts Review' page, to being told 'We don't know about any upcoming elections in your area'. 
My instinct is towards having a 'WeDontKnowView' in this file, which says "you're not in Walsall. If you're looking for a polling station to vote in an upcoming election click 'here'".  
What do you think?

GeoWill (commented):
Think this all looks sensible. I've put a couple comments/questions inline.

chris48s:
Thanks. Pushed a couple of extra commits. I think I am going to roll with this to start with and think about the various flavours of "we don't know" as a follow-up.

GeoWill (approved):
Looks good. Fine to follow up on the we don't knows later.

PR #8650, opened 2025-06-04
remove most usage of NEXT_CHARISMATIC_ELECTION_DATE(S)

chris48s:
Refs https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1210123314406393?focus=true

First off, a statement of the core problem this PR solves:

If we set one or more `NEXT_CHARISMATIC_ELECTION_DATES` and we also have polling station data imported for other elections happening after that date, we will say "We don't know of any upcoming elections in your area" for areas with "non-charismatic" elections instead of showing a result.
To reproduce this state locally with a checkout from master:

Import data for upcoming `local.nottinghamshire.2025-06-12` and `local.greenwich.2025-06-26`:
- `./manage.py import_mansfield`
- `./manage.py import_greenwich`

Now you will find that if you set `NEXT_CHARISMATIC_ELECTION_DATES = []` searching for SE18 3JL will serve a result (correct/expected) but if you set `NEXT_CHARISMATIC_ELECTION_DATES = ["2025-06-12"]` it will tell you "We don't know of any upcoming elections in your area" (wrong/unexpected).

To fix this, I have mostly removed the use of `NEXT_CHARISMATIC_ELECTION_DATE` and replaced it with `next_election_date` from EE

There are two places where I've kept `NEXT_CHARISMATIC_ELECTION_DATE`.

1. The "polling day is today" card we sometimes show on the homepage. In this case the user hasn't searched for anything yet so we have to rely on a hard-coded value. We can't rely on getting a value from EE.
2. The advance voting stations pilot feature. This is not currently in use and I've slated it as something to possibly remove in https://app.asana.com/1/1204880536137786/project/1205090401342417/task/1210055377993847?focus=true so let's review that then

### DIFF
--- a/polling_stations/apps/data_finder/helpers/every_election.py
+++ b/polling_stations/apps/data_finder/helpers/every_election.py
@@ -117,6 +117,10 @@ class BaseEEWrapper(metaclass=abc.ABCMeta):
         pass
 
     @abc.abstractmethod
+    def get_next_election_date(self) -> Optional[str]:
+        pass
+
+    @abc.abstractmethod
     def get_ballots_for_next_date(self) -> List:
         pass
 
@@ -152,6 +156,9 @@ class EmptyEEWrapper(BaseEEWrapper):
         return False
 
     def get_metadata(self) -> None:
+        return None
+
+    def get_next_election_date(self) -> Optional[str]:
         return None
 
     def get_ballots_for_next_date(self) -> List:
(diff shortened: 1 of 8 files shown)


GeoWill (approved):
This looks good. I haven't done the work of setting up City of London elections locally to check that the logic works, but I have had a good read, and I think we're doing the right thing. It also all behaves when City of London isn't involved.

PR #8651, opened 2025-06-05
improve performance when deleting lots of councils

chris48s:
Fundamentally, this moves us from doing

```py
for council in councils:
    Thing.objects.filter(council=council).delete()
    ...
```

which is slow with lots of councils
to doing

```py
council_ids = [c.id for c in councils]
Thing.objects.filter(council_id__in=council_ids).delete()
...
```

which is much faster.

I'd like to try this again on production to see how it performs with real-life data volume/DB latency before I go ahead and schedule it, but my instinct is this should get the job done well enough without us having to spend a bunch of time on optimising it more :crossed_fingers:

I also tidied up a couple of other bits while I was here.


GeoWill (approved):
:+1:

PR #8670, opened 2025-06-26
Add termination hook to deployment group

GeoWill:
### What this pr is doing

Setting up a cron job to drop failed replication slots and report to slack. This will hopefully give us some visibility on the issue, and let us debug it better. It will also stop the stored wal files getting too big and causing problems. 

-----------------
Below this line is not so relevant for the PR as it currently stands.

------------------

### What's the problem?

When we do more deploys/deploys involving more instances we end up with inactive replication slots on RDS.
These can be seen by running:

```sql
=> select * from pg_replication_slots where active = 'f';
```

### What's been done here?

Add an `AfterBlockTraffic` hook, and make sure it runs when we drop the instance count in the ASG.

We're running codedeploy blue/green deployments. The order of the hooks in a deployment are described
[here](https://docs.aws.amazon.com/codedeploy/latest/userguide/reference-appspec-file-structure-hooks.html#reference-appspec-file-structure-hooks-run-order) (scroll down to `Blue/green deployments` section).

This makes it seem like the `AfterBlockTraffic` hook would be the perfect place to run:

```
systemctl stop polling_stations_db_replication.service
```

This should call the `ExecStop` section of `deploy/files/systemd/db_replication.service`. Which in turn runs the script at `deploy/files/scripts/remove_db_replication.sh`.
Which in turn drops the subscription which should tell the publisher to get rid of any replication slots.

There are two kinds of deployments which happen.

1. User action
These are normally the result of a new commit on `master`. As part of CI the `deploy/create_deployment.py` script is called. This kicks of a new deploy.

2. Auto Scaling group action
This results from bumping the desired instance count on the ASG. This might be done in the console, or in code, which will then be implemented by the `deploy/update_autoscaling_group.py`

My reading of the docs is that the `AfterBlockTraffic` hook will be picked up by the instances terminated as part of User Action deploys, but maybe not in the case of AutoScaling deploys.
However AWS seems to have a mechanism for this, 'termination deploys' [docs](https://docs.aws.amazon.com/codedeploy/latest/userguide/integrations-aws-auto-scaling.html#integrations-aws-auto-scaling-behaviors-hook-enable).
I've enabled these with `terminationHookEnabled=True,` in the `deploy/create_deployment_group.py` script. That's really just record keeping though, because the deployment group is only created once, (and so that script is normally a no-op). So I actually just went in and did it in the 

In summary there are two things this PR does. First adding in an `AfterBlockTraffic` hook to `appspec.yml`. Second making sure this runs on AutoScaling scale out (dropping the instance number) events.

The second part works (i.e. we get termination deploys when dropping the ASG desired count). Sadly afaict the `AfterBlockTraffic` hook doesn't run in either case. My way of verifying this is to look at the syslog in cloudwatch. Also when playing with cranking the ASG desired count up and down I got a few failed slots.

#### Gotcha when working with  the `AfterBlockTraffic` hook

The hook needs to be on the instance where you want it to run.
So if you update `appspec.yml` and do a deploy, the outgoing instances in that deploy won't have your latest changes on them, so they won't do the thing you want.
You need to do _another_ deploy, so that the instances being removed are the ones with your new `appspec.yml`. This is obviously boring.

### Next steps.

The simplest option is probably still to just write a cron job that does:

 ```sql
 select pg_drop_replication_slot(slot_name) from pg_replication_slots where active = 'f';
```

Otherwise I think creating an `AfterBlockTraffic` hook that explicitly logs something, and trying to get that to run is the next step. the [docs](https://docs.aws.amazon.com/codedeploy/latest/userguide/reference-appspec-file-structure-hooks.html) say  `AfterBlockTraffic – You can use this deployment lifecycle event to run tasks on instances after they are deregistered from their respective load balancer.` I haven't looked at what's going on with the load balancer during a deployment, but that's maybe something to check.

### DIFF
--- a/polling_stations/management/commands/drop_failed_replication_slots.py
+++ b/polling_stations/management/commands/drop_failed_replication_slots.py
@@ -1,0 +1,149 @@
+from datetime import date
+from typing import Dict, Any
+
+from django.core.management.base import BaseCommand
+from django.conf import settings
+
+from core.slack_client import SlackClient
+from settings.constants.slack import BOTS_CHANNEL, BOTS_TESTING_CHANNEL
+from polling_stations.db_routers import (
+    get_principal_db_name,
+    get_principal_db_connection,
+)
+
+DB_NAME = get_principal_db_name()
+
+
+class Command(BaseCommand):
+    help = "Checks for failed replication slots in RDS and drops them, then reports to Slack"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.post_to_slack = False
+        self.dc_environment = None
+        self.cursor = get_principal_db_connection().cursor()
+        self.slack_client = SlackClient(channel=self.slack_channel)
+        self.failed_slots = []
+        self.drop_succeeded = []
+        self.drop_failed = []
+
+    def add_arguments(self, parser):
+        super().add_arguments(parser)
+        parser.add_argument(
+            "--send-slack-report",
+            action="store_true",
+            help="Ask import_eoni command to report to slack",
+        )
+
+    def handle(self, *args, **options):
+        try:
+            self.dc_environment = settings.DC_ENVIRONMENT
+            self.post_to_slack = options.get("post_to_slack")
+            self.check_replication_slots()
+
+            if self.failed_slots:
+                self.drop_failed_slots()
+                self.report_to_slack()
+                return
+
+            if date.today().weekday() == 0:
+                # Report once a week even if no failed slots found, so we know bot is working.
+                self.report_to_slack()
+                return
+
+        except Exception as e:
+            self.send_failure_to_slack(e)
+            raise e
+
+    @property
+    def slack_channel(self):
+        if self.dc_environment == "production":
+            return BOTS_CHANNEL
+        else:
+            return BOTS_TESTING_CHANNEL
+
+    def check_replication_slots(self):
+        with get_principal_db_connection().cursor() as cursor:
+            query = """
+                    SELECT
+                        slot_name
+                    FROM pg_replication_slots
+                    WHERE
+                       active = 'f'
+                       AND database = %s
+                    ORDER BY slot_name;
+                    """
+
+            cursor.execute(query, [DB_NAME])
+
+            for row in cursor.fetchall():
+                self.failed_slots.append(row[0])
+
+    def report_to_slack(self):
+        if not self.post_to_slack:
+            return
+        try:
+            response = self.post_thread_header()
+            thread_ts = response.get("ts")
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: polling_stations/management/commands/drop_failed_replication_slots.py [hunk 1]
```diff
@@ -0,0 +1,151 @@
...
+                self.drop_failed_slots()
+                self.report_to_slack()
+                return
+
+            if date.today().weekday() == 0:
+                # Report once a week even if no failed slots found, so we know bot is working.
+                self.report_to_slack()
+                return
+
+        except (KeyboardInterrupt, SystemExit):
```
symroe, line 54: This is redundant as both inherit from `BaseException` not `Exception`. See https://docs.python.org/3/library/exceptions.html#exception-hierarchy

polling_stations/management/commands/drop_failed_replication_slots.py
```diff
@@ -0,0 +1,151 @@
...
+            return
+        try:
+            response = self.post_thread_header()
+            thread_ts = response.get("ts")
+            if self.failed_slots:
+                self.post_failed_details(thread_ts)
+            else:
+                self.post_check_running(thread_ts)
+        except Exception as e:
+            self.stderr.write(self.style.ERROR(f"Error posting to Slack: {str(e)}"))
```
symroe, line 95: Is it worth posting a Sentry exception here?

polling_stations/management/commands/drop_failed_replication_slots.py
```diff
@@ -0,0 +1,151 @@
...
+        if not self.post_to_slack:
+            return
+        try:
+            response = self.slack_client.send_message(
+                message=":warning: *drop_failed_replication_slots command failed*",
+            )
+            self.slack_client.send_message(
+                message=f"Error: {str(exception)}", thread_ts=response.get("ts")
+            )
+        except (KeyboardInterrupt, SystemExit):
```
symroe, line 147: Same as above

polling_stations/management/commands/drop_failed_replication_slots.py
GeoWill: Fixed in 0067957b2. I cribbed `from polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py` - but some archaeology suggests that got [added in unnecessarily](https://github.com/DemocracyClub/UK-Polling-Stations/pull/8192#discussion_r1888669924).

polling_stations/management/commands/drop_failed_replication_slots.py
GeoWill: 3b19c910b

symroe (approved):

PR #8673, opened 2025-07-15
Make sure we rebuild wdiv-s3-trigger/requirements.txt before deploy

GeoWill:
I think this is the issue. I.e. the lambda was getting bundled up with an empty `requirements.txt`. This means `wdiv-s3-trigger/requirements.txt` should be populated before we do the deploy.

### DIFF
--- a/.circleci/deploy-config.yml
+++ b/.circleci/deploy-config.yml
@@ -176,7 +176,10 @@ jobs:
           uv run --group cdk npx cdk synth --all
     - persist_to_workspace:
         root: ~/repo/
-        paths: [ cdk.out ]
+        paths: [
+          cdk.out,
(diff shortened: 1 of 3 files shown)


chris48s:
Ah. Good catch. Instead of re-generating this each time we need it, I think the way we normally do this is by building it once and using `persist_to_workspace` so it is available at later build steps. So for example:

https://github.com/DemocracyClub/aggregator-api/blob/b22232b2cbb31c094d1ca407b8ef6a6a69ec6708/.circleci/config.yml#L25-L31

https://github.com/DemocracyClub/Website/blob/e1ed1f3d74ade47d3557ac3fdd62726ccfbc4304/.circleci/config.yml#L24-L30

chris48s (approved):

PR #8678, opened 2025-07-18
Remove embedded ee

GeoWill:


chris48s:
Looks like this covers pretty much everything.
I think the one extra thing we might do here is remove the ability to set `EE_BASE` with an env var as this really is a constant now, not something we'd expect to fiddle with throughout the year

GeoWill:
I think it could be useful if we were doing something where we want to point dev-WDIV at dev-EE.

chris48s (approved):

PR #8700, opened 2025-08-22
Bump to UV 0.8.*

symroe:


chris48s:
This is another project where we were caching the virtualenv, and then relying on the venv being cached in later build steps.
Do you want to switch to not caching the virtualenv between steps, and then rebuilding it in later build steps here? I've kind of lost track of the reason why it was desirable to stop caching the venv..

PR #8706, opened 2025-09-03
respect noindex flag

chris48s:
This is an example of some wrong code that looked right :facepalm: 

This was actually rendering `<meta name="robots" content="noindex">` on _every_ page
We totally disappeared from the internet

Whoops

### DIFF
--- a/polling_stations/templates/base.html
+++ b/polling_stations/templates/base.html
@@ -12,11 +12,11 @@
     {%  endblock extra_page_css %}
 {% endblock extra_site_css %}
 
-{% if noindex %}
-    {% block page_meta %}
+{% block page_meta %}
+    {% if noindex %}
         <meta name="robots" content="noindex">
-    {% endblock page_meta %}
-{% endif %}
+    {% endif %}
+{% endblock page_meta %}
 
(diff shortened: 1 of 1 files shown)


GeoWill (approved):
Well spotted. Checked it's all working locally.
I don't understand why this fix works though - i.e. why `{% if noindex %}` evaluates differently depending on whether it's inside the `page_meta` block or outside of it :confused: .  

I guess does `block page_meta` get evaluated independently of the `if` because it's inherited from a parent template?

chris48s:
Block tags evaluate first then everything else. It is documented behaviour

https://docs.djangoproject.com/en/4.2/ref/templates/language/#:~:text=%7B%25%20block%20%25%7D%20tags%20are%20evaluated%20first.%20That%E2%80%99s%20why%20the%20content%20of%20a%20block%20is%20always%20overridden%2C%20regardless%20of%20the%20truthiness%20of%20surrounding%20tags.%20For%20example%2C%20this%20template%20will%20always%20override%20the%20content%20of%20the%20title%20block%3A

just not totally intuitive

PR #8740, opened 2025-11-06
Fix syntax for teardown_expired_data job command

chris48s:


awdem:
Will check on Monday if this works.

PR #8742, opened 2025-11-12
upgrade urllib==2.5.0

chris48s:
I rebuilt the yaml files using the method in
https://github.com/kevin1024/vcrpy/issues/719#issuecomment-1811544263
See that issues for more detail on why this is needed


chris48s:
The failing test on this one is
https://github.com/DemocracyClub/UK-Polling-Stations/pull/8741
block on that

PR #8746, opened 2025-11-18
expose AccessibilityInformation object in the API

chris48s:
This PR exposes `AccessibilityInformation` DB objects as a property of the polling station if we hold it. If we don't hold it, the key exists but is `null`.

WDIV's API is private so I haven't bothered putting this behind a flag here. We will deal with that on the devs.DC side.

That said, I suspect we might just pass through the `polling_station` property over there, so I might need to write the PR over there that excludes this by default before we go ahead and merge this.

### DIFF
--- a/polling_stations/apps/api/pollingstations.py
+++ b/polling_stations/apps/api/pollingstations.py
@@ -7,6 +7,7 @@ from rest_framework.reverse import reverse
 from rest_framework.serializers import (
     CharField,
     HyperlinkedModelSerializer,
+    ModelSerializer,
     SerializerMethodField,
 )
 from rest_framework.viewsets import GenericViewSet
@@ -14,9 +15,6 @@ from rest_framework_gis.serializers import GeoFeatureModelSerializer
 
 from .mixins import PollingEntityMixin
 from pollingstations.models import AccessibilityInformation
-from rest_framework.serializers import (
-    ModelSerializer,
-)
 
 
 class AccessibilityInformationSerializer(ModelSerializer):


polling_stations/apps/api/pollingstations.py
```diff
@@ -13,6 +13,28 @@
...
+
+class AccessibilityInformationSerializer(ModelSerializer):
+    class Meta:
+        model = AccessibilityInformation
+        fields = (
+            "is_temporary",
+            "nearby_parking",
+            "disabled_parking",
+            "level_access",
+            "temporary_ramp",
```
GeoWill: There's a [bit of logic](https://github.com/DemocracyClub/UK-Polling-Stations/blob/ef00d82fff37149ad9d926302306bfcd2d8c2d46/polling_stations/apps/pollingstations/models.py#L133) in the model that makes the text for this bit, but we rely on the po files for translation. I don't think it's essential, but I remember you saying the api should have text that clients can use, so thought I'd flag it in case you want to include it. I can see that the lack of on model translation might mean its easier to hand it off to the clients to deal with.

GeoWill (approved):
A nit pick and a bit of context/suggestion. Both can be ignored if you just want to merge it :)

polling_stations/apps/api/pollingstations.py
chris48s: I think the upshot of that conversation was we agreed that we would:

- For the text fields, return both the _en and _cy versions
- For the boolean fields, just return all the boolean fields in the API and make it the client code's responsibility to present them as text/icons/whatever

PR #8757, opened 2025-11-25
add 'has a11y info' to stations table

chris48s:
This adds a column to the stations table showing if we've got accessibility info for this station.

Because only admins can upload this info, the column is only showed to admins.

I also filtered it to just Wales because I think it is better not to waste a column on information we don't expect to hold in other areas for the time being.


GeoWill (approved):
This should be useful.

I guess if councils need to see it we can always take make it visible to them down the line.

PR #8871, opened 2026-02-17
sort all addresses together in pickers

chris48s:
This is the change we talked about in surgery the other day.
I've added a comment with a link to the previous commit in case we want to revisit it again


GeoWill (approved):
👍

PR #9027, opened 2026-03-05
Format ssm command runner errors before email

GeoWill:


cdk/lambdas/format_ssm_command_event/handler.py
```diff
@@ -0,0 +1,86 @@
...
+import logging
+import os
+import textwrap
+
+import boto3
+import sentry_sdk
+
+SNS_TOPIC_ARN = os.environ.get("SNS_TOPIC_ARN")
+DC_ENVIRONMENT = os.environ.get("DC_ENVIRONMENT")
+SENTRY_DSN = os.environ.get("SENTRY_DSN", "")
```
GeoWill: This uses WDIV sentry dsn. This is different to what was done with WDIV-Trigger lambda, but to me it makes more sense to group it under the WDIV sentry project than to create a new one.

PR #9196, opened 2026-03-19
Improve import logging

GeoWill:
* Councils seem to often have a 'placeholder' for when they don't have a UPRN, eg `N/A`. The duplicate uprn check did inadvertent double duty at stripping this. [6c82d0a](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9196/commits/6c82d0a1d6bca112e79b2bdc08ce63fc65844b84) catches non numeric uprns, reducing the false positives in the duplicate check. This doesn't deal with placeholders like `99999999999`. 
* Previously we didn't get loads more information when running an import with `-v 3`. [e2cb7de](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9196/commits/e2cb7ded5de021951c9f5ec1aaac7dd9bdfe6fb7) improves that.

I've added in pseudo 'error codes'. I'd like to actually enumerate these in a class, and eventually write these infos to the db so we can expose them on the site somewhere. But thought I'd do this as a first pass since it was mostly needed to deal with a query we had.

### DIFF
--- a/polling_stations/apps/data_importers/data_types.py
+++ b/polling_stations/apps/data_importers/data_types.py
@@ -6,6 +6,7 @@ import abc
 import logging
 from collections import namedtuple
 
+
 from addressbase.models import Address, UprnToCouncil, get_uprn_hash_table
 from councils.models import Council
 from pollingstations.models import PollingDistrict, PollingStation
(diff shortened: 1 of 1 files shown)


awdem (approved):
Tried this out locally and LGTM. Although I am wondering if it would be a nicer to separate the logging messages for the UPRNs that fail from the logging messages for stations added to the station set? Just so it's easier to trawl through the output when I'm just interested in what the bad UPRNs were. Or maybe we could have more fine-grained flags.

I think what you've done is totally fine to merge, though. The above could be future work, if you agree.

PR #9240, opened 2026-03-24
Convert advance voting station to many to many

GeoWill:
Still need:
* To do a big old rename: `s/advance/alternative/`.
* ~Deploy to dev~
* Make tickets for frontend improvements. 
* Do a complete addressbase import cycle to check it still works.
* Do we want alternative voting stations in the admin? (I think 'yes', but can be a bit of future work.)

---
  - https://app.asana.com/0/0/1213418260496026

### DIFF
--- a/polling_stations/apps/api/address.py
+++ b/polling_stations/apps/api/address.py
@@ -104,7 +104,9 @@ class PostcodeResponseSerializer(serializers.Serializer):
         read_only=True,
         help_text="A GeoJSON Feature containing a Point object describing the centroid of the input postcode.",
     )
-    advance_voting_stations = AdvanceVotingStationSerializer(read_only=True, many=True)
+    alternative_voting_stations = AdvanceVotingStationSerializer(
+        read_only=True, many=True
+    )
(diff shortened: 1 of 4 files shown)


chris48s:
There's a few things I want to comment on about the frontend code, but I don't really want to get bogged down in it on this PR. I would rather think of https://app.asana.com/1/1204880536137786/project/1213418260496019/task/1213418260496036?focus=true as a separate job.

On the frontend code: What do you think of this plan?

On this branch,

change `BasePollingStationView.get_advance_voting_station` (singular)

to

```py
def get_advance_voting_station(self):
    return None
```

and revert all the template changes in this PR.

That would allow us to merge the data model changes with no impact on the frontend. Any data that exists would just be totally ignored at the template layer. And that allows us to get the data work in and work on the API changes in devs.DC without thinking about frontend at all. Then would could cherry-pick your frontend commit onto a new branch as a starting point for https://app.asana.com/1/1204880536137786/project/1213418260496019/task/1213418260496036?focus=true and iterate on that stuff independently.

PR #9325, opened 2026-04-16
Pilot info url

GeoWill:
Snuck in the reordering too...

I've put the link on every `AdvanceVotingStation` object rather than change `alternative_voting_stations` to be an object. I figured this involved less change to the API even if we're the only ones to use it. I guess I could've put a top level sibling key, but not sure it matters given most of this code should be deleted.

### DIFF
--- a/polling_stations/apps/api/address.py
+++ b/polling_stations/apps/api/address.py
@@ -69,9 +69,17 @@ class AddressSerializer(serializers.HyperlinkedModelSerializer):
 class AdvanceVotingStationSerializer(serializers.ModelSerializer):
     class Meta:
         model = AdvanceVotingStation
-        fields = ("name", "address", "postcode", "location", "opening_times")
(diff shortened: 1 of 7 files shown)


polling_stations/apps/api/address.py
```diff
@@ -69,9 +69,17 @@ class Meta:
...
     class Meta:
         model = AdvanceVotingStation
-        fields = ("name", "address", "postcode", "location", "opening_times")
+        fields = (
+            "name",
+            "address",
+            "postcode",
+            "location",
+            "opening_times",
+            "pilot_info_url",
```
chris48s, line 78: Given this is the easiest way to just get a thing done, fine.

If we ever do this again, lets make this something that exists in the API once, not repeated for every hub. One of the multiple things to fix about this API response.

chris48s (approved):

PR #9416, opened 2026-05-14
Ci import slack notifications

awdem:
This PR makes it so imports scripts run in CI will report their success or failure to the bots channel in slack. This is meant to address the issue of us missing when import scripts fail to import after merging. We could have made `run_new_imports` fail instead, but this is less intrusive, while still achieving the goal of increasing visibility. I roughly copied the pattern for slack messaging used in [import_eoni](https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_importers/management/commands/import_eoni.py).

Rather than post one summary message per CI command, I chose to post one message per script, so that I could attach the data quality report or error message for each success/failure. 

In order to capture all of an import's logging in once place, I had to tweak the base importer command to pass its `stdout` to both the `LogHelper` and the `DataQualityReportBuilder`  ([898b191](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9416/commits/898b191aa5013f1633d8fe0ca0f7e10937f22876) and [e37e49c](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9416/commits/e37e49c7e184038f5e7f76195f0c803a82b0c2a8)).

Tested with unit tests and by deploying to dev.

---
  - https://app.asana.com/0/0/1213357905938200

### DIFF
--- a/polling_stations/apps/data_importers/management/commands/run_new_imports.py
+++ b/polling_stations/apps/data_importers/management/commands/run_new_imports.py
@@ -92,6 +92,7 @@ class Command(BaseCommand):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.slack_client = None
+        self.messages = []
 
     @property
     def slack_channel(self):
@@ -132,11 +133,12 @@ class Command(BaseCommand):
                         f"Ran import script: {script_path.name}",
                     )
                 )
-                if should_post_to_slack:
-                    self.post_to_slack(
-                        header_message=f":pollingstation: *Successfuly ran {script_path.stem}*",
-                        detail_message=f"```{out.getvalue()}```",
-                    )
+                self.messages.append(
+                    {
+                        "header_message": f":pollingstation: *Successfuly ran {script_path.stem}*",
+                        "detail_message": f"```{out.getvalue()}```",
+                    }
+                )
             except Exception as e:
                 # usually we want to handle a specific exception, but in this situation
                 # if there is any issue (at all) trying to run the command,
@@ -144,11 +146,12 @@ class Command(BaseCommand):
                 self.summary.append(
                     ("WARNING", f"{script_path.name} could not be run. Due to {e}")
                 )
-                if should_post_to_slack:
-                    self.post_to_slack(
-                        header_message=f":warning: *Failed to run {script_path.stem}*",
-                        detail_message=f"Error: {str(e)}",
-                    )
+                self.messages.append(
+                    {
+                        "header_message": f":warning: *Failed to run {script_path.stem}*",
+                        "detail_message": f"Error: {str(e)}",
+                    }
+                )
                 continue
 
     def run_misc_fixes(self):
@@ -274,3 +277,7 @@ class Command(BaseCommand):
             self.stdout.write("Not running import scripts")
 
         self.output_summary()
+
+        if should_post_to_slack:
+            for message in self.messages:
+                self.post_to_slack(**message)
--- a/polling_stations/apps/data_importers/management/commands/run_new_imports.py
+++ b/polling_stations/apps/data_importers/management/commands/run_new_imports.py
@@ -1,12 +1,19 @@
 import re
 import subprocess
+from io import StringIO
 from pathlib import Path
+from typing import Any, Dict
 
 import boto3
 import botocore
+from core.slack_client import SlackClient
 from django.apps import apps
 from django.conf import settings
 from django.core.management import BaseCommand, call_command
+from polling_stations.settings.constants.slack import (
+    BOTS_CHANNEL,
+    BOTS_TESTING_CHANNEL,
+)
 
 
 def get_paths_changed(from_sha, to_sha):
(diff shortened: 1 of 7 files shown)


polling_stations/apps/data_importers/management/commands/run_new_imports.py
```diff
@@ -160,13 +194,38 @@ def update_last_import_sha_on_ssm(self, to_sha, dc_environment=None):
...
             )
 
+    def post_to_slack(self, header_message: str, detail_message: str = None) -> None:
+        try:
+            response = self.post_header(header_message)
+            if detail_message:
+                thread_ts = response.get("ts")
+                self.post_details(thread_ts, detail_message)
+        except Exception as e:
+            self.stderr.write(f"Error posting to Slack: {str(e)}")
```
chris48s, line 209: We're creating another place where errors will be silently swallowed here.
If we want to plough on through this failure, can we send it to sentry so we have some visibility on the error.

We can use `capture_exception(e)` to do this, like this:

https://github.com/DemocracyClub/electionleaflets/blob/177b9fff874027edf29be3b0cdaf445d6eef3af8/electionleaflets/apps/leaflets/forms.py#L146-L148

polling_stations/apps/data_importers/management/commands/run_new_imports.py
```diff
@@ -96,25 +114,41 @@ def add_arguments(self, parser):
...
                 # if there is any issue (at all) trying to run the command,
                 # we just want to log it and move on to the next script
                 self.summary.append(
                     ("WARNING", f"{script_path.name} could not be run. Due to {e}")
                 )
+                if should_post_to_slack:
+                    self.post_to_slack(
+                        header_message=f":warning: *Failed to run {script_path.stem}*",
+                        detail_message=f"Error: {str(e)}",
+                    )
```
chris48s: Question: Is slack the right place to surface this?

Would it make sense to `capture_exception(e)` here and send it to Sentry, either instead of or as well as sending a notification to slack? In my mind, sentry is where exceptions go and it has the tooling to give us stacktrace, etc already

chris48s:
I'm going to make this one an optional comment, but my preferred pattern for something like this would be to build up messages as we go, and then post them all at the end instead of posting as we go.

So (pseudo-code)

```py
messages = []

for thing in things:
    try:
        # do the things
        messages.append({"header": "yay", "detail": "foo"})
    except Exception:
        messages.append({"header": "oh no", "detail": "bar"})

if should_post_to_slack:
    post_to_slack(messages)
```

I'll leave you to consider whether that's a useful approach here or not.. maybe there's not much in it here.

polling_stations/apps/data_importers/management/commands/run_new_imports.py
awdem: We post `import_eoni` failures to slack and not sentry in: https://github.com/DemocracyClub/UK-Polling-Stations/blob/ci-import-slack-notifications/polling_stations/apps/data_importers/management/commands/import_eoni.py#L327-L337

But I think what you're saying makes sense and maybe we should send failures to sentry.

awdem:
> I'm going to make this one an optional comment, but my preferred pattern for something like this would be to build up messages as we go, and then post them all at the end instead of posting as we go.

I went ahead and did this in https://github.com/DemocracyClub/UK-Polling-Stations/pull/9416/commits/6ea88d21b2860f9788ded0b6b3ccf8dbaead087c.

polling_stations/apps/data_importers/management/commands/run_new_imports.py
awdem: Have made sentry capture exceptions for both import script and slack failures in [aebc871](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9416/commits/aebc8719ab364248fd90875ee433265f00a27253).


I also made sentry capture slack failures in `import_eoni` in [b8bf161](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9416/commits/b8bf16190120a999ea08f51791ff2a0c07690523).

awdem: > ~~We post `import_eoni` failures to slack and not sentry in: https://github.com/DemocracyClub/UK-Polling-Stations/blob/ci-import-slack-notifications/polling_stations/apps/data_importers/management/commands/import_eoni.py#L327-L337~~

Actually we do capture `import_eoni` failures because we raise the errors after posting the failure to slack: https://github.com/DemocracyClub/UK-Polling-Stations/blob/ci-import-slack-notifications/polling_stations/apps/data_importers/management/commands/import_eoni.py#L105-L109

(I also checked `import_eoni_from_s3` and we're allowing the `import_eoni` errors to raise there too)

polling_stations/apps/data_importers/management/commands/run_new_imports.py
```diff
@@ -96,25 +116,46 @@ def add_arguments(self, parser):
...
+                out = StringIO()
+                call_command(script_path.stem, stdout=out, **opts)
                 self.summary.append(
                     (
                         "INFO",
                         f"Ran import script: {script_path.name}",
                     )
                 )
+                if should_post_to_slack:
+                    self.messages.append(
```
chris48s: One more tiny nitpick: You can remove the `if should_post_to_slack:` here and on line 152. We're going to check `if should_post_to_slack:` at the end so if we don't want to send them we can just bin it.

This then makes it easy to do something like

```py
if should_post_to_slack:
    for message in self.messages:
        self.post_to_slack(**message)
else:
    print(messages)
```

locally if we need to work on this, and removes a layer of nesting

awdem: Have done so in [e2a7d87](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9416/commits/e2a7d87a124c7202d8b0202337f5e1762f7b432a)

chris48s (approved):

PR #9417, opened 2026-05-14
Import improvements 20260505

awdem:
This PR makes several improvements to the import process:

- The first simply sorts the split postcodes output in the report ([0d0f87d](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9417/commits/0d0f87dfca56a875d7b4ee4052782568a48e904b))
- ~Makes it possible to hide a station's location in the admin by adding a third `LOCATION_HIDDEN` option to `VisibilityChoices` ([95b4a50](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9417/commits/95b4a50a9b077a2bfe449c8037cf473369a6de0d)..[cada2ec](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9417/commits/cada2ec0a2a1f37fe8ec4f7ff89ad27204f0c928))~
- Merges the two halarose importers into one (ref: https://github.com/DemocracyClub/UK-Polling-Stations/pull/8852)

The first so is so that's it easier to compare diffs when we've copy-pasted the split postcodes in. ~The second is to make it easier to remove maps when dealing with bug reports (and to allow non-devs to be able to do it).~

I also noticed that unpublished stations were still included in the `/pollingstations/` API endpoint, so I remove them in https://github.com/DemocracyClub/UK-Polling-Stations/pull/9417/commits/10f026bd352680db3e7c0e3f1a9476ee440d0331.

---
  - https://app.asana.com/0/0/1214888937251634


awdem:
As discussed in standup, I've removed the commits for hiding station location from this PR so we can discuss them separately.

polling_stations/apps/api/pollingstations.py
```diff
@@ -117,13 +117,16 @@ def get_queryset(self):
         council_id = self.request.query_params.get("council_id", None)
         station_id = self.request.query_params.get(self.id_field, None)
 
+        published_stations = PollingStation.objects.exclude(
+            visibility=VisibilityChoices.UNPUBLISHED
+        )
```
chris48s, line 122: This won't do any harm, but tbh I think there might be a case for just completely removing the pollingstation and pollingdistrict endpoints from this API

Do you know if they are used anywhere anymore? I think they probably aren't

chris48s (approved):

polling_stations/apps/api/pollingstations.py
awdem: I don't know of any uses of it, but probably worth checking with @GeoWill and Sym when he's back.

GeoWill: I thought it was used in the dashboard. But a quick skim of the code suggests not. https://github.com/DemocracyClub/UK-Polling-Stations/blob/902481a11f7ed3b3512988f009b8a3fe0febaaff/polling_stations/apps/dashboard/views.py#L76-L78

PR #9425, opened 2026-05-20
Add location source

awdem:
This PR adds a `location_source` field to the `PollingStation` model with four possible values, representing the data source we used to geocode a station:
- None
- Coordinates
- UPRN
- Postcode

To minimize repetition, I've replaced most of the per EMS `get_station_point` methods with a generic one on the parent class, that tries to geocode in the following order:
- Coordinates
- UPRN
- Postcode (if allowed)

I've also added the field as read-only to admin for easy access.

---
  - https://app.asana.com/0/0/1214961412363639


polling_stations/apps/data_importers/base_importers.py
```diff
@@ -1027,6 +1038,83 @@ def geocode_from_uprn(self, record):
...
+            return None
+
+    def geocode_from_coordinates(self, record):
+        return Point(
+            float(getattr(record, self.station_easting_field)),
+            float(getattr(record, self.station_northing_field)),
+            srid=self.srid,
+        )
+
+    def get_station_point(self, record) -> tuple[Point | None, str]:
```
awdem: In this method, I use generic fields like `self.station_northing_field` that are defined on the child classes. I wasn't sure if I should also add them as abstract methods to the `BaseCsvStationsCsvAddressesImporter` since the method won't work without them, e.g:

    @property
    @abc.abstractmethod
    def station_northing_field(self):
        pass

However, the `BaseCsvStationsCsvAddressesImporter` class isn't currently isn't currently marked as an abstract base class and I'm not sure if that is intentional or not.

polling_stations/apps/data_importers/ems_importers.py
```diff
@@ -461,6 +371,10 @@ class BaseDemocracyCountsCsvImporter(
     address_fields = ["add1", "add2", "add3", "add4", "add5", "add6"]
     postcode_field = "postcode"
     station_id_field = "stationcode"
+    station_uprn_field = ""
+    station_easting_field = "xordinate"
+    station_northing_field = "yordinate"
```
chris48s, line 369: The old `get_station_point()` method on this class hard-coded `srid=27700` when we did

```py
location = Point(
    float(record.xordinate), float(record.yordinate), srid=27700
)
```

Your new higher-level method just uses `srid=self.srid`

I think you need to set

```
srid = 27700
```

here at the class level now

polling_stations/apps/data_importers/ems_importers.py
chris48s, line 254: Similar point here. The old method hard-coded `srid=27700`

```py
location = Point(x_coord, y_coord, srid=27700)
```

so you need to set

```
srid = 27700
```

here, don't you?

polling_stations/apps/data_importers/base_importers.py
chris48s: First thing: I would say make the class abstract.

Then, if there's any properties which we _require_ descendents to implement for this method to make sense, make those abstract properties and we don't need to bother checking if they exist.

But also, there are some properties where you're checking:

```
if hasattr(record, self.some_property)
```

if `some_property` is genuinely optional for base classes to implement then don't make them abstract because that forces base classes to implement them.

So for example, if a child class can communicate "this EMS doesn't store/export UPRNs for polling stations" by not implementing `station_uprn_field` then don't make it a required property.
But if there's properties we always need to exist then make them abstract so it is an error to not implement them.

Maybe the best way to think of it is: There are 2 patterns:


## Required Field

```py
from abc import ABC, abstractmethod

class Parent(ABC):
    @property
    @abstractmethod
    def required_field(self):
        pass

    def do_thing(self):
        print(self.required_field)

class Child1(Parent):
   ...

class Child2(Parent):
   required_field = "value"

c = Child1() # Throws TypeError: Can't instantiate abstract class Child without an implementation for abstract method 'required_field'

c = Child2() # works just fine
```

## Optional Field

```py
class Parent():
    def do_thing(self):
        if hasattr(self, 'optional_field'):
            print(self.optional_field)

class Child(Parent):
   ...

c = Child() # works just fine
```

Use each one accordingly depending on what you want to achieve.

Then if you've done required fields, you'll probably need to make proper stub classes to unit test it.

polling_stations/apps/data_importers/base_importers.py
```diff
@@ -1027,6 +1038,83 @@ def geocode_from_uprn(self, record):
...
+            try:
+                location = self.geocode_from_uprn(record)
+                location_source = LocationSourceChoices.UPRN
+                self.logger.log_message(
+                    logging.INFO,
+                    "using UPRN for station %s",
+                    getattr(record, self.station_id_field),
+                )
+            except ObjectDoesNotExist:
+                pass
```
chris48s: I reckon write a log message if we take this path saying we tried to code from UPRN and failed.

polling_stations/apps/data_importers/base_importers.py
```diff
@@ -1027,6 +1038,83 @@ def geocode_from_uprn(self, record):
...
+        # if no coords or uprn, try postcode (if allowed)
+        if (
+            location is None
+            and self.allow_station_point_from_postcode
+            and hasattr(record, self.station_postcode_field)
+        ):
```
chris48s: Here lets check if postcode is `None` or empty string. No point trying to look it up if we know it will fail.

polling_stations/apps/pollingstations/admin.py
```diff
@@ -35,15 +35,20 @@ class UnmodfiableOSMWidget(OSMWidget):
         "internal_council_id",
     ]
 
-    # Filter on published status or council name
-    list_filter = ["visibility", ("council", admin.RelatedOnlyFieldListFilter)]
+    # Filter on published status, location source, or council name
+    list_filter = [
+        "visibility",
+        "location_source",
```
chris48s, line 41: Putting this in the admin is useful. Is there anywhere on the dashboard or uploader views it would also make sense to expose this data point?

polling_stations/apps/data_importers/ems_importers.py
awdem: The base importer sets the srid already:

https://github.com/DemocracyClub/UK-Polling-Stations/blob/0da365ec10a4653c1cbfd71e0bfdefce04bd657c/polling_stations/apps/data_importers/base_importers.py#L81

GeoWill:
Bit of a drive by comment, but I think we have enough FCS councils (and if they continue their rate of growth we will have a lot more) that we should either include the FCS importer in this work, or make a clear plan for follow up to implement it. At the moment the FCS importer only gets the station point from lat/long, but the FCS Api includes station UPRNs, so we could also use those if there are cases where lat/long is missing. This might have a bearing on where the new `get_station_point` method goes - i.e. on `BaseStationsImporter` rather than `BaseCsvStationsCsvAddressesImporter`. That might be too complicated though, but then I think it would be worth modifying the `BaseFcsDemocracyClubApiImporter` class.

chris48s:
Also a drive-by. Shall we have a quick chat about where we want to go with FCS importers next week and then work out the impact on this PR. I think it is also worth considering how we make this work with a larger number of councils using FCS. I don't think the way we're doing it now will work well if/when they have tens of users so I can see us wanting to make other changes to that base importer for that as well.

polling_stations/apps/data_importers/ems_importers.py
chris48s: Right, yes. I thought the default was 4326

awdem:
Summarizing our post standup conversation:

I've moved `get_station_point` to the `BaseStationsImporter` because the logic of what data we use to geocode a station and in what order should be shared across all importers (prefer coordinates over UPRN, postcode, etc.). Currently all the importers we use regularly should work. However, I haven't done anything to make Import classes that inherit from `BaseStationsDistrictsImporter` work and I'm not sure if they'd work or not. 

Our conclusion after standup is that we should remove the `BaseDistrictsImporter` and its child classes in a future PR since we never use them anymore.

We also talked the the FCS importer, which works more or less the same way as CSV importer, but could be much different since its source is a JSON API. Thinking about how to do this better should also be a separate PR.

I could hypothetically make the FCS importer inherit from `BaseStationsAddressesImporter` and remove some duplication from it, but I think if we're going to redo how we import FCS anyway, then I'll just leave it as it is for now since nothing is broken.

polling_stations/apps/data_importers/base_importers.py
```diff
@@ -336,6 +342,120 @@ def station_record_to_dict(self, record):
...
+            station_postcode = Postcode(
+                self.get_station_postcode(record),
+                validate=True,
+            ).with_space
+        except ValueError:
+            station_postcode = None
+
+        bad_values = ["", "0", "0.00", 0, None]
+        # try coords
+        if station_coords and station_coords not in bad_values:
```
chris48s: You're evaluating `station_coords not in bad_values` here, but `station_coords` is a 2-tuple and `bad_values` is a list of things that are all not tuples, so this is never going to be a match.

You need to unpack `station_coords` first.

polling_stations/apps/data_importers/base_importers.py
```diff
@@ -336,6 +342,120 @@ def station_record_to_dict(self, record):
...
+            location_source = LocationSourceChoices.COORDINATES
+            self.logger.log_message(
+                logging.INFO,
+                "using grid reference for station %s",
+                station_id,
+            )
+        # if no coords, try uprn
+        if location is None and station_uprn and station_uprn not in bad_values:
+            try:
+                ab_rec = Address.objects.get(uprn=station_uprn)
```
chris48s: The old `geocode_from_uprn()` function used to do a `lstrip("0")` to remove leading zeros.

I think you have lost that behaviour in this refactor. I'd suggest doing it further up when we first assign `station_uprn`.

So `station_uprn = self.get_station_uprn(record).lstrip("0")`

polling_stations/apps/data_importers/base_importers.py
```diff
@@ -336,6 +342,120 @@ def station_record_to_dict(self, record):
...
+            self.logger.log_message(logging.WARNING, message + "\n")
+
+    def get_station_postcode(self, record):
+        return None
+
+    def geocode_from_postcode(self, postcode):
+        raise NotImplementedError
+
+    def get_station_id(self, record):
+        raise NotImplementedError
```
chris48s: I think all these methods that raise `NotImplementedError` and maybe also `get_station_uprn` and `get_station_coordinates` look like they should really be abstract methods (`@abc.abstractmethod`) shouldn't they?
If a subclass needs to define them to make any sense, I'd say make them abstract.

`@abc.abstractmethod` is usually preferable to `raise NotImplementedError` because a class that doesn't implement it will fail at object construct time not method call time.

polling_stations/apps/pollingstations/admin.py
chris48s: Any further thoughts on this point?

polling_stations/apps/data_importers/base_importers.py
awdem: Ah my thought process was that not all station sources will have UPRNs or coordinates (e.g. Democracy Counts doesn't have a UPRN column), so I figured the get methods for UPRN, postcode, and coords should be optional. However, at at the minimum we need them to have a `station_id` so I made that method `raise NotImplementedError`. I can make that an abstract method instead (that will definitely break all the importers that inherit from `BaseDistrictsImporter`).

For `geocode_from_postcode`, though, I didn't make it an abstract method, because you only need to define it if you define `get_station_postcode`.

Will and I talked about making the coordinate methods required, but that would mean we expect all data sources to include a field with coordinate data.

polling_stations/apps/pollingstations/admin.py
awdem: Ah yeah I was gonna add it to the dashboard and possibly also the uploader. Will do that today.

polling_stations/apps/data_importers/base_importers.py
chris48s: Right yeah - that makes sense that not all EMSs export station UPRN/coords so those are actually optional :+1: 

`station_id` is required for any importer, but if changing it to abstract breaks old stations/districts tests then we can leave it as `raise NotImplementedError` with a TODO to make it abstract later.

I feel like station postcode isn't really optional (do we deal with any data where the stations don't have postcodes?). Tbh, if we're going to hoist all this stuff up to this level `geocode_from_postcode` feels like it should just be defined on this class. The method of getting a point from a postcode shouldn't depend on the input data, should it..

awdem: You're about `geocode_from_postcode` being defined here so I've done that. I've also added a TODO to `get_station_id`.

We don't deal with any postcode-less station data, but I suppose we might in the future? Although maybe we can safely say that we require a postcode field in any data we receive. In any case, making any method in this class an abstract method will break the district importers so I would probably just change `get_station_postcode` to `raise NotImplementedError` as well.

awdem: Have done the first two things above in https://github.com/DemocracyClub/UK-Polling-Stations/pull/9425/commits/f91e53a93afb56d9bd4e835fe4ad54ccc0ce785f and https://github.com/DemocracyClub/UK-Polling-Stations/pull/9425/commits/022284c3be6943052c09db2e325d4f57cedea682

polling_stations/apps/data_importers/base_importers.py
awdem: Fixed in https://github.com/DemocracyClub/UK-Polling-Stations/pull/9425/commits/70f052601c9c0f7b2c0a0f3fdcab4b66c9c13ded

polling_stations/apps/data_importers/base_importers.py
awdem: I ended up adding it here rather than higher up because `get_station_uprn` returns `None` if it isn't defined so it was breaking some tests. Done in https://github.com/DemocracyClub/UK-Polling-Stations/pull/9425/commits/29a717def5a300e035108c1f52332bab37f7e42c .

polling_stations/apps/pollingstations/admin.py
awdem: I've added it to to the dashboard in [08a922a](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9425/commits/08a922a1a1c49d22f3cf2acc13bf98f57d16e1e6) and to the uploader (for staff only) in [d014f45](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9425/commits/d014f45b01189bc1b3fc82e46cc7018b001d283a)

image

polling_stations/apps/pollingstations/models.py
```diff
@@ -54,6 +61,11 @@ class PollingStation(models.Model):
     postcode = models.CharField(blank=True, null=True, max_length=100)
     address = models.TextField(blank=True, null=True)
     location = models.PointField(null=True, blank=True)
+    location_source = models.CharField(
+        choices=LocationSourceChoices.choices,
+        default=LocationSourceChoices.NONE,
+        blank=True,
```
chris48s, line 67: I missed this first time round, but can we set a `max_length` on this `CharField`. `CharField` with no `max_length` is basically a full text field

polling_stations/apps/data_importers/ems_importers.py
```diff
@@ -337,47 +270,18 @@ def get_station_address(self, record):
             ]
         )
 
-    def get_station_point(self, record):
-        location = None
-        # Try coords first
-        x_coord = float(record.pollingvenueeasting)
-        y_coord = float(record.pollingvenuenorthing)
-        if x_coord > 0 and y_coord > 0:
-            location = Point(x_coord, y_coord, srid=27700)
```
chris48s, line 346: There's a little bit of validation here we lost in just checking against bad_values. Negative values are fine for srid=4326 but if we know we are dealing with easting/northing, then we can consider negative values invalid for our purposes (although django will still construct a `Point` object with them).

We will still need to allow both because FCS uses lat/lon. I wonder if it make sense to either:

- Throw an error at construct time if a class defines the stations srid is anything other than `(4326, 27700)` and just add a check for this somewhere if `srid=27700`
- Have each EMS class implement a `validate_coords()` method we can call from the bass class

polling_stations/apps/data_importers/base_importers.py
chris48s: Makes sense

polling_stations/apps/pollingstations/models.py
awdem: Done in [b6bc553](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9425/commits/b6bc553a5d31e14a73a1d6985dd4bbf134b4963b)

polling_stations/apps/data_importers/ems_importers.py
awdem: I opted for a `validate_coordinates()` method in [7ca15b9](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9425/commits/7ca15b9a117ca7cb9806fb90bda1daf1b59c8cb9) that has srid-specific validation.

chris48s: Yeah nice - that makes sense :+1:

chris48s (approved):
I reckon this is good. Lets :shipit: 

Lets also note the follow up work somewhere. I think there is some value in simplifying everything by getting rid of code we don't need any more and we're at the point in the year where that kind of thing is sensible to be working on.

PR #9434, opened 2026-05-28
Fix/import slack notifications

### LINKED ISSUE #9431: import slack notifications only post to bots testing
I need to to not check settings here, but rather the CI env:
https://github.com/DemocracyClub/UK-Polling-Stations/blob/import-haringey-2026-06-25/polling_stations/apps/data_importers/management/commands/run_new_imports.py#L98-L103

And I need to check that CI has an appropriate env variable in the jobs.

awdem:
This PR adds the `DC_ENVIRONMENT` to the CI env for the `run_new_imports` and `run_new_imports_post_deploy`. That way the `run_new_imports` command will know what env it's running in to determine the correct slack channel to post to. 

(In the issue I say I need to change where were getting the `DC_ENVIRONMENT` variable from in the command. This is wrong because in `base.py`, we get the `DC_ENVIRONMENT` from the env and when the `manage.py run_new_imports --slack ` runs, it populates the django settings from the CI env.)

I also did a drive-by fix of the CI syntax that was causing my IDE to underline everything in red.

Fixes https://github.com/DemocracyClub/UK-Polling-Stations/issues/9431

---
  - https://app.asana.com/0/0/1215160353214967

### DIFF
--- a/.circleci/deploy-config.yml
+++ b/.circleci/deploy-config.yml
@@ -136,6 +136,7 @@ jobs:
             export CI_DB_NAME=`aws ssm get-parameter --query Parameter.Value --name 'RDS_DB_NAME' --output text`
             export CI_DB_HOST=`aws ssm get-parameter --query Parameter.Value --name 'RDS_DB_HOST' --output text`
             export SLACK_TOKEN=`aws ssm get-parameter --query Parameter.Value --name 'SLACK_TOKEN' --output text`
+            export BOTS_CHANNEL=`aws ssm get-parameter --query Parameter.Value --name 'BOTS_CHANNEL' --output text`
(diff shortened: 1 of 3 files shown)


awdem:
I've changed this PR to get the slack channel from AWS parameter store rather than telling the CI jobs which env their running `manage.py run_new_imports` in so that CI is not running django as if it were in production, for instance.

I'll need to create the parameters before merging this.

chris48s:
This seems sensible. I think more generally, I want to get rid of directly accessing `os.environ` in runtime scripts and move all parsing of env vars into django settings but this isn't the only example of it in this codebase. We can deal with it under https://app.asana.com/1/1204880536137786/project/1215330308772861/task/1211598731391149?focus=true

chris48s (approved):

PR #9470, opened 2026-06-10
fix string interpolation

awdem:
I made a mistake with the string interpolation and it's breaking import scripts that trigger this logging.


chris48s:
I think this bug also exists in at least one other place:

https://github.com/DemocracyClub/UK-Polling-Stations/blob/e59446414022acadc747d3bf2bdad2bf3a076762/polling_stations/apps/data_importers/base_importers.py#L422-L427

Can you double-check for any more

awdem:
> I think this bug also exists in at least one other place:
> 
> https://github.com/DemocracyClub/UK-Polling-Stations/blob/e59446414022acadc747d3bf2bdad2bf3a076762/polling_stations/apps/data_importers/base_importers.py#L422-L427
> 
> Can you double-check for any more

Thanks for spotting this. I've double checked the rest of the code and I don't see any others.

chris48s (approved):

PR #9509, opened 2026-06-30
delete a bunch of stuff we don't use anymore

chris48s:
Refs https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1215587202006575?focus=true

This started off as a PR to remove the Google Directions API
Once I started getting trigger happy with the delete button, I realised there were various other bits of vestidual code I can ditch here. So I've also deleted a bunch of code related to:

- MapQuest (tile layer)
- MapZen (another directions service, now defunct)
- Google Analytics
- Raven (superseded by Sentry SDK)

The one reservation I have with removing Google Directions is that it does allow low volume use without an API key, so you can spin up the app and directions will work. Given that, I wonder if it is actually worth keeping `GoogleDirectionsClient` around as a fallback and just removing the code that allows us to set API keys for it.


GeoWill (approved):
I can't think of any places where there would be other references lurking, so I think this is good.

PR #9512, opened 2026-07-01
remove unused importers

awdem:
This PR removes a bunch of importers that we no longer use or intend to use again. I've also changed the parent class of 
`BaseFcsDemocracyClubApiImporter` for neatness, but I haven't made any further changes to it ahead of talking about how we import from FCS.


chris48s:
A few thoughts on this:

I think the stuff we're removing here makes sense, and it does simplify things a lot. Tbh, we might be able to strip out more layers. If every real importer we've got inherits from either `BaseStationsAddressesImporter` or `BaseCsvStationsCsvAddressesImporter` then everything above `BaseStationsAddressesImporter` in the hierarchy is fair game to collapse further. However it might be best to review where we are going with FCS as the next step before we go further on that.

Tbh, I would be inclined to remove the `BaseShpStationsCsvAddressesImporter`
If you look at the stuff we've moved up to `BaseStationsImporter` in #9425 stations in a shapefile wouldn't really make sense any more. `BaseStationsImporter` is too tightly coupled to what we see in EMS exports now, and we have zero real examples of `BaseShpStationsCsvAddressesImporter`.

I think it is reasonable to not do too much work on `BaseFcsDemocracyClubApiImporter` here given we're probably going to re-think that a bit, but if there are now whole methods on `BaseFcsDemocracyClubApiImporter` that are just duplicates or inferior versions of methods on the parent class, we might as well deal with that here. For example, I think you could just delete `BaseFcsDemocracyClubApiImporter.import_data` now. Maybe also `BaseFcsDemocracyClubApiImporter.get_station_point`?

There is also a question of: Where do we stop pulling this thread? The maximalist position would be that we completely delete the `PollingDistrict` model, the API endpoint, all code for querying it, etc. Tbh, I'm fairly convinced that none of that code is doing anything helpful at this point - particularly if we delete all the code that allows us to import any useful data into it. But is today the day to unravel all that?

I feel like there's definitely a bunch of code in
https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_importers/filehelpers.py
that we can now delete.

awdem:
I've done a bit more removing based on your comment above. 
- https://github.com/DemocracyClub/UK-Polling-Stations/pull/9512/commits/e13ce6d6861097f1062558690e726c058ca2b5d3 -  I've removed the now unused filetypes from `filehelpers.py`.
- [bc46573](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9512/commits/bc46573fd4ae70acf82090ea6a3e6aa698df03d4) - I removed the `BaseShpStationsCsvAddressesImporter`
- [2056e47](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9512/commits/2056e47fe49bd66bdfd2d90da81f7d721ac115c8) - I removed the redundant methods from the FCS importer, although I had to implement some new methods to in order to use `BaseStationsImporter.get_station_point`. 

I've also addressed a minor #TODO that I noticed in [2056e47](https://github.com/DemocracyClub/UK-Polling-Stations/pull/9512/commits/2056e47fe49bd66bdfd2d90da81f7d721ac115c8)


I left the `PollingDistrict` model alone for the time being. I agree that it could be deleted, but it's a bigger job than I set out to do at the beginning. However, I'm going to make a card for it to follow up in the future.

chris48s (approved):
I think this represents a reasonable place to stop pulling the thread

PR #9530, opened 2026-07-20
fix fcs station coordinates source

awdem:
Noticed while I was working on other things. I was trying to get the coordinates from fields that aren't set on the importer class.

### DIFF
--- a/polling_stations/apps/data_importers/ems_importers.py
+++ b/polling_stations/apps/data_importers/ems_importers.py
@@ -500,7 +500,7 @@ class BaseFcsDemocracyClubApiImporter(
         return record.get(self.postcode_field).strip()
 
     def get_station_coordinates(self, record):
-        return record.get(self.latitude_field), record.get(self.longitude_field)
+        return record["longitude"], record["latitude"]
 
     def station_record_to_dict(self, record):
         address = format_polling_station_address(


polling_stations/apps/data_importers/ems_importers.py
```diff
@@ -500,7 +500,7 @@ def get_station_postcode(self, record):
         return record.get(self.postcode_field).strip()
 
     def get_station_coordinates(self, record):
-        return record.get(self.latitude_field), record.get(self.longitude_field)
+        return record["latitude"], record["longitude"]
```
chris48s: As well as fixing the fields, should these be the other way round?
I think when we make a `Point()` object we expect lon, lat order - right?

GeoWill: Yes. 

We also have a method on `BaseCsvStationsCsvAddressesImporter` which does this:

https://github.com/DemocracyClub/UK-Polling-Stations/blob/6ad1c4ca1dafa084206e04e1a4981e5a73d0dd31/polling_stations/apps/data_importers/base_importers.py#L776-L784

But obviously FCS doesn't use that class, but I wonder if having it done once, and using common `self.<field_name>` is better than hardcoding the field names in different places? I appreciate that may not be the rabbit hole to go down with this PR.

awdem: Have reversed them. I agree that it's better not to hardcode the field names in different places, but I don't want to mess around with the importer any more because we're likely to change it again before it's even used.

chris48s (approved):

PR #9552, opened 2026-07-29
Import FCS data from s3

chris48s:
At a high level, this PR does 4 core things:

1. Allow us to store FCS API credentials in the DB via the admin
2. Allow admins to take a snapshots of data from FCS APIs and save the data to S3. I've implemented this as a 3-step "wizard", but each step can be skipped if the user doesn't actually need to make a choice (e.g: if there's only one possible date, skip the "data select" step, etc).
3. Allow the S3 lambda trigger to be able to recognise JSON data from FCS
4. Update the FCS base importer to consume files from S3

Along the way I have also given the uploader a bit of attention and fixed a few unrelated bugs that were annoying me while I was working on this:

1. There were some long-standing missing permission checks, which I have fixed
2. There was a formatting/display bug on the file detail card
3. Always show end users nicely formatted dates instead of YYYY-MM-DD

Follow up jobs:

- Delete the FCS creds from Circle CI
- Add them into the prod WDIV DB attached to the relevant council
- Next time there is a by-election happening in an area where we know they use FCS, do a proper end-to-end test


polling_stations/apps/file_uploads/migrations/0008_fcscredential_council_onetoone.py
chris48s, line 1: **TODO**: squash these two migrations

cdk/lambdas/wdiv-s3-trigger/trigger/json_helpers.py
```diff
@@ -0,0 +1,51 @@
...
+    if isinstance(data, list) and len(data) > 0:
+        if all(key in data[0] for key in fcs_fields):
+            return "FCS"
+    return "unknown"
+
+
+def get_json_report(response, key):
+    report = {
+        "csv_valid": False,
+        "csv_rows": 0,
```
chris48s, line 39: **TODO**: replace this with a count of the number of address objects

chris48s (commented):
couple of things I need to finish off..

polling_stations/apps/file_uploads/fcs_views.py
```diff
@@ -0,0 +1,188 @@
...
+
+from councils.models import Council
+from data_finder.helpers.every_election import EEFetcher, EEWrapper
+
+from .forms import ElectionDateForm, ElectionForm
+from .mixins import StaffUserRequiredMixin
+from .models import Upload
+
+
+class FcsDateSelectView(StaffUserRequiredMixin, TemplateView):
```
chris48s, line 24: As agreed, I have made all these new views only accesible to admins. We will dogfood it ourselves before exposing to councils, but the end goal should be to expose this to council users.

If we want to expose to councils, we change all these to use `CouncilMatchesMixin` instead of `StaffUserRequiredMixin` and update the link/text in `council_detail.html`

polling_stations/apps/file_uploads/mixins.py
```diff
@@ -0,0 +1,48 @@
...
+    permission_denied_message = "You must be a staff user to access this page."
+
+    def get_login_url(self):
+        return reverse_lazy("file_uploads:council_login_view")
+
+    def test_func(self):
+        return self.request.user.is_staff
+
+
+class CouncilMatchesMixin(ABC, UserPassesTestMixin):
```
chris48s, line 17: This fixes that long-standing permissions bug. Some of the council views were only checking for an active user.

polling_stations/apps/file_uploads/templates/file_uploads/fcs/election_select.html
```diff
@@ -0,0 +1,37 @@
...
+{% load static %}
+{% load pipeline %}
+
+{% block council_content %}
+    <h1>{{ council.name }}</h1>
+    {% if NO_UPCOMING_ELECTIONS %}
+        <p>
+            ⛔ We queried the FCS API, but we couldn't find any
+            elections happening on {{ date|human_date }}.
+        </p>
```
chris48s, line 17: I wonder if there is a more helpful way to express this if we've managed to end up here?

In general, feedback on the additional template text I've added would be good. Although initially these views will be only exposed to admins I've tried to write the copy as if it were council-facing and it would be good to get another perspective on how to explain what is happening as the user navigates through this process.

polling_stations/apps/file_uploads/urls.py
```diff
@@ -30,6 +36,21 @@
         FileUploadView.as_view(),
         name="file_upload",
     ),
+    path(
+        "fcs/date_select/<str:council_id>/",
```
chris48s, line 40: I've used `council_id` as the param name for all the new routes I've added.

I've not done anything about the fact one of the old ones use `gss` despite the fact that the value is not a GSS code (its a 3 letter code). I've just left that bit of scope creep for now. Naming things well is hard, particularly if you've already named them badly.

cdk/lambdas/wdiv-s3-trigger/trigger/json_helpers.py
```diff
@@ -0,0 +1,51 @@
...
+        "errors": [],
+    }
+    body = response["Body"].read()
+    if len(body) == 0:
+        report["errors"].append("File is empty")
+        return report
+
+    try:
+        data = json.loads(body)
+        report["csv_valid"] = True
```
chris48s, line 51: I've not bothered changing the field names here. So we're still using `csv_valid`, `csv_encoding`, etc here even though we're now dealing with JSON too. I think we just live with that tbh. I don't think it is enough of an issue to deal with the lift of changing it. Again, naming things well is hard, especially when you've already named them badly.

chris48s:
There's a few things I need to circle back to before I mark it ready for review.

polling_stations/apps/file_uploads/fcs_views.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+
+        resp.raise_for_status()
+
+        conn = boto3.client("s3")
+        conn.put_object(
+            Bucket=settings.S3_UPLOADS_BUCKET,
+            Key=f"{council.council_id}/{election_date}/{now}/snapshot.json",
+            Body=resp.text,
+        )
```
chris48s: Here I am both fetching the data from the FCS API and saving it to s3 inside the request/response cycle. As noted when we discussed this, for a sufficiently large response this could take too long. The solution would be to farm this out to a background task, but I reckon lets try this to start with and see how far it takes us.

**TODO**: Find the largest real-world response we can, deploy to dev and see how long this takes end-to-end on EC2

chris48s: I've tested this with the biggest response I could (~30Mb JSON, ~150k addresses). It took ~8-10 seconds.

That's not _good_ but I think it is workable.

I think a really big council (side note: within the next few years councils of ~500k population is going to be the norm so at that point most councils will be big - something to think about) probably would break this so we do ultimately need to kick this into the background.

That said, I think we can run with this implementation as it is and do something useful/workable without having to do that from day one.

polling_stations/apps/data_importers/ems_importers.py
awdem, line 1: Minor nitpick, but `BaseFcsDemocracyClubImporter` is the only class in this file now without a docstring.

polling_stations/apps/file_uploads/mixins.py
awdem: Nice one

polling_stations/apps/file_uploads/fcs_views.py
awdem: I think this is fine while it's just us pushing the button as well, but it's definitely something we'd want to sort out before letting councils do it themselves.

awdem: Just testing locally, and if I hammer on the snapshot button I can trigger a multiple requests in a row to the API. We might want to explicitly guard against this by disabling the button on click or something before we let council users hit the button - especially since we're fetching a non-trivial amount of data from an external API. 

Not sure if it needs to be in this PR, but just wanted to note it.

polling_stations/apps/file_uploads/templates/file_uploads/fcs/date_select.html
```diff
@@ -0,0 +1,39 @@
...
+    {% else %}
+        <form id="file_upload_form" method="POST">
+            {% csrf_token %}
+            <fieldset>
+                <legend>Election date</legend>
+                <p>
+                    There is more than one upcoming election in your region. Please pick the date this upload is for.
+                    if you have more than one upcoming election, you will need to upload different data exports for each,
+                    or email us if you would like us to reuse the data.
+                </p>
```
awdem, line 21: This is the same text as for the file upload path, and I don't think it will make sense for future FCS users. Maybe something like:


```suggestion
                <p>
                    There is more than one upcoming election in your region. Please pick the date you'd like to submit a data snapshot for.
                    If you have more than one upcoming election, you will need to submit a data snapshot for each. 
                    If your data has changed since you last submitted a snapshot, you'll need to submit another snapshot in order to see those changes.
                </p>
```

Not sure about the last line..

polling_stations/apps/file_uploads/templates/file_uploads/fcs/election_select.html
awdem: I don't know how much to assume about the council users, but it's possible they won't know what we mean by FCS API. Maybe it's better to assume zero knowledge and say something more mushy like: `We've gotten in touch with your EMS provider and they don't have any elections happening on {{ date|human_date }}`

Regardless of the wording, I think we should also add a line telling them to get in touch with us (or maybe FCS directly?) about it.

polling_stations/apps/file_uploads/fcs_views.py
```diff
@@ -0,0 +1,188 @@
...
+                f"/api/DemocracyClub/Election/{election_id}/PollingStation",
+            ),
+            headers={
+                "Accept": "*/*",
+                "User-Agent": "Scraper/DemocracyClub",
+                "X-API-KEY": council.fcscredential.token,
+            },
+        )
+
+        resp.raise_for_status()
```
awdem: Once this is council-facing, we should consider how we handle http errors from FCS, since it's an external API. 

 We might want a custom error page telling users to try again later, and/or get in touch with us (or maybe FCS) if the issue persists, etc. 

I don't think this needs to be in this PR though.

awdem (changes requested):
I had a go at snapshotting some data locally, plus I when through the import process for a `snapshot.json` file to try it out and it all seems to be working fine.

I've made a few suggestions RE the templates and in a few other place, but I think it could all wait for the future PR(s) that will precede exposing this to council users as the groundwork is solid.

One last thing that I was thinking of - Is the only way we let users know an upload/snapshot was successful via an email? Would it be a good idea to include a status message after being redirected back to the API page? Again, this can also wait.

polling_stations/apps/file_uploads/templates/file_uploads/fcs/election_select.html
chris48s: Yeah good point.
I think it is reasonable to name their software supplier, but "API" is probably not helpful.
FCS refer to their software as ElectionsPro (see https://fcssoftware.co.uk/fcs-elections )
so I've incorporated that into the message here.

9fa384d7

polling_stations/apps/data_importers/ems_importers.py
chris48s: :+1: 

bbb88a50

polling_stations/apps/file_uploads/templates/file_uploads/fcs/date_select.html
chris48s: Thanks
I've gone with a slight variation on this

918c6a46

polling_stations/apps/file_uploads/fcs_views.py
chris48s: Yeah agreed. I think at this stage it is useful if this manifests as a notification in Sentry.
Then we will get an idea of how flaky or reliable these APIs are.

polling_stations/apps/file_uploads/fcs_views.py
chris48s: This is a good point. I'm going to have a quick think about how I can improve this

chris48s: In 90a8bb61 I have implemented a bare-bones cache lock for this request.

https://github.com/DemocracyClub/UK-Polling-Stations/commit/90a8bb616048cf29def4ce6aa98c9fbc3de7ace4?w=1 for diff without whitespace. I am going to leave a few inline comments specifically on that diff further down.

polling_stations/settings/local.example.py
```diff
@@ -19,7 +19,7 @@
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
 CACHES = {
     "default": {
-        "BACKEND": "django.core.cache.backends.dummy.DummyCache",
+        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
```
chris48s, line 22: If you have the cache engine set to `DummyCache` in your `local.py` you'll need to switch it to something else to test this (local memory being the easiest).
When we deploy, this will use the DynamoDB cache

polling_stations/apps/file_uploads/fcs_views.py
```diff
@@ -0,0 +1,207 @@
...
+        # validate
+        dt.datetime.strptime(election_date, "%Y-%m-%d")
+
+        election_id = kwargs["election_id"]
+
+        now = dt.datetime.now().strftime("%Y-%m-%dT%H:%M:%S.%f")
+
+        # acquire lock
+        # https://github.com/DemocracyClub/UK-Polling-Stations/pull/9552#discussion_r3706676826
+        cache_key = f"fcslock-{council.council_id}-{election_date}"
```
chris48s, line 162: This limits us to one simultaneous upload for any given council/election date combo

polling_stations/apps/file_uploads/fcs_views.py
```diff
@@ -0,0 +1,207 @@
...
+        if cache.get(cache_key):
+            messages.error(
+                self.request, "snapshot already in progress - refresh page in 1 minute"
+            )
+            return redirect(
+                reverse("file_uploads:councils_detail", kwargs={"pk": council.pk}),
+            )
+        else:
+            cache.set(cache_key, True, timeout=40)
```
chris48s, line 171: Using a try/finally should mean this usually gets unlocked cleanly at the end of the request.

In the absolute worst case scenario, the user can try again in 40 seconds when the cache key expires.

polling_stations/apps/file_uploads/fcs_views.py
```diff
@@ -0,0 +1,207 @@
...
+        # acquire lock
+        # https://github.com/DemocracyClub/UK-Polling-Stations/pull/9552#discussion_r3706676826
+        cache_key = f"fcslock-{council.council_id}-{election_date}"
+        if cache.get(cache_key):
+            messages.error(
+                self.request, "snapshot already in progress - refresh page in 1 minute"
+            )
+            return redirect(
+                reverse("file_uploads:councils_detail", kwargs={"pk": council.pk}),
+            )
```
chris48s, line 169: Not amazing, but will do for MVP I think

polling_stations/apps/file_uploads/fcs_views.py
chris48s: I've also added a simple js button disable in 9ff53ee3 so we have 2 layers

awdem (approved):
Tested the caching locally and seems good to me. I think this is good merge once you've removed the dev deploy commit.

chris48s:
For my own reference, the next things to do on this are:

1. Remove the "dev deploy" commit
2. Squash the fixups
3. Merge this
4. Delete FCS credentials from Circle CI env vars
5. Add all the FCS details to WDIV prod in the admin
6. Explain what we've changed to Peter

I've got another PR that I want to try and finish up before the end of the day, so I will probably do this in the morning.

PR #9559, opened 2026-07-30
Fix import EONI from s3

GeoWill:
EONI changed how they name the files they upload, and that broke this command. This means we've been showing stale data. But also they've been updating the LC scheme, so there haven't been many updates to the WM scheme.

This uses s3 'last modified' rather than file name to figure out the most recent file. I'm not certain that is better, but hopefully.

This can be tested locally by changing
applying:
```diff
diff --git a/polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py b/polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py
index 86d7aafc5..3ca2105db 100644
--- a/polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py
+++ b/polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py
@@ -150,7 +150,7 @@ class Command(BaseCommand):
         try:
             # Set env.
             # Command works without, but requires '--input-uri' and will only post to test channel in slack.
-            self.dc_environment = settings.DC_ENVIRONMENT
+            self.dc_environment = "production"  # settings.DC_ENVIRONMENT
 
             # Set up s3 config
             self.set_s3_key_and_wrapper(options)
```

and running: `AWS_PROFILE=prod-wdiv-dc uv run ./manage.py import_eoni_from_s3`

### DIFF
--- a/polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py
+++ b/polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py
@@ -1,5 +1,4 @@
 import csv
-import re
 from pathlib import Path
 
 import chardet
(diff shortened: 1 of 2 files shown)


polling_stations/apps/data_importers/management/commands/import_eoni_from_s3.py
```diff
@@ -252,16 +224,18 @@ def download_file(self):
...
+        ]
 
-        if len(object_keys) == 0:
+        if len(objects) == 0:
             raise CommandError(
                 f"Failed to find any EONI export keys for {self.bucket_name}"
             )
-        return object_keys[-1]
+
+        return sorted(objects, key=lambda obj: obj.key, reverse=True)[0].key
```
chris48s: isn't using .key here still fundamentally sorting them by filename?

GeoWill: Yes. Typing across from the debugger fail. 
Fixup incoming.

chris48s (approved):

PR #9561, opened 2026-07-30
Report EONI file to slack

GeoWill:
We can fail in two places - the import_eoni_from_s3.py command and the import_eoni.py command. That's why both files are changed. We only report success from import_eoni.py.

---
  - https://app.asana.com/0/0/1217011285414717

### DIFF
--- a/polling_stations/apps/data_importers/management/commands/import_eoni.py
+++ b/polling_stations/apps/data_importers/management/commands/import_eoni.py
@@ -73,6 +73,11 @@ class Command(BaseStationsImporter, CsvMixin):
             action="store_true",
         )
         parser.add_argument(
+            "--s3-source",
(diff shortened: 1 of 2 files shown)


chris48s (approved):
Thanks, I hate it.

This will work, but it is a bit hideous. I think this is a sign we need to refactor this a bit. My gut instinct is that `import_eoni` shouldn't directly post to slack and `import_eoni_from_s3` should be able to:

- catch exceptions thrown by `handle()`
- interrogate the `eoni_importer` object for the information needed to report success

and then `import_eoni_from_s3` does all the slack reporting in one place. That both would negate the need for things like this, and make `import_eoni` look more like a "normal" importer.

For the moment, lets roll with this but I would like also like to clean it up.

